PR #4, opened 2022-08-23
Intial Rabbitmq operator integration

dprince:


custom-bundle.Dockerfile
```diff
@@ -23,6 +23,7 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o csv-merger csv-merger.g
 #ARG KEYSTONE_BUNDLE=quay.io/openstack-k8s-operators/keystone-operator-bundle:latest
 FROM quay.io/openstack-k8s-operators/keystone-operator-bundle:latest as keystone-bundle
 FROM quay.io/openstack-k8s-operators/mariadb-operator-bundle:latest as mariadb-bundle
+FROM quay.io/dprince/rabbitmq-cluster-operator-bundle:latest as rabbitmq-bundle
```
abays, line 26: Should we push this image to `quay.io/openstack-k8s-operators` instead?

pkg/openstack/keystone.go
```diff
@@ -0,0 +1,58 @@
...
+			return err
+		}
+		return nil
+	})
+
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
+			helper.GetLogger().Info("KeystoneAPI %s not found, reconcile in 10s")
+			return ctrl.Result{RequeueAfter: time.Duration(10) * time.Second}, nil
+		}
```
abays, line 49: For this and the other resources (MariaDB and RabbitMQ), is it even possible to get a "not found" here?  Wouldn't the `CreateOrPatch` have created the resource if it was in fact not there beforehand?

custom-bundle.Dockerfile
dprince: Yes. Will correct it.

dprince: Just pushed. Actually had to fix issues with the new bundle build first. The latest build shows as master-latest instead of 'latest' which seems fine but may need to follow up on this to make sure it tracks correctly. Not sure why other repos are using just 'latest'

pkg/openstack/keystone.go
dprince: I think it should actually be:
        if err != nil && !k8s_errors.IsNotFound(err) {

I had copied this over from our lib-common statefulset but it looks wrong as well. So I guess I should push a fix there too.

abays (approved):
/lgtm

PR #5, opened 2022-08-30
Update rabbitmq-operator-bundle image to use latest tag

dprince:
The build scripts have been fixed to generate just :latest


abays (approved):
/lgtm

PR #18, opened 2022-09-19
Add top level nodeSelector

dprince:
Currently implemented for keystoneAPI only


apis/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -34,6 +34,10 @@ type OpenStackControlPlaneSpec struct {
 	// StorageClass -
 	StorageClass string `json:"storageClass,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// NodeSelector to target subset of worker nodes running control plane services
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
abays, line 39: Maybe we should add a temporary comment here that indicates that this only currently works for Keystone?  Otherwise I think this patch is ready

dprince: I can do that. MariaDB doesn't support it yet. And Rabbit will work differently. Need to have discussion on that one.

dprince: @abays comment added

abays (approved):
/lgtm

PR #20, opened 2022-09-20
Add PlacementAPI to ControlPlane CR

dprince:


pkg/openstack/placement.go
```diff
@@ -0,0 +1,54 @@
...
+// ReconcilePlacementAPI -
+func ReconcilePlacementAPI(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, helper *helper.Helper) (ctrl.Result, error) {
+	placementAPI := &placementv1.PlacementAPI{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "placement",
+			Namespace: instance.Namespace,
+		},
+	}
+
+	helper.GetLogger().Info("Reconciling placementAPI", "placementAPI.Namespace", instance.Namespace, "placementAPI.Name", "keystone")
```
abays, line 26: I think `"placementAPI.Name"` shouldn't have double-quotes around it here

abays: Nevermind, I see what you're doing

abays, line 26: But `s/keystone/placement`

gibizer, line 26: Based on the discussion in https://coreos.slack.com/archives/CQXJFGMK6/p1663595278584359 we should use the logging wrappers from https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/util/log.go But probably that is for a separate patch to replace all the helper.GetLogger() calls in the openstack-operator.

pkg/openstack/placement.go
```diff
@@ -0,0 +1,54 @@
...
+	op, err := controllerutil.CreateOrPatch(ctx, helper.GetClient(), placementAPI, func() error {
+		instance.Spec.PlacementTemplate.DeepCopyInto(&placementAPI.Spec)
+		if placementAPI.Spec.Secret == "" {
+			placementAPI.Spec.Secret = instance.Spec.Secret
+		}
+		if placementAPI.Spec.NodeSelector == nil && instance.Spec.NodeSelector != nil {
+			placementAPI.Spec.NodeSelector = instance.Spec.NodeSelector
+		}
+		if placementAPI.Spec.DatabaseInstance == "" {
+			placementAPI.Spec.DatabaseInstance = "openstack" //FIXME: see above
```
gibizer, line 36: The FIXME probably need to refer to the comment in https://github.com/openstack-k8s-operators/openstack-operator/blob/becaa3f96db8e694e0a0dbec91a73d2ab89afac0/pkg/openstack/keystone.go#L36

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -140,6 +149,7 @@ func (r *OpenStackControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) err
 		For(&corev1beta1.OpenStackControlPlane{}).
 		Owns(&mariadbv1.MariaDB{}).
 		Owns(&keystonev1.KeystoneAPI{}).
+		Owns(&placementv1.PlacementAPI{}).
```
gibizer, line 152: Just out of curiosity does this line means
 - a) the openstack-operator controller watches the PlacementAPI CRs *it creates*
 - b) the openstack-operator controller watches *all* PlacementAPI CR creation even if it is created manually outside of the openstack-operator controller.

abays: It would be `b`.  But it is possible to only have it watch specific `PlacementAPI` CRs, if they are properly labeled and then filtered through a watch function, like we do for ConfigMaps here: https://github.com/openstack-k8s-operators/cinder-operator/blob/77b8278c88f1327ff4ccc050f360a33d246b4c47/controllers/cinderapi_controller.go#L199-L247

gibizer: Thanks @abays

pkg/openstack/placement.go
dprince: agreed

pkg/openstack/placement.go
dprince: I think we can remove it for now

gibizer:
looks good to me

abays (approved):
/lgtm

PR #24, opened 2022-10-03
Add Cinder to ControlPlane CR

dprince:


abays (commented):
Looking good so far

abays:
I think this might be missing changes to `custom-bundle.Dockerfile` to inject the actual Cinder operator into the bundle.  When I build and deploy it I don't see a Cinder controller-manager get created yet.

abays:
`main.go` also needs to register the Cinder `v1beta1` scheme

Akrog:
It doesn't deploy the Cinder operator for me either.

dprince:
Sorry guys, That was a sloppy rebase on me. Re-adding those things and will re-test again myself

custom-bundle.Dockerfile
```diff
@@ -27,6 +27,7 @@ FROM quay.io/openstack-k8s-operators/mariadb-operator-bundle:latest as mariadb-b
 FROM quay.io/openstack-k8s-operators/rabbitmq-cluster-operator-bundle:latest as rabbitmq-bundle
 FROM quay.io/openstack-k8s-operators/placement-operator-bundle:latest as placement-bundle
 FROM quay.io/openstack-k8s-operators/glance-operator-bundle:latest as glance-bundle
+FROM quay.io/openstack-k8s-operators/cinder-operator-bundle:latest as cinder-bundle
```
abays, line 30: We'll also need to `COPY` from `cinder-bundle` and then pass its CSV to the `csv-merger` below

dprince: I did this in a follow up. Thanks.

PR #27, opened 2022-10-06
Rabbitmq TransportURL CR support

dprince:
Add logic to generate transport-url secrets

This implements logic to generate a k8s secret containing a transport-url for each TransportURL CR. The current implementation uses the default rabbitmq username, password, host information (similar to OSP today).


apis/rabbitmq/v1beta1/transporturl_types.go
```diff
@@ -0,0 +1,62 @@
...
+//+kubebuilder:subresource:status
+
+// TransportURL is the Schema for the transporturls API
+type TransportURL struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   TransportURLSpec   `json:"spec,omitempty"`
+	Status TransportURLStatus `json:"status,omitempty"`
+}
```
Akrog, line 49: ?: I assume the `cinder-operator` would watch this CR and should update its services when it changes, but how would it know that the the `Spec` and `Status` are in sync?

Akrog (commented):
Thank you for working on this. In my mind there are multiple approaches to this, and neither of them match this implementation, so I must be missing something.

The approaches I see are:

#### Approach 1: Openstack-operator driven
- SysAdmin creates `OpenStackControlPlane` CR that contains the `rabbitmqTemplate`
- The Openstack Operator based on the rabbitmqTemplate:
  - Deploys the RabbitMQ operator and cluster
  - Decides on a name for `TransportURL` for cinder and builds the `Cinder` CR using that name
  - Deploys the cinder-operator and the generated `Cinder` CR
- The cinder-operator waits until the `TransportURL` CR from its spec exists (new condition for cinder-operator)
- Once the RabbitMQ cluster is up and running the openstack-operator creates the `TransportURL`
- Cinder-operator detects the existence of its `TransportURL` and continues deploying

#### Approach 2: Cinder-operator driven

- The `OpenStackControlPlane` CR doesn't have the `rabbitmqTemplate` at a top level, it is defined on the Cinder part
- The openStack-operator only deploys the RabbitMQ operator (not an actual RabbitMQ cluster) and the Cinder operator and CR.
- The cinder-operator sees the definition of the RabbitMQ and creates the request for a RabbitMQ cluster using a CR
- The openstack-operator has a controller for that request, deploys the rabbit cluster and once it's done updates the cinder request CR with the name of the transport url secret.

This has the benefit of supporting what Sean was asking for, support for multiple RabbitMQs, but makes the `OpenStackControlPlane` CR more cumbersome.

#### Approach 3: Both

This approach supports both, setting the rabbitmq template at the top level and at a service level.

--------

The problem I see with the current approach is that I don't see the need for a controller for the `TransportURL` CRD, it's something that the opensntack-operator could create in one shot during the `OpenStackControlPlane` reconciliation with a new condition.

Akrog:
@dprince gave me the reasons why we are doing it this way.
Capturing them here for completeness:

- One side is that we want to try to make this operator transparent in regards to the other operators' section of the spec, so we don't want the openstack-operator having to check the cinder-spec section to look for the rabbitmq information.

- The other side is that, for security reasons, we don't want to give permissions to all the different operators to access the rabbitmqclusters secrets, so we want to keep it in the openstack-operator.

controllers/rabbitmq/transporturl_controller.go
```diff
@@ -0,0 +1,193 @@
...
+		r.Kclient,
+		r.Scheme,
+		r.Log,
+	)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Lookup RabbitmqCluster instance by name
+	labelSelector := map[string]string{
```
gibizer, line 108: I'm wondering why don't we use the Name and namespace directly to query the RabbitMqCluster instance. We could use h.GetClient().Get() instead of  h.GetClient().List() as names are unique in a namespace.

controllers/rabbitmq/transporturl_controller.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// Wait on RabbitmqCluster to be ready
+	rabbitReady := false
+	for _, condition := range rabbit.Status.Conditions {
+		if condition.Reason == "AllPodsAreReady" && condition.Status == "True" {
+			rabbitReady = true
+		}
+	}
+	if !rabbitReady {
+		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
```
gibizer, line 125: could you please make this configurable as a time.Duration? In envtest these waits can be sortened significantly to speed up the test execution. See https://github.com/openstack-k8s-operators/lib-common/pull/68 for reference.

controllers/rabbitmq/transporturl_controller.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	// Create a new secret with the transport URL for this CR
+	secret := createTransportURLSecret(instance, string(username), string(password), string(host))
+	_, op, err := oko_secret.CreateOrPatchSecret(ctx, helper, instance, secret)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if op != controllerutil.OperationResultNone {
+		return ctrl.Result{RequeueAfter: time.Second * 5}, util.WrapErrorForObject(
```
gibizer, line 158: ditto

controllers/rabbitmq/transporturl_controller.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	host, err := base64.StdEncoding.DecodeString(string(defaultUserSecret.Data["host"]))
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Create a new secret with the transport URL for this CR
+	secret := createTransportURLSecret(instance, string(username), string(password), string(host))
```
gibizer, line 152: Do I understand correctly that this means today all services will use the same user / password to access the message bus? If so could you please note that limitation here or in some documentation.

pkg/rabbitmq/funcs.go
```diff
@@ -0,0 +1,66 @@
...
+	listOpts := []client.ListOption{
+		client.InNamespace(namespace),
+	}
+
+	if len(labelSelector) > 0 {
+		labels := client.MatchingLabels(labelSelector)
+		listOpts = append(listOpts, labels)
+	}
+
+	err := h.GetClient().List(ctx, rabbitmqClusterList, listOpts...)
```
gibizer, line 49: Asthe TransportURL CR gets the name of the RabbitMqCluster CR we could use Client.Get() instead of List here to get the RabbitMqCluster CR. This would simplify the logic here.

gibizer (changes requested):
The overall approach looks good to me. I have some requests inline.

controllers/rabbitmq/transporturl_controller.go
dprince: For now, yes. All services will use the default rabbit username and password (this is how it works in OSP today right?).

However, in the future I think we could use vhosts to isolate things and thus have each service get its own rabbit vhost and associated username/password.

apis/rabbitmq/v1beta1/transporturl_types.go
Akrog: And an admission webhook that would clear that flag when the Spec is changed, right?

abays: Or you could hash the spec and store it in the status, and then change condition when old hash does not equal current hash

controllers/rabbitmq/transporturl_controller.go
```diff
@@ -0,0 +1,193 @@
...
+		return ctrl.Result{}, err
+	}
+	password, err := base64.StdEncoding.DecodeString(string(defaultUserSecret.Data["password"]))
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	host, err := base64.StdEncoding.DecodeString(string(defaultUserSecret.Data["host"]))
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
abays, line 149: I wonder if the secret acquisition here should use the `lib-common` library [1] we have for such things?

[1] https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/secret/secret.go

apis/rabbitmq/v1beta1/transporturl_types.go
dprince: As I originally had this the existence of the secret means it was successful. This wouldn't cover the case for updates however so we should have some conditions and use those instead to determine readyness.

apis/rabbitmq/v1beta1/transporturl_types.go
```diff
@@ -0,0 +1,62 @@
...
+type TransportURLSpec struct {
+	// +kubebuilder:validation:Required
+	// RabbitmqClusterName the name of the Rabbitmq cluster which to configure the transport URL
+	RabbitmqClusterName string `json:"rabbitmqClusterName,omitempty"`
+}
+
+// TransportURLStatus defines the observed state of TransportURL
+type TransportURLStatus struct {
+	// SecretName - name of the secret containing the rabbitmq transport URL
+	SecretName string `json:"secretName,omitempty"`
```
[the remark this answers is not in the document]
dprince: I will add them

controllers/rabbitmq/transporturl_controller.go
gibizer: Thanks. It is OK for the short term.

controllers/rabbitmq/transporturl_controller.go
abays: I have found in testing this PR with an attempted Cinder integration that the labeling here doesn't seem to work out and the `RabbitMqCluster` is never found

apis/rabbitmq/v1beta1/transporturl_types.go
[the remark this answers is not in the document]
dprince: @gibizer Andrew is working on adding conditions in general to the openstack-operator. Could we hold on adding them and do them once we have basic condition support in.

controllers/rabbitmq/transporturl_controller.go
dprince: @abays thanks for the new logic.

pkg/rabbitmq/funcs.go
dprince: done. thanks.

controllers/rabbitmq/transporturl_controller.go
dprince: thanks @abays

config/samples/rabbitmq_v1beta1_transporturl.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: rabbitmq.openstack.org/v1beta1
+kind: TransportURL
+metadata:
+  # the name of this resource is used in the transport URL and to name the k8s secret
+  name: cinder
+spec:
+  rabbitmqClusterName: openstack
```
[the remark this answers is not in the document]
abays: That is true: `rabbitmq` is the cluster name in the `make openstack_deploy` context.  But I do wonder if we will even be using this sample YAML in `install_yamls`.

abays (commented):
+1 from me once @Akrog and @gibizer are good with it

apis/rabbitmq/v1beta1/transporturl_types.go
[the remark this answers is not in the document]
gibizer: Is TransportURL expected to be created by openstack-operator and not nova- and cinder-operator? If so then I'm OK to add condition to TransportURL when that condition is first used by the openstack-operator.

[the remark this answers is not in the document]
dprince: It will end up being similar to MariadbDatabase. We'll instantiate the creation of the TransportURL CR's from cinder/nova/neutron. Will likely have a lib-common function to help with this however.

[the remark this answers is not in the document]
gibizer: I see. Thanks for the explanation. I'm OK adding the condition later.

gibizer (approved):

apis/rabbitmq/v1beta1/transporturl_types.go
```diff
@@ -0,0 +1,62 @@
...
+limitations under the License.
+*/
+
+package v1beta1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
Akrog, line 23: I believe we should remove at least this line.

config/samples/rabbitmq_v1beta1_transporturl.yaml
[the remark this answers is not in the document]
Akrog: Even if we are not using I thought it would be good to have the example reflect the current openstack-operator deployment.

controllers/rabbitmq/transporturl_controller.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// Lookup RabbitmqCluster instance by name
+	labelSelector := map[string]string{
+		"namespace": instance.Namespace,
+		"name":      instance.Spec.RabbitmqClusterName,
+	}
+	rabbit, err := rabbitmq.GetRabbitmqCluster(ctx, helper, instance, labelSelector)
```
Akrog, line 111: Should we add a `TODO` to reflect that at some point we should actually watch the `rabbitmqclusters.rabbitmq.com` CRs so that if the change we can update the `transporturls.rabbitmq.openstack.org` accordingly.

main.go
```diff
@@ -59,6 +61,8 @@ func init() {
 	utilruntime.Must(placementv1.AddToScheme(scheme))
 	utilruntime.Must(glancev1.AddToScheme(scheme))
 	utilruntime.Must(cinderv1.AddToScheme(scheme))
+	utilruntime.Must(rabbitmqv1beta1.AddToScheme(scheme))
+	utilruntime.Must(cinderv1.AddToScheme(scheme))
```
Akrog, line 65: -1: This is a duplicate of L63, so we should only leave one of them, right?

Akrog (commented):
Code works great, I just have some comments that we may want to address now instead of having to submit a new PR for them in the future.

apis/rabbitmq/v1beta1/transporturl_types.go
dprince: Agreed. Will be done in the next rebase here in a few minutes.

main.go
dprince: Yes, will remove.

apis/rabbitmq/v1beta1/transporturl_types.go
```diff
@@ -0,0 +1,69 @@
...
+	Items           []TransportURL `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&TransportURL{}, &TransportURLList{})
+}
+
+// IsReady - returns true if service is ready to serve requests
+func (instance TransportURL) IsReady() bool {
+	return instance.Status.Conditions.IsTrue(TransportURLReadyCondition)
```
abays, line 68: Is there maybe a `conditions.go` that you meant to add that contained this constant?

dprince: ha, it was late and I forgot that one didn't I :)

controllers/rabbitmq/transporturl_controller.go
dprince: done

controllers/rabbitmq/transporturl_controller.go
```diff
@@ -0,0 +1,193 @@
...
+	rabbit, err := rabbitmq.GetRabbitmqCluster(ctx, helper, instance.Namespace, labelSelector)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Wait on RabbitmqCluster to be ready
+	rabbitReady := false
+	for _, condition := range rabbit.Status.Conditions {
+		if condition.Reason == "AllPodsAreReady" && condition.Status == "True" {
+			rabbitReady = true
```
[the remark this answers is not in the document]
dprince: done

controllers/rabbitmq/transporturl_controller.go
```diff
@@ -0,0 +1,270 @@
...
+			condition.SeverityInfo,
+			rabbitmqv1beta1.TransportURLReadyInitMessage))
+		return ctrl.Result{RequeueAfter: time.Second * 5}, nil
+	}
+
+	// Update the CR and return
+	instance.Status.SecretName = secret.Name
+	if err := r.Client.Status().Update(ctx, instance); err != nil {
+		return ctrl.Result{}, err
+	}
```
abays, line 243: I wonder if this is necessary with the `defer()` status func in the main reconcile loop?  It `Patch`es the status there, so it should catch and apply the `SecretName` change

dprince: good call, let me try it

controllers/rabbitmq/transporturl_controller.go
```diff
@@ -0,0 +1,267 @@
...
+		Data: map[string][]byte{
+			"transport_url": []byte(fmt.Sprintf("rabbit://%s:%s@%s:5672", username, password, host)),
+		},
+	}
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *TransportURLReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&rabbitmqv1beta1.TransportURL{}).
```
gibizer, line 265: I think we need `Owns(&Sercret{})` here too as this controller creates Secret objects.

pkg/rabbitmq/funcs.go
```diff
@@ -0,0 +1,41 @@
...
+	rabbitmqv1 "github.com/rabbitmq/cluster-operator/api/v1beta1"
+)
+
+//
+// GetRabbitmqCluster - get RabbitmqCluster object in namespace
+func GetRabbitmqCluster(
+	ctx context.Context,
+	h *helper.Helper,
+	instance *rabbitmqv1beta1.TransportURL,
+	labelSelector map[string]string,
```
gibizer, line 34: `labelSelector` seems to be unused

apis/rabbitmq/v1beta1/conditions.go
```diff
@@ -0,0 +1,54 @@
...
+	TransportURLReadyErrorMessage = "TransportURL error occured %s"
+
+	// TransportURLReadyInitMessage
+	TransportURLReadyInitMessage = "TransportURL not configured"
+
+	// TransportURLReadyMessage
+	TransportURLReadyMessage = "TransportURL completed"
+
+	// TransportURLReadyRunningMessage
+	TransportURLInProgressMessage = "TransportURL in progress"
```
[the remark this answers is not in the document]
dprince: fixed in latest commit

apis/rabbitmq/v1beta1/transporturl_types.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TransportURLSpec defines the desired state of TransportURL
+type TransportURLSpec struct {
+	// +kubebuilder:validation:Required
+	// RabbitmqClusterName the name of the Rabbitmq cluster which to configure the transport URL
+	RabbitmqClusterName string `json:"rabbitmqClusterName,omitempty"`
```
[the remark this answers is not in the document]
dprince: done

controllers/rabbitmq/transporturl_controller.go
dprince: good call its in the latest commit

pkg/rabbitmq/funcs.go
dprince: done

abays (approved):
/lgtm

PR #28, opened 2022-10-07
Add storageClass to sample CR's Glance template

abays:
We need to explicitly include `glanceTemplate.storageClass: ""` in the sample `OpenStackControlPlane` CR so that an upcoming change to `install_yamls` will be able to use `Kustomize` to replace the value with the value of an environment variable. This is needed to allow for flexibility in storage classes across CRC dev environments and non-CRC CI environments. The empty-string is chosen because this is what the CR already had implicitly for the `glanceTemplate.storageClass` field by excluding the field from the sample spec.

### DIFF
--- a/config/samples/core_v1beta1_openstackcontrolplane.yaml
+++ b/config/samples/core_v1beta1_openstackcontrolplane.yaml
@@ -24,4 +24,5 @@ spec:
     containerImage: quay.io/tripleotraincentos8/centos-binary-placement-api:current-tripleo
   glanceTemplate:
     containerImage: quay.io/tripleotraincentos8/centos-binary-glance-api:current-tripleo
+    storageClass: ""
     storageRequest: 10G


config/samples/core_v1beta1_openstackcontrolplane.yaml
```diff
@@ -24,4 +24,5 @@ spec:
     containerImage: quay.io/tripleotraincentos8/centos-binary-placement-api:current-tripleo
   glanceTemplate:
     containerImage: quay.io/tripleotraincentos8/centos-binary-glance-api:current-tripleo
+    storageClass: ""
```
fmount, line 27: Is this change required by PR#29 ?

abays: It's actually needed for an upcoming PR in `install_yamls` that will allow the user to specify `storageClass` (via a `STORAGE_CLASS` env var) for the `Glance` and `OpenStack` operators (and other operators if they expose `storageClass` as an option).  We're beginning to see issues with `storageClass` between dev and CI due to differences between CRC (dev) and non-CRC (CI).  With this field explicitly included in the sample CR, we can use `Kustomize` in `install_yamls` to `replace` the value as needed.

fmount: ack I see, thanks for clarifying, and the fact we extensively use `kustomize` justifies this change.
In theory if `storageClass` is not added an environment should use the default one, but then I saw issues in the past with `WaitForFirstConsumer` vs `Immediate` binding mode, so it would make sense adding this parameter at `openstack-operator` level and propagate it to the underlying operators according to the input definition.
This would also solve both `dev` and `CI` related problems.

dprince (approved):

PR #29, opened 2022-10-07
Propagate storageClass to Glance template

abays:
If the `glanceTemplate` in an `OpenStackControlPlane` CR lacks `storageClass`, set it to the `OpenStackControlPlane`'s `storageClass` (as we do for other embedded templates)

### DIFF
--- a/pkg/openstack/glance.go
+++ b/pkg/openstack/glance.go
@@ -32,6 +32,9 @@ func ReconcileGlance(ctx context.Context, instance *corev1beta1.OpenStackControl
 		if glance.Spec.DatabaseInstance == "" {
 			glance.Spec.DatabaseInstance = "openstack"
 		}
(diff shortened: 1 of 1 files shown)


pkg/openstack/glance.go
```diff
@@ -32,6 +32,9 @@ func ReconcileGlance(ctx context.Context, instance *corev1beta1.OpenStackControl
 		if glance.Spec.DatabaseInstance == "" {
 			glance.Spec.DatabaseInstance = "openstack"
 		}
+		if glance.Spec.StorageClass == "" {
+			glance.Spec.StorageClass = instance.Spec.StorageClass
```
fmount, line 36: This should also help to avoid issues when the `pvc` is created by the `glance-operator` during the bootstrap process

fmount (approved):

PR #30, opened 2022-10-11
Makefile: Support insecure registry

Akrog:
This patch adds `VERIFY_TLS` and `USE_HTTP` environmental variables to the Makefile to support using insecure registries both through https and http.

The `VERIFY_TLS=false` enables pushing (targets ending in -push) to https registries without verifying them as well as to http registries and also using `make catalog-build` without verifying https registries.

To use an HTTP registry on `make catalog-build` it is necessary to set `USE_HTTP=true` environmental variable.

This functionality is very convenient for local development without the need of an external registry, since it can use the `registry` container image to deploy a quick toy local registry.


abays (commented):
+1

Akrog:
Latest submission fixed the commit message that was saying `USE_TLS` instead of `VERIFY_TLS`

PR #36, opened 2022-10-19
Use ISO8601 time encoding for logging

jeckersb:


main.go
```diff
@@ -73,6 +75,7 @@ func main() {
 			"Enabling this will ensure there is only one active controller manager.")
 	opts := zap.Options{
 		Development: true,
+		TimeEncoder: zapcore.ISO8601TimeEncoder,
```
dprince, line 78: Fine by me. But for consistency I'm thinking we'll want all of our operators to do similar. So maybe we land this... but eventually move a common function into lib-common

dprince (approved):

PR #37, opened 2022-10-31
Use OSP17 RabbitMQ image by default instead of the upstream one

jeckersb:


pkg/openstack/rabbitmq.go
```diff
@@ -52,7 +74,7 @@ func ReconcileRabbitMQ(ctx context.Context, instance *corev1beta1.OpenStackContr
 		// operator to continuously mutate the CR when setting it:
 		// https://github.com/rabbitmq/cluster-operator/blob/main/controllers/reconcile_operator_defaults.go#L19
 		if rabbitmq.Spec.Image == "" {
-			rabbitmq.Spec.Image = "rabbitmq:3.10.2-management"
+			rabbitmq.Spec.Image = "registry.redhat.io/rhosp-rhel9/openstack-rabbitmq:17.0"
```
dprince, line 77: Eventually we'll want to make an ENV variable out of this and store the default in the CSV. But as I had this hard coded to begin with that can wait till later I think. No need to change it now.

pkg/openstack/rabbitmq.go
```diff
@@ -65,6 +87,13 @@ func ReconcileRabbitMQ(ctx context.Context, instance *corev1beta1.OpenStackContr
 			rabbitmq.Spec.Override.StatefulSet = &defaultStatefulSet
 		}
 
+		if rabbitmq.Spec.Rabbitmq.AdditionalConfig == "" {
+			helper.GetLogger().Info("Setting AdditionalConfig")
+			// This is the same situation as RABBITMQ_UPGRADE_LOG above,
+			// except for the "main" rabbitmq log we can just force it to use the console.
+			rabbitmq.Spec.Rabbitmq.AdditionalConfig = "log.console = true"
```
dprince, line 94: only question here is should we append this config option if for example the user supplies their own AdditionalConfig via the rabbitmqTemplate?

jeckersb: Yeah good point, I was torn about that.  It's mildly annoying that AdditionalConfig is just a plain string instead of something more structured, which makes it harder to parse and reason about.  I was worried about the case where the user supplied something like:

```
"foo = bar
log.console = false
log.syslog = true
# maybe even throw in a comment for good measure
baz = baz"
```

At that point we probably don't want to just blindly append `log.console = true`, we'd need to parse it and fully understand the contents and only set `log.console = true` if the user hasn't already supplied some other value.

In the end I went with the easier approach for now, with the (probably poor) assumption that if the user is overriding the default but they still want to keep the default, then it's up to them to include it in whatever config they supply.

pkg/openstack/rabbitmq.go
```diff
@@ -37,7 +38,28 @@ func ReconcileRabbitMQ(ctx context.Context, instance *corev1beta1.OpenStackContr
...
+									// the logs there.  The OSP image however has the directory more constrained,
+									// so the random uid cannot write the logs there.  Force it into /var/lib
+									// where it can create the file without crashing.
+									Name:  "RABBITMQ_UPGRADE_LOG",
+									Value: "/var/lib/rabbitmq/rabbitmq_upgrade.log",
+								},
+							},
+							Command: []string{
+								// The OSP17 image wants to start with kolla, override it to run rabbitmq directly.
+								// For some reason HOME needs to be explictly set here even though the entry
```
olliewalsh, line 57: Probably due to the entrypoint being set to dumb-init (https://github.com/openstack/tripleo-common/blob/master/container-images/tcib/base/base.yaml#L56)

jeckersb: Setting `Command` on the `Pod` overrides the `ENTRYPOINT` supplied in the image (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#container-v1-core).  So this version of the patch actually doesn't use dumb-init at all.  I also tried explicitly passing `-l` to `/bin/sh` to force a login shell, but that still does not populate `$HOME` correctly.

Ultimately I've reorganized this a bit (about to push a new version), split the `HOME` part out into the `EnvVar` array, and keeping the default entrypoint (dumb-init) but overriding the args to run `rabbitmq-server` instead of `kolla_start`.  I think that's cleaner and more idiomatic.

jeckersb:
Got linting working in my editor and per its complaining I removed unnecessary type annotations

dprince (approved):

PR #50, opened 2022-11-17
Check if our go.mod files are in sync

gibizer:
Both the main go.mod and the ./apis/go.mod file points to other operators. But if the version of those operators are not in sync between the two go.mod files then we can get strange CRD validation errors when the openstack-operator is deployed. To avoid that a check is added to the Makefile and to our github workflows

### DIFF
--- a/.github/workflows/go-mod-sync.yaml
+++ b/.github/workflows/go-mod-sync.yaml
@@ -1,0 +1,13 @@
+name: Golang lint, vet and unit test pipeline
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    name: go-mod-sync
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout project code
+        uses: actions/checkout@v2
+      - name: run verify-go-mod-sync.sh
+        run: ./verify-go-mod-sync.sh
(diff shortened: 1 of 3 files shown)


gibizer:
hm, the workflow should not pass as I intentionally based this on the older broken master....

dprince:
@gibizer I took a slightly different approach with https://github.com/openstack-k8s-operators/openstack-operator/pull/49. That and we'd have to automatically tag git commits as dependabot currently doesn't support pseudo versions. See also my comment on slack

gibizer:
> @gibizer I took a slightly different approach with #49. That and we'd have to automatically tag git commits as dependabot currently doesn't support pseudo versions. See also my comment on slack

Sure, your approach is the more forward looking one so I'm totally onboard to do that.

dprince:
> > @gibizer I took a slightly different approach with #49. That and we'd have to automatically tag git commits as dependabot currently doesn't support pseudo versions. See also my comment on slack
> 
> Sure, your approach is the more forward looking one so I'm totally onboard to do that.

To clarify I think the approaches are potentially both useful. Like having dependabot push the updates but our own workflow verify it isn't bad. It is possible a developer could manually get things out of sync and this would/should catch it immediately where as the dependabot automation is more on a schedule and may take days to show up.

In short we might still want to do both of these to guard against them getting out of sync.

PR #53, opened 2022-11-21
Allow deploying a subset of services

jistr:


jistr:
I tested it and it seems to work as intended, so i'm marking the PR as 'ready for review'.

dprince (approved):

PR #58, opened 2022-11-22
Add renovatebot config to update openstack deps

dprince:
This adds a workflow to run the renovate github action periodically (every 30 minutes).

The renovate config file is configured to update
openstack-k8s-operators as soon as an update is detected. And once a month it will attempt a general dependency update for the go.mod.

Additionally, a new pin-custom-bundle-dockerfile.sh is wired into the Makefiles 'bundle' target to pin
the respective openstack bundle images to the
commit tag referenced in go.mod for that operator.

This should help prevent our artifacts from getting out of sync with the :latest tagged bundles.

NOTE: Renovate supports a postUpgradeTasks but I choose note to use them as they require "self hosting".

### DIFF
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -1,0 +1,18 @@
+name: Run Renovate Bot
+
+on:
+  schedule:
+    - cron:  '0/30 * * * *'
+
+jobs:
+  renovate:
+    name: Renovatebot
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Self-hosted Renovate
+        uses: renovatebot/github-action@v34.29.2
+        with:
+          configurationFile: .github/renovate-config.js
+          token: ${{ secrets.RENOVATE_TOKEN }}
--- a/hack/pin-custom-bundle-dockerfile.sh
+++ b/hack/pin-custom-bundle-dockerfile.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+set -e
+cp custom-bundle.Dockerfile custom-bundle.Dockerfile.pinned
+#loop over each openstack-k8s-operators go.mod entry
+for X in $(go list -m -json all | jq -r '. | select(.Path | contains("openstack")) | "\(.Path)=\(.Version)"'); do
+  #example: github.com/openstack-k8s-operators/placement-operator/api=v0.0.0-20221007105015-13dce7450573
+  BASE=$(echo $X | sed -e 's|github.com/openstack-k8s-operators/\([^-\)]*\).*|\1|')
+  REF=$(echo $X | sed -e 's|github.com/[^\=]*=v0.0.0-[0-9]*-\(.*\)$|\1|')
+  #echo "BASE: $BASE REF: $REF"
+  if  ! echo "$BASE" | grep -e "lib" -e "openstack" &> /dev/null; then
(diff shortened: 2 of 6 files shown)


hack/pin-custom-bundle-dockerfile.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -e
+cp custom-bundle.Dockerfile custom-bundle.Dockerfile.pinned
+#loop over each openstack-k8s-operators go.mod entry
+for X in $(go list -m -json all | jq -r '. | select(.Path | contains("openstack")) | "\(.Path)=\(.Version)"'); do
```
bshephar, line 5: This looks like it should be fine, but running it locally on this repo I seem to be getting an error:
```
[fedora@fedora-work openstack-operator]$ go list -m -json all
go: github.com/openshift/api@v3.9.0+incompatible: invalid version: unknown revision v3.9.0
go: github.com/openshift/api@v3.9.0+incompatible: reading github.com/openshift/api/go.mod at revision v3.9.0: unknown revision v3.9.0
```

Is this a scenario we need to handle in this script?

bshephar (commented):
Nice initiative, should make life easier. Few comments in-line.

.github/workflows/renovate.yaml
```diff
@@ -0,0 +1,18 @@
+name: Run Renovate Bot
+
+on:
+  schedule:
+    - cron:  '0/30 * * * *'
```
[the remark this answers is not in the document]
dprince: Agree once a day is probably enough. A couple times a day won't hurt anything. It will only create the PRs once. Leaning towards running it a bit faster out of the gate in case we need/want to iterate on the config options more quickly at the beginning.

hack/pin-custom-bundle-dockerfile.sh
dprince: what 'go version' are you using locally? We require v1.18.x. I'm at go version go1.18.6 linux/amd64

.github/workflows/renovate.yaml
```diff
@@ -0,0 +1,18 @@
...
+    name: Renovatebot
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Self-hosted Renovate
+        uses: renovatebot/github-action@v34.29.2
+        with:
+          configurationFile: .github/renovate-config.js
+          token: ${{ secrets.RENOVATE_TOKEN }}
```
stuggi, line 18: how to create this RENOVATE_TOKEN ?

dprince: Well its a Github access token. I've already created one that has only the "public.repo" access. I was going to suggest we create a doc which contains all of our secrets and what they do.

.github/workflows/renovate.yaml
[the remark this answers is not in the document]
dprince: Bumped it down to once an hour

.github/workflows/renovate.yaml
stuggi: yes I think that'd be good. otherwise we might lose track

stuggi (approved):
/lgtm

hack/pin-custom-bundle-dockerfile.sh
bshephar: The one that doesn't work is Fedora with a slightly different minor version to you:
```
go version go1.18.7 linux/amd64
```

But interestingly, it works when I try it on my Mac which is:
```
❯ go version
go version go1.19.2 darwin/arm64
```

Not sure what the deal is there. But if we're happy that we should never need to handle that, we should be good here.

PR #73, opened 2022-11-25
Update crd with latest keystone/placement updates

karelyatin:
renovate bot updated keystone/placement[1][2] but didn't updated the OpenStackControlPlane CRD.

This patch updates the CRD and also the CR accordingly.

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/63
[2] https://github.com/openstack-k8s-operators/openstack-operator/pull/70

### DIFF
--- a/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
+++ b/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
@@ -152,24 +152,21 @@ spec:
                               for running the API service
                             type: object
                           passwordSelectors:
+                            default:
+                              database: CinderDatabasePassword
+                              service: CinderPassword
                             description: PasswordSelectors - Selectors to identify
-                              the DB and AdminUser password and TransportURL from
-                              the Secret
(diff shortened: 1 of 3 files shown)


gibizer:
I have no real opinion about the CRD doc generation part. But I totally agree that we need to sync the CRD yaml in repo with the CRD definition in the golang structs. Also I see the breakage even with the officially built bundle but there only because of the sample file is out of sync.

dprince:
As for RenovateBot it does support a postInstallTasks option that would allow us to run 'make manifests' along with each commit. That would be ideal however it requires it to be self hosted.

Pretty sure 'make bundle', etc runs make manifests automatically right? Is the breakage here the fact that install_yamls doesn't and would a simpler fix today be to have it do so? (basically run make manifests in install_yamls)

dprince:
So essentially the long term plan would be to have everything in the commit that syncs the repo deps using something like this: https://docs.renovatebot.com/configuration-options/#postupgradetasks

dprince (approved):

Issue #55, opened 2022-11-22
Make install fails due to annotation description length


bshephar:
**Problem**
When running `make install` the following error is shown:
```
[m3@osp-df-3 openstack-operator]$ make install
test -s /home/m3/openstack-operator/bin/controller-gen || GOBIN=/home/m3/openstack-operator/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.2
/home/m3/openstack-operator/bin/controller-gen rbac:roleName=manager-role crd:maxDescLen=0 crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases && \
    rm -f apis/bases/* && cp -a config/crd/bases apis/
/home/m3/openstack-operator/bin/kustomize build config/crd | kubectl apply -f -
customresourcedefinition.apiextensions.k8s.io/transporturls.rabbitmq.openstack.org unchanged
The CustomResourceDefinition "openstackcontrolplanes.core.openstack.org" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
make: *** [Makefile:138: install] Error 1
```

This is due to the description length for the annotation field in the CRD.

PR #74, opened 2022-11-25
Use crd:maxDescLen=0 only when required

karelyatin:
The workaround is required when applying CRDs in
the cluster but not during other operations like
make manifests, bundle, test etc.

Related: #55


bshephar:
Looks fine to me. We might want to revise this and give it a non-zero number at some point. 

We should merge this one, but maybe we can open another PR where we can debate an appropriate non-zero number for this option.

dprince (approved):

PR #75, opened 2022-11-25
Add OVN DBCluster and Northd to ControlPlane CR

karelyatin:


apis/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -139,6 +143,29 @@ type RabbitmqSection struct {
...
+	// +kubebuilder:validation:Optional
+	// Template - Overrides to use when creating the OVN services
+	Template OvnResources `json:"template,omitempty"`
+}
+
+// OvnResources defines the desired state of OVN services
+type OvnResources struct {
+	// +kubebuilder:validation:Optional
+	// OVNDBCluster - Overrides to use when creating the OVNDBCluster services
+	OVNDBCluster map[string]ovnv1.OVNDBClusterSpec `json:"DBCluster,omitempty"`
```
abays, line 162: The JSON name should match the Go field name (with the exception of exactly matching the case, of course).  This also applies for the `OVNNorthd` field below.

karelyatin: Thanks Andrew, sure i can update it. The reason i used like this was to avoid duplicating ovn in nested struct.
ovn:
  template:
     dbcluster

karelyatin: Done

PR #77, opened 2022-11-28
update container images to stable zed

SeanMooney:


stuggi (approved):
/lgtm

SeanMooney:
we may want to hold off merging this until glance fix there operator

SeanMooney:
its fine to use a mix of different service version but currently the glance operator is not correctly deploying i belive based on the testing gibi did breilfy in brno

that said i would like to make the switch over sooner rather then later as i would like to see this done before the end of year demo.

PR #82, opened 2022-11-28
Pre commit

dprince:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,68 @@
+ci:
+  autofix_prs: false
+  skip:
+   - 'make-manifests'
+   - 'make-generate'
+   - 'go-vet'
+   - 'golangci-lint'
+   - 'go-lint'
+   - 'make-operator-lint'
```
gibizer, line 9: if you don't want to use pre-commit.ci to run pre-commit through github then I think you can remove the whole ci section.

dprince: dropped it. thanks!

gibizer (approved):
looks good to me

PR #85, opened 2022-11-29
fix maxDescLen for deploy target

stuggi:


karelyatin (commented):
Thanks Martin for fixing it

PR #89, opened 2022-11-30
Integrate nova-operator

gibizer:


go.mod
```diff
@@ -93,3 +94,5 @@ require (
 )
 
 replace github.com/openstack-k8s-operators/openstack-operator/apis => ./apis
+
+replace github.com/openstack-k8s-operators/nova-operator/api => github.com/gibizer/nova-operator/api v0.0.0-20221129170244-4e3123df52f6
```
gibizer, line 98: This depends on https://github.com/openstack-k8s-operators/nova-operator/pull/168

apis/go.mod
```diff
@@ -74,3 +75,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/openstack-k8s-operators/nova-operator/api => github.com/gibizer/nova-operator/api v0.0.0-20221129170244-4e3123df52f6
```
gibizer, line 79: This depends on https://github.com/openstack-k8s-operators/nova-operator/pull/168

custom-bundle.Dockerfile
```diff
@@ -28,6 +28,7 @@ FROM quay.io/openstack-k8s-operators/rabbitmq-cluster-operator-bundle:latest as
 FROM quay.io/openstack-k8s-operators/placement-operator-bundle:latest as placement-bundle
 FROM quay.io/openstack-k8s-operators/glance-operator-bundle:latest as glance-bundle
 FROM quay.io/openstack-k8s-operators/cinder-operator-bundle:latest as cinder-bundle
+FROM quay.io/gibi/nova-operator-bundle:v0.0.182 as nova-bundle
```
gibizer, line 31: This is built from https://github.com/openstack-k8s-operators/nova-operator/pull/168

config/samples/core_v1beta1_openstackcontrolplane.yaml
```diff
@@ -54,3 +54,6 @@ spec:
         volume1:
           containerImage: quay.io/tripleozedcentos9/openstack-cinder-volume:current-tripleo
           replicas: 1
+  nova:
+    template:
+      secret: osp-secret
```
gibizer, line 59: Nova defaults is set in a way that this sample will deploy:
* nova-api
* cell0 with nova-conductor
* (eventually nova-scheduler (in progress))

This is enough to show that the control plane works. I.e services are up, controlplane only operations can be done, e.g. flavor create, keypair create.

The nova-operator also supports multiple cells computes in other than cell0 with
```
  secret: osp-secret
  cellTemplates:
    cell0:
      cellDatabaseUser: nova_cell0
      cellMessageBusInstance: rabbitmq
      hasAPIAccess: true
    cell1:
      cellDatabaseUser: nova_cell1
      cellMessageBusInstance: rabbitmq-cell1
      hasAPIAccess: true
```
But please note that cell0 and cell1 needs separate RabbitMQCluster instance (i.e. rabbitmq vs rabbitmq-cell1) otherwise nova-computes will not communicate properly.

pkg/openstack/nova.go
```diff
@@ -0,0 +1,98 @@
...
+		// Nova.Spec.APIDatabaseInstance and each NovaCell.CellDatabaseInstance
+		// aredefaulted to "openstack" in nova-operator and the MariaDB created
+		// by openstack-operator is also named "openstack". This works but
+		// in production we might want to have separate DB service instances
+		// per cell.
+		//
+		// 2)
+		// Each NovaCell.CellMessageBusInstance in defaulted to "rabbitmq" by
+		// nova-operator and openstack-operator creates RabbitMQCluster named
+		// "rabbitmq" as well. This is will not work as sharing rabbitmq
```
bshephar, line 59: Small type nit. Should read, "This will not work"

pkg/openstack/nova.go
```diff
@@ -0,0 +1,98 @@
...
+		// between cells will prevent the nvoa-computes to register itself
```
bshephar, line 60: Should be "nova-compute"

pkg/openstack/nova.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	helper.GetLogger().Info("Reconciling Nova", "Nova.Namespace", instance.Namespace, "Nova.Name", nova.Name)
+	op, err := controllerutil.CreateOrPatch(ctx, helper.GetClient(), nova, func() error {
+		// 1)
+		// Nova.Spec.APIDatabaseInstance and each NovaCell.CellDatabaseInstance
+		// aredefaulted to "openstack" in nova-operator and the MariaDB created
+		// by openstack-operator is also named "openstack". This works but
+		// in production we might want to have separate DB service instances
+		// per cell.
```
bshephar, line 54: Can we open a new issue for this point and link it here in this comment? That way we can discuss it and come up with a solution for this problem. If it remains here as a comment alone it might be overlooked for some time.

bshephar (commented):
Couple of typos in the comment that I have pointed out.

I know we have already been discussing the RabbitMQ vhosts problem in various meetings. I think we just need something to track the first point you raised in that comment around the Database names though. Either a new issue here on GitHub, or a Jira issue maybe? Something so that we don't forget to actually look into it. :).

pkg/openstack/nova.go
gibizer: I agree that we need to track these things but it is not clear how could I do it properly.
1. We have an Jira where I have ticket for this (OSP-20623). The ask was to create tracking tickets there. But that Jira is not public so I should not add that Jira ticket into this public repo.
2. I could create a github issue for this but then there would be a duplication between the internal Jira ticket and the external github action. Such duplication will lead to divergence which I dont like.

@dprince @stuggi : please advice what would be a good practice here?

gibizer:
> Couple of typos in the comment that I have pointed out.
> 
> I know we have already been discussing the RabbitMQ vhosts problem in various meetings. I think we just need something to track the first point you raised in that comment around the Database names though. Either a new issue here on GitHub, or a Jira issue maybe? Something so that we don't forget to actually look into it. :).

See my answer inline. I don't have a perfect way to track this publicly without duplication. (also for the multi rabbitmq support I have OSP-20624 internally)

pkg/openstack/nova.go
stuggi: As discussed on slack, lets bring this up in our next weeks meeting.

gibizer:
I built this PR into OPENSTACK_IMG=quay.io/gibi/openstack-operator-index:v0.0.14 and `make openstack` and `make openstack_deploy` successfully deployed nova too. openstack flavor create works. So this is ready to land from my perspective

abays (commented):
+1 from me

abays (approved):

PR #103, opened 2022-12-05
Configure renovate to run go mod tidy

dprince:
This should configure renovate to run go mod tidy before updating a PR.


abays (approved):
/lgtm

PR #110, opened 2022-12-07
Pin operator-lint to v0.1.0

gibizer:
Originally I pointed the Makefile to @latest but that make the version uncontrollable and potentially operator-lint can break the repo with a new release.

So lets pin the operator-lint version.


abays (approved):
/lgtm

PR #112, opened 2022-12-07
Controlled OsCtlPlane teardown

abays:
We are leaving certain child resources behind when we `oc delete osctlplane`.  We currently rely on `ownerRefs` and automatic k8s cleanup processes to remove these child resources.  This does not work in the scenario of `oc delete osctlplane` because of race conditions that then occur (if certain resources are removed before others, the latter get orphaned because they need the former to be present in order for them to be removed).  This patch introduces one approach to controlling the teardown of the control plane to better ensure that everything is removed.

Depends-on: https://github.com/openstack-k8s-operators/keystone-operator/pull/135


controllers/core/openstackcontrolplane_controller.go
```diff
@@ -110,6 +112,25 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
...
 	}
 
+	helper, err := helper.NewHelper(
+		instance,
+		r.Client,
+		r.Kclient,
+		r.Scheme,
+		r.Log,
+	)
+	if err != nil {
```
bogdando, line 122: nit: logging something for a reconcile failure before returning nil is a good practice

abays: The error is already logged because it is returned, but we can add additional context around it

pkg/openstack/common.go
```diff
@@ -0,0 +1,37 @@
...
+				}
+			}
+
+			found = true
+		}
+	}
+
+	if found {
+		res = ctrl.Result{RequeueAfter: time.Duration(5) * time.Second}
+		helper.GetLogger().Info(fmt.Sprintf("OpenStackControlPlane %s deletion waiting for deletion of %s %s", instance.Name, subRes.GetObjectKind().GroupVersionKind().Kind, subRes.GetName()))
```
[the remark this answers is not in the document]
abays: Yes, I need to change all of these

pkg/openstack/cinder.go
```diff
@@ -75,3 +75,26 @@ func ReconcileCinder(ctx context.Context, instance *corev1beta1.OpenStackControl
...
 
 }
+
+// DeleteCinder -
+func DeleteCinder(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, helper *helper.Helper) (ctrl.Result, error) {
+	overallRes := ctrl.Result{}
+
+	cinderList := &cinderv1.CinderList{}
+
+	if err := helper.GetClient().List(context.Background(), cinderList); err != nil {
```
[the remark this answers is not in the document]
abays: No, that was a mistake here and elsewhere.  Thanks for pointing these out.

pkg/openstack/common.go
```diff
@@ -0,0 +1,37 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// checkDeleteSubresource -
+func checkDeleteSubresource(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, helper *helper.Helper, subRes client.Object) (ctrl.Result, error) {
+	found := false
+	res := ctrl.Result{}
+
+	for _, ownerRef := range subRes.GetOwnerReferences() {
+		if ownerRef.Kind == instance.Kind && ownerRef.Name == instance.Name {
```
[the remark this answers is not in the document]
abays: We could do that.  It's a longer path to get to the actual `Kind` string (`helper.GetBeforeObject().GetObjectKind().GroupVersionKind().Kind`), but I like the idea of cleaner signatures for the funcs.

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -245,6 +260,59 @@ func (r *OpenStackControlPlaneReconciler) reconcileNormal(ctx context.Context, i
...
+		}
+	}
+
+	// Everything is cleared, so remove the finalizer so this OpenStackControlPlane can be fully removed
+	controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
+	if err := r.Update(ctx, instance); err != nil && !k8s_errors.IsNotFound(err) {
+		return overallRes, err
+	}
+
+	r.Log.Info("Reconciled Service delete successfully")
```
dprince, line 311: Should this say 'ControlPlane' instead of service (sort of a collection of services

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -245,6 +260,59 @@ func (r *OpenStackControlPlaneReconciler) reconcileNormal(ctx context.Context, i
...
+			&rabbitmqv1.RabbitmqClusterList{},
+			&keystonev1.KeystoneAPIList{},
+		},
+		{
+			&mariadbv1.MariaDBList{},
+		},
+	}
+
+	for _, step := range steps {
+		if overallRes, err = openstack.CheckDeleteSubresources(ctx, helper, step); err != nil {
```
dprince, line 294: Since we are checking all subresources here does that imply we could be "lazy" and only rely on setting the ownership of subresources in the service operators to do actual cleanup. Versus adding implicit finalizers in each operator and having logic in each of the service operators similar to this.

pkg/openstack/common.go
```diff
@@ -0,0 +1,82 @@
...
+		for _, ownerRef := range obj.GetOwnerReferences() {
+			if ownerRef.Kind == ctlPlaneObj.GetObjectKind().GroupVersionKind().Kind && ownerRef.Name == ctlPlaneObj.GetName() {
+				if obj.GetDeletionTimestamp().IsZero() {
+					if err := helper.GetClient().Delete(ctx, obj); err != nil {
+						return err
+					}
+				}
+
+				found = true
+				util.LogForObject(helper, fmt.Sprintf("OpenStackControlPlane %s deletion waiting for deletion of %s %s", ctlPlaneObj.GetName(), obj.GetObjectKind().GroupVersionKind().Kind, obj.GetName()), ctlPlaneObj)
```
bogdando, line 60: nit: these 3 can be stored in local vars which saves calling it twice

controllers/core/openstackcontrolplane_controller.go
abays: Yeah, we can do something more specific.  I will change it.

controllers/core/openstackcontrolplane_controller.go
dprince: we spoke on slack. what I'm suggesting here would require perhaps extra fancy logic to guard against some of the cross/operator teardown steps. Like if mariadb disappears first it would cause some resources to hang and they would need to realize mariadb was gone and skip the finalizer. I think what we have here is fine.

dprince (approved):

PR #116, opened 2022-12-11
renovate: group all dependencies into a single PR

dprince:


renovate.json
```diff
@@ -27,8 +28,7 @@
       "excludePackagePatterns": [
         "^github.com/openstack-k8s-operators"
```
stuggi, line 29: not sure I understand this excludePackagePatterns and the above matchPackageNames. Doesn't the above include it, but here we exclude it?

dprince: I've reworked this now to be explicit with the packages we are disabling.

renovate.json
```diff
@@ -2,33 +2,32 @@
...
-      ],
-      "extends": ["schedule:monthly"],
-      "groupName": "deps"
+      "matchPackagePatterns": ["^k8s.io", "^sigs.k8s.io", "^github.com/rabbitmq/cluster-operator", "^github.com/operator-framework/api"],
+      "enabled": false
     }
-  ]
+  ],
+  "postUpgradeTasks": {
+    "commands": ["go mod tidy", "make manifests"],
```
bogdando, line 29: plus make generate?

renovate.json
```diff
@@ -2,33 +2,32 @@
...
   "extends": [
-    "config:base"
+    "config:base",
+    "group:all"
   ],
-  "dependencyDashboard": true,
-  "includeForks": true,
   "logFileLevel": "trace",
   "enabledManagers": ["gomod"],
   "postUpdateOptions": ["gomodTidy"],
```
gibizer, line 10: I would add
```
"constraints": {
    "go": "1.18"
  },
```
to keep bumps compatible with or 1.18 limit.

renovate.json
```diff
@@ -2,33 +2,32 @@
   "timezone": "America/New_York",
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:base"
+    "config:base",
+    "group:all"
```
gibizer, line 6: yeah that `group:all` seems unnecessary. I ended up with the following grouping in nova:

```
"packageRules": [
    {
      "matchPackageNames": ["github.com/openstack-k8s-operators/nova-operator/api"],
      "enabled": false
    },
    {
      "groupName": "openstack-k8s-operators",
      "matchPackagePrefixes": ["github.com/openstack-k8s-operators"],
      "excludePackageNames": ["github.com/openstack-k8s-operators/nova-operator/api"],
      "schedule": [
        "every weekend"
      ]
    },
    {
      "groupName": "k8s.io",
      "matchPackagePrefixes": [
        "k8s.io",
        "sigs.k8s.io"
      ],
      "schedule": [
        "every weekend"
      ],
      "allowedVersions": "< 1.0.0"
    }
  ]
```
* The first rule disables updating nova/api in nova
* The second rule creates a group for all the openstack-k8s-operators dependency. (excluding nova/api again)
* The third rule groups the k8s.io deps and pin them < 1.0.0 (see the versioning rule in their repo readme, it is changed)
This means the rest will be independent bumps but for nova that is only gomega and ginkgo which we do want to handle independently anyhow.

renovate.json
```diff
@@ -2,33 +2,32 @@
...
-      "extends": ["schedule:monthly"],
-      "groupName": "deps"
+      "matchPackagePatterns": ["^k8s.io", "^sigs.k8s.io", "^github.com/rabbitmq/cluster-operator", "^github.com/operator-framework/api"],
+      "enabled": false
     }
-  ]
+  ],
+  "postUpgradeTasks": {
+    "commands": ["go mod tidy", "make manifests"],
+    "fileFilters": ["go.mod", "go.sum", "**/*.go", "**/*.yaml"],
```
gibizer, line 30: why we need this filters?

renovate.json
gibizer: Also I would add `"dependencyDashboard": true,` to get pinned issue collecting all the pending bumps in a single place.

gibizer (changes requested):

renovate.json
dprince: I can re-add the dashboard. Fwiw I removed it but it is still getting created so it doesn't seem to matter: https://github.com/dprince/openstack-operator/issues/15

renovate.json
dprince: thanks, added

renovate.json
dprince: hmmm, pretty sure I tried without 'group: all' and it left me with lots of PR's for each operator.

renovate.json
gibizer: ahh so it is not a filter to ignore things :)

gibizer (approved):
looks good. thanks.

PR #121, opened 2022-12-12
Introduce ExtraVolumes support

fmount:
This patch introduces the ExtraVolumes support in openstack-operator: ExtraVolumes is the feature that provides the ability to propagate the top level defined extra volumes to the underlying operators. It relies on the propagation policy defined in the lib-common/storage module.

Signed-off-by: Francesco Pantano <fpantano@redhat.com>


pkg/openstack/cinder.go
```diff
@@ -41,6 +41,20 @@ func ReconcileCinder(ctx context.Context, instance *corev1beta1.OpenStackControl
 			//cinder.Spec.DatabaseInstance = instance.Name // name of MariaDB we create here
 			cinder.Spec.DatabaseInstance = "openstack" //FIXME: see above
 		}
+		// if already defined in the top-level crd, we don't override
```
stuggi, line 44: I am not sure if this comment reflects the code, or I miss-read it. isn't it that if you haven't specified it a specific volume list in the cinder template it will use the global?

pkg/openstack/cinder.go
```diff
@@ -41,6 +41,20 @@ func ReconcileCinder(ctx context.Context, instance *corev1beta1.OpenStackControl
...
+		if len(cinder.Spec.ExtraMounts) == 0 {
```
stuggi, line 45: wouldn't it better to merge cinder.Spec.ExtraMounts and instance.Spec.ExtraMounts, so that you can have global ones and per service extra volumes? if there are duplicate entries the service one should probably overwrite the global.

stuggi: or have  a parameter to control if those should be merged

pkg/openstack/cinder.go
fmount: Yes, I'll clarify this comment in my next push: the design choice is that if you define `extravolumes` in the template section we don't merge them with the top-level defined extra volumes, hence that `if statement` is needed to entirely skip the part where we merged `cinder.Spec.ExtraMounts` with `instance.Spec.ExtraMounts`.
We can change the approach in a follow up, but for now let's try to move forward with this change.

pkg/openstack/cinder.go
fmount: see comment above

apis/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -82,6 +83,10 @@ type OpenStackControlPlaneSpec struct {
 	// +kubebuilder:validation:Optional
 	// Nova - Parameters related to the Nova services
 	Nova NovaSection `json:"nova,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// ExtraMounts containing conf files and credentials
```
gibizer, line 88: Please document that his will be ignored if the per service level extraMounts is defined.

gibizer (dismissed):

apis/core/v1beta1/openstackcontrolplane_types.go
fmount: Done

gibizer (commented):
Thank you. This looks good to me.

stuggi (commented):
looks good to me as well. lets get this in if @dprince is also good with it

dprince (approved):

PR #130, opened 2022-12-21
workflow rename, disable cron

dprince:
Renames the release workflow to 'OpenStack Operator'

Also, Disables cron builds on the build. Now that we are pinning this should no longer be required.


stuggi (approved):
/lgtm

PR #132, opened 2023-01-04
Add missing Cinder status considerations

abays:
Now that Cinder is using `TransportURL`, we can add the missing Cinder status checks to the overall `OsControlPlane` status


stuggi (approved):
/lgtm

PR #134, opened 2023-01-10
Use PatchInstance instead of Update()

dprince:
This avoids changing the generation ID during reconciliation which caused issues in integrating the galera-operator


abays (approved):
/lgtm

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -125,11 +125,22 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, err
 	}
 
+	// Always patch the instance status when exiting this function so we can persist any changes.
+	defer func() {
+		// update the overall status condition if service is ready
+		if instance.IsReady() {
```
bogdando, line 131: a beautiful example of hiding low-level implementation details :)

PR #136, opened 2023-01-16
Initial AnsibleEE work

fao89:
closes [OSP-20072](https://issues.redhat.com/browse/OSP-20072)

Signed-off-by: Fabricio Aguiar <fabricio.aguiar@gmail.com>

### DIFF
--- a/controllers/core/openstackcontrolplane_controller.go
+++ b/controllers/core/openstackcontrolplane_controller.go
@@ -85,6 +85,7 @@ type OpenStackControlPlaneReconciler struct {
 //+kubebuilder:rbac:groups=nova.openstack.org,resources=nova,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=mariadb.openstack.org,resources=mariadbs,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=neutron.openstack.org,resources=neutronapis,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=ansibleee.openstack.org,resources=openstackansibleees,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=ovn.openstack.org,resources=ovndbclusters,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=ovn.openstack.org,resources=ovnnorthds,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=ovs.openstack.org,resources=ovs,verbs=get;list;watch;create;update;patch;delete
--- a/custom-bundle.Dockerfile
+++ b/custom-bundle.Dockerfile
@@ -31,6 +31,7 @@ FROM quay.io/openstack-k8s-operators/cinder-operator-bundle:latest as cinder-bun
 FROM quay.io/openstack-k8s-operators/ovn-operator-bundle:latest as ovn-bundle
 FROM quay.io/openstack-k8s-operators/ovs-operator-bundle:latest as ovs-bundle
 FROM quay.io/openstack-k8s-operators/neutron-operator-bundle:latest as neutron-bundle
+FROM quay.io/openstack-k8s-operators/openstack-ansibleee-operator-bundle:latest as openstack-ansibleee-bundle
 FROM quay.io/openstack-k8s-operators/nova-operator-bundle:latest as nova-bundle
 
 FROM golang:1.18 as merger
@@ -50,6 +51,7 @@ COPY --from=cinder-bundle /manifests/* /manifests/
 COPY --from=ovn-bundle /manifests/* /manifests/
 COPY --from=ovs-bundle /manifests/* /manifests/
 COPY --from=neutron-bundle /manifests/* /manifests/
+COPY --from=openstack-ansibleee-bundle /manifests/* /manifests/
 COPY --from=nova-bundle /manifests/* /manifests/
 
 RUN /workspace/csv-merger \
@@ -62,6 +64,7 @@ RUN /workspace/csv-merger \
   --ovn-csv=/manifests/ovn-operator.clusterserviceversion.yaml \
   --ovs-csv=/manifests/ovs-operator.clusterserviceversion.yaml \
   --neutron-csv=/manifests/neutron-operator.clusterserviceversion.yaml \
+  --ansibleee-csv=/manifests/openstack-ansibleee-operator.clusterserviceversion.yaml \
   --nova-csv=/manifests/nova-operator.clusterserviceversion.yaml \
   --openstack-csv=/manifests/openstack-operator.clusterserviceversion.yaml | tee /openstack-operator.clusterserviceversion.yaml.new
 
--- a/main.go
+++ b/main.go
@@ -32,6 +32,7 @@ import (
 	mariadbv1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
 	neutronv1 "github.com/openstack-k8s-operators/neutron-operator/api/v1beta1"
 	novav1 "github.com/openstack-k8s-operators/nova-operator/api/v1beta1"
+	ansibleeev1 "github.com/openstack-k8s-operators/openstack-ansibleee-operator/api/v1alpha1"
 	ovnv1 "github.com/openstack-k8s-operators/ovn-operator/api/v1beta1"
 	ovsv1 "github.com/openstack-k8s-operators/ovs-operator/api/v1beta1"
 	placementv1 "github.com/openstack-k8s-operators/placement-operator/api/v1beta1"
@@ -72,6 +73,7 @@ func init() {
 	utilruntime.Must(ovnv1.AddToScheme(scheme))
 	utilruntime.Must(ovsv1.AddToScheme(scheme))
 	utilruntime.Must(neutronv1.AddToScheme(scheme))
+	utilruntime.Must(ansibleeev1.AddToScheme(scheme))
 	//+kubebuilder:scaffold:scheme
 }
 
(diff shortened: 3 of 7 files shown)


controllers/core/openstackcontrolplane_controller.go
```diff
@@ -85,6 +86,7 @@ type OpenStackControlPlaneReconciler struct {
 //+kubebuilder:rbac:groups=nova.openstack.org,resources=nova,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=mariadb.openstack.org,resources=mariadbs,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=neutron.openstack.org,resources=neutronapis,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=redhat.com,resources=openstackansibleees,verbs=get;list;watch;create;update;patch;delete
```
fao89, line 89: @jlarriba @slagle Ansible EE group `redhat.com` while the group for the other operators are `openstack.org`

should we change it to follow the pattern?

pkg/openstack/ansibleee.go
```diff
@@ -0,0 +1,65 @@
...
+			condition.SeverityWarning,
+			corev1beta1.OpenStackControlPlaneAnsibleEEReadyErrorMessage,
+			err.Error()))
+		return ctrl.Result{}, err
+	}
+	if op != controllerutil.OperationResultNone {
+		helper.GetLogger().Info(fmt.Sprintf("AnsibleEE %s - %s", ansibleEE.Name, op))
+	}
+
+	// TODO: if ansibleEE.IsReady() {
```
fao89, line 53: @jlarriba @slagle
the other operators implement `IsReady^, should we implement it on Ansible EE?

fmount: @fao89 I'm not sure about the purpose of `IsReady` in this context: in general that function is used to check the Status of the pods deployed by the operator and `Mark` the `Status` with a given message.
With `AnsibleEE`, you're building and running `Jobs`: they can be tracked during the execution with an `UID`, but at the end the job will be automatically set to `Completed`.
Not saying we don't need that kind of function, just not sure about the purpose.

pkg/openstack/ansibleee.go
```diff
@@ -0,0 +1,65 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	ansibleeev1 "github.com/openstack-k8s-operators/openstack-ansibleee-operator/api/v1alpha1"
+	corev1beta1 "github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// ReconcileAnsibleEE -
+func ReconcileAnsibleEE(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, helper *helper.Helper) (ctrl.Result, error) {
```
fmount, line 19: Not sure I follow here: I remember we tried something similar in a POC with [1], but we changed approach as the `openstack-operator` is not supposed to run this operator.
@gibizer @dprince I guess the question here is: which component is supposed to build the `DataPlane` CR and run it? Should be the `nova-operator`, or meta-operator is supposed to watch for that CR triggering an `AnsibleEE` job when it's found?



[1] https://github.com/openstack-k8s-operators/openstack-operator/compare/master...fmount:openstack-operator:ansible_operator_support

controllers/core/openstackcontrolplane_controller.go
jlarriba: Yes

pkg/openstack/ansibleee.go
jlarriba: I agree with @fmount

pkg/openstack/ansibleee.go
gibizer: I'm not sure what design discussions happened around this space since the Brno meetup. What I recall from there is that we need a place that can collect the data that needed to deploy the external node and synchronize / order the ansible execution requests coming from different parts of the control plane. At least I can foresee that nova and neutron will contribute such data. Nova for the nova-compute service neutron for the neutron nic switch agent for SRIOV. So I don't think that simply exposing the OpenStackAnsibleEESpec via the openstack-operator is the right abstraction here. 

During Brno there was a CRD draft added https://issues.redhat.com/browse/OSP-20001 with such an abstraction. I think we should pick up that work. I heard that James Slagle might pick that work up from EDPM perspective but I'm not sure.

controllers/core/openstackcontrolplane_controller.go
fao89: https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/56

pkg/openstack/ansibleee.go
slagle: Yes, I agree with @fmount / @gibizer here. The purpose of this PR should be to just make sure that openstack-ansibleee-operator is installed and running in the cluster when openstack-operator is installed.

fao89: could you all please take another look?

fmount:
Technically it looks good, the scope of this PR was reduces to just include the `AnsibleEE` in the `openstack-operator` and make it "consumable" under certain conditions.
For this reason the change itself looks good to me, but I'd like other people (@slagle  mostly, who is working on the `DataPlane` POC/prototype) to validate.

fmount (approved):

apis/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -95,6 +100,18 @@ type OpenStackControlPlaneSpec struct {
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=true
+	// Enabled - Whether OpenStackAnsibleEE service should be deployed and managed
+	Enabled bool `json:"enabled,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// Template - Overrides to use when creating the OpenStackAnsibleEE service
+	Template ansibleeev1.OpenStackAnsibleEESpec `json:"template,omitempty"`
+}
+
```
gibizer, line 114: Based on the scope:

> The purpose of this PR should be to just make sure that openstack-ansibleee-operator is installed and running in the cluster when openstack-operator is installed.

I don't think we need to change the API of the OpenStackControlPlane CR

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -270,6 +272,7 @@ func (r *OpenStackControlPlaneReconciler) reconcileDelete(ctx context.Context, i
 			&ovnv1.OVNDBClusterList{},
 			&ovnv1.OVNNorthdList{},
 			&ovsv1.OVSList{},
+			&ansibleeev1.OpenStackAnsibleEEList{},
```
gibizer, line 275: Based on the scope we don't need this.

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -319,5 +322,6 @@ func (r *OpenStackControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) err
 		Owns(&ovsv1.OVS{}).
 		Owns(&neutronv1.NeutronAPI{}).
 		Owns(&novav1.Nova{}).
+		Owns(&ansibleeev1.OpenStackAnsibleEE{}).
```
gibizer, line 325: ditto

custom-bundle.Dockerfile
```diff
@@ -62,6 +64,7 @@ RUN /workspace/csv-merger \
   --ovn-csv=/manifests/ovn-operator.clusterserviceversion.yaml \
   --ovs-csv=/manifests/ovs-operator.clusterserviceversion.yaml \
   --neutron-csv=/manifests/neutron-operator.clusterserviceversion.yaml \
+  --ansibleee-csv=/manifests/openstack-ansibleee-operator.clusterserviceversion.yaml \
```
gibizer, line 67: this (and the csv-merger.go change) is the meat of the change.

main.go
```diff
@@ -72,6 +73,7 @@ func init() {
 	utilruntime.Must(ovnv1.AddToScheme(scheme))
 	utilruntime.Must(ovsv1.AddToScheme(scheme))
 	utilruntime.Must(neutronv1.AddToScheme(scheme))
+	utilruntime.Must(ansibleeev1.AddToScheme(scheme))
```
gibizer, line 76: we don't need this.

gibizer (changes requested):

main.go
fao89: I'm keeping it for now because yesterday @slagle mentioned I needed to modify `main.go`

gibizer (commented):
looks good to me

gibizer (approved):

main.go
slagle: I think we were just trying to copy the existing patterns we see in the operator. @gibizer can you explain why this wouldn't be needed? How does it differs from the other listed operators?

slagle (approved):

main.go
gibizer: I think (but I could be wrong here) that if the controller code does not try to read / write any AnsibleEE CR instance then we don't need to load the schema either.

abays (commented):
So the intent is to get the AnsibleEE operator pulled into this operator without actually having it be deployed yet?

fao89:
> So the intent is to get the AnsibleEE operator pulled into this operator without actually having it be deployed yet?

yes!

abays (approved):
/lgtm

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -85,6 +85,7 @@ type OpenStackControlPlaneReconciler struct {
 //+kubebuilder:rbac:groups=nova.openstack.org,resources=nova,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=mariadb.openstack.org,resources=mariadbs,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=neutron.openstack.org,resources=neutronapis,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=ansibleee.openstack.org,resources=openstackansibleees,verbs=get;list;watch;create;update;patch;delete
```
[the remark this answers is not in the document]
fao89: I still not sure if we will have an openstackdataplane_controller, @slagle is working in a prototype

PR #142, opened 2023-01-19
Use inclusive language for branch naming

bshephar:
In line with Red Hat's commitment to using inclusive language in our products. This change adds the necessary modifications to switch the default branch to use "main".

### DIFF
--- a/controllers/core/openstackcontrolplane_controller.go
+++ b/controllers/core/openstackcontrolplane_controller.go
@@ -27,7 +27,6 @@ import (
 	keystonev1 "github.com/openstack-k8s-operators/keystone-operator/api/v1beta1"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
-	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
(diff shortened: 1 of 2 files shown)


bshephar:
Branch renaming can be done as described here:
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch

dprince:
Project is inclusive! But do we have to do this.... seems like a lot of disruption for not much gain to me.

Does RH intent to pepper this same thing across all of OpenShift too?

bshephar:
We have publicly stated that we plan to do this:
https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/17.0/html/release_notes/making-open-source-more-inclusive

But from a practicality perspective, we have some of these operators using a main branch (newer ones) and some using a master branch(older ones from CNOSP). So for consistency sake, we need to change some of them anyway. It might as well be the ones using master? Then we're in-line with a "making open-source more inclusive" statement and we have nice consistent branch naming.

Happy to leave it up to the general consensus though. But that's my thoughts on the topic.

dprince:
Most important thing to me is if we do it, we do it everywhere. And agree also that doing it sooner is better than later. But it will break some CI, scripts, and cause general developer disruptions. A slight nit is it isn't the GIT default. But folks can correct that with a local config setting I guess.

PR #143, opened 2023-01-19
Add pin hack to bundle-build target

rabi:
We use custom-bundle.Dockerfile.pinned in bundle-build. It won't exist without running bundle target atm.


Makefile
```diff
@@ -194,6 +194,7 @@ bundle: manifests kustomize ## Generate bundle manifests and metadata, then vali
 
 .PHONY: bundle-build
 bundle-build: ## Build the bundle image.
+	/bin/bash hack/pin-custom-bundle-dockerfile.sh
```
stuggi, line 197: should we just run bundle target as a dependency?

rabi: I did think about that, which would be different from all other operator Makefiles. Also, we do have a default custom-bundle.Dockerfile that can be used (just the hack needed I guess)

https://github.com/openstack-k8s-operators/openstack-operator/blob/master/custom-bundle.Dockerfile

stuggi: the openstack-operator is special, so I guess we'll have differences to other operators. I am ok with both ways. either add the script as in this patch, or run bundle as a dependency. Lets see what @dprince think

dprince: @rabi good catch. I think think running the entire 'bundle' target as a dependency is probably better here for now.

As a side note I would also add that I'm considering calling hack/pin-custom-bundle-dockerfile.sh when we bump our dependencies here: https://github.com/openstack-k8s-operators/openstack-operator/blob/master/renovate.json#L45 Essentially doing it this way would eliminate the need for dynamic pinning and just making everything static in the single bundle.Dockerfile

rabi: OK, I would add the dep to fix the broken target for the time being.

stuggi (approved):
/lgtm

PR #153, opened 2023-01-20
Skip already customized operator bundle images

viroel:
This patch adds a condition to verify if a operator bundle image needs to be updated or not.
The following scenario is broken in current version of the script, and is the workflow that will be used on CI:
1. A PR to *-operator is proposed
2. Job runs and build operator, bundle and catalog images.
3. Images are pushed to CI registry.
4. openstack-operator go.mod is updated with new api reference
5. custom-bundle.Dockerfile is modified to point to CI operator bundle image
6. Job builds openstack-operator images, and push to CI registry
7. Deploy steps validates openstack-operator image

When generating openstack-operator bundle, it fails to find the bundle image on user's quay.io account.

The new condition ignores bundle images that were already modified.

### DIFF
--- a/hack/pin-custom-bundle-dockerfile.sh
+++ b/hack/pin-custom-bundle-dockerfile.sh
@@ -11,6 +11,12 @@ for MOD_PATH in $(go list -m -json all | jq -r '. | select(.Path | contains("ope
 
   BASE=$(echo $MOD_PATH | sed -e 's|github.com/.*/\(.*\)-operator/.*|\1|')
 
+  SRC_BUNDLE=("FROM quay.io/openstack-k8s-operators/${BASE}-operator-bundle.*")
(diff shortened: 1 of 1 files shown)


viroel:
FYI, where I am replacing operator bundle image, before build openstack-operator[1].
It fails when github user doesn't have the image built in its quay registry. 
In this case, CI is building and pushing to RDO registry.

[1] https://github.com/openshift/release/pull/35553/files#diff-f535e0344a0aeb7b9ebdd1d5273e0c372d24a40e397d1f1ea78a45bba82338d8R140-R144

viroel:
/assign @dprince 
Hi @dprince @stuggi - we need a way to avoid replacing image bundle path in bundle.Dockerfile when the api is set to a GH user commit. This is the scenario that CI will cover, since CI will build and push images to its own registry. We are manually doing the replacement in job steps.
Can you take a look on this PR?

dprince (approved):

PR #155, opened 2023-01-23
Honor enabled: false for services during update/patch

rabi:
We currently don't cleanup resources when enabled flag is toggeled. Though there would dependencies for services to be enabled (we would implement with validation webhooks) in subsequent patches, this implements basic cleanup to honor the flag in the CRD.

Also, adds scaffolding for the validation webhook.


main.go
```diff
@@ -146,6 +147,12 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "TransportURL")
 		os.Exit(1)
 	}
+	if strings.ToLower(os.Getenv("ENABLE_WEBHOOKS")) != "false" {
+		if err = (&corev1beta1.OpenStackControlPlane{}).SetupWebhookWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create webhook", "webhook", "OpenStackControlPlane")
+			os.Exit(1)
+		}
+	}
```
abays, line 155: Just a heads-up regarding webhooks and running the operator locally.  There is a scenario where webhooks can complicate dev envs.  Sometimes for dev purposes we want to install the operator via OLM and then remove the actual `openstack-operator-controller-manager` deployment from the CSV, but leave the other children controller-managers (i.e Glance, Cinder, Keystone, etc) running while we iterate on the `openstack-operator-controller-manager` and use `GOWORK= OPERATOR_TEMPLATES=./templates make run` to run the manager locally.  In such a case, the `ValidatingWebhookConfiguration` installed by OLM will cause errors to occur whenever you try to create/update `OpenStackControlPlane` CRs, like so:

```
oc kustomize /home/abays/rh/install_yamls/out/openstack/openstack/cr | oc apply -f -
Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "vopenstackcontrolplane.kb.io": failed to call webhook: Post "https://openstack-operator-controller-manager-service.openstack.svc:443/validate-core-openstack-org-v1beta1-openstackcontrolplane?timeout=10s": dial tcp 10.217.0.102:9443: connect: connection refused
```

This is because the webhook server isn't running in the local process (as per `ENABLE_WEBHOOKS=false`).  Leaving the `ENABLE_WEBHOOKS=false` env var out of course results in a different error because the cert for the web server can't be found outside of an OLM deployment (but there might be a way to work around that).

Anyhow, I just wanted to mention this.  We need to start using webhooks though, so it's just something we'll have to deal with going forward.

abays: I suppose another solution is to perhaps patch the CSV to remove the webhook definition.

rabi: @abays Thanks for checking the PR. I've not been able to get the webhook work either in CI/locally(crc). It failed in CI[1] with the same error. I deployed cert-manager operator locally (missing in CI I guess) and could also see webhook configuration is fine[2], but it seems apiserver can't talk to openstack-operator-controller-manager(webhook) for some reason.

It looks like some routing issue rather than cert validation. Not sure what I'm missing. Do we test webhooks with osp-director-operator?

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_openstack-operator/155/pull-ci-openstack-k8s-operators-openstack-operator-master-openstack-operator-build-deploy/1617716507368755200/build-log.txt

[2] 

$ oc get ValidatingWebhookConfiguration/vopenstackcontrolplane.kb.io-65ffw -o yaml
....
webhooks:
- admissionReviewVersions:
  - v1
  clientConfig:
    caBundle: <removed>
    service:
      name: openstack-operator-controller-manager-service
      namespace: openstack
      path: /validate-core-openstack-org-v1beta1-openstackcontrolplane
      port: 443

$oc get svc/openstack-operator-webhook-service -o yaml
....
spec:
  clusterIP: 10.217.5.30
  clusterIPs:
  - 10.217.5.30
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - port: 443
    protocol: TCP
    targetPort: 9443
  selector:
    control-plane: controller-manager
  sessionAffinity: None
  type: ClusterIP

abays: Is the webhook working for you when you deploy via OLM, however?  Or is it not working ever at all?

rabi: It's not working with OLM bundles too. I linked the CI openstack-operator-build-deploy job where it failed with the same error. I just worked around in CI with the latest patchset by setting failurePolicy: Ingore, as it's just scaffolding atm.

abays: Ah okay.  We definitely have webhooks working in OSPdO if you want to consult that as a reference, but the state of the operator-SDK support for webhooks was different back then (so what we have there is not necessarily the "right" way to do it anymore).  I did make a pass at a mutating webhook for the Glance operator a while back using the new operator-SDK approach and it was working for me: https://github.com/openstack-k8s-operators/glance-operator/pull/52.  Even though it's a mutating webhook, the basic infrastructure is mostly the same as validating webhooks.  Maybe something in that PR can help provide clues.  I can also try your PR via OLM and see if I can find the problem.

rabi: I checked the glance-operator PR and did not find any notable difference. Probably the operator-sdk is doing something wrong which I'm not able to trace. The error is weird to decipher. Initially cert-manager operator was missing on my setup and I did deploy that. Maybe some iptable rule missing or incorrectly configured on the nodes. If you can try the PR and find the issue that would be great:)

rabi: So the issue was "selector labels" for the controller manager/webhook services. It's the same for all operators and the calls are routed to controller pods that don't have the webhook running (load-balanced across all endpoints with the same lebel). It worked at times when it would end up with the openstack-operator-controller. After changing the lables for openstack-operator-controller in this PR, webhook works fine.

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
+  newName: quay.io/rabi/openstack-operator
```
abays, line 15: `s/rabi/openstack-k8s-operators`.  Or, instead, we can add this file to `.gitignore` as the newer versions of Operator SDK set by default during scaffolding: https://github.com/openstack-k8s-operators/openstack-baremetal-operator/blob/6df055e5341bef713e051364e7b38b34b5162bec/.gitignore#L23

pkg/openstack/common.go
```diff
@@ -7,12 +7,32 @@ import (
...
+	// Delete the object
+	if obj.GetDeletionTimestamp().IsZero() {
+		if err := helper.GetClient().Delete(ctx, obj); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+	return ctrl.Result{}, nil
+
+}
+
```
abays, line 35: There's a lot of talk about enabling/disabling services going on in another active PR [1] for this operator.  Perhaps at this point we need to reconcile what is happening there with these "disable-and-delete" logic changes here. @bshephar @gibizer thoughts?

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/162

config/manager/kustomization.yaml
rabi: Yeah, I have it in .gitignore now, not sure how it made it to the PR earlier.

pkg/openstack/common.go
rabi: I guess that one is meant to update the 'status', as disabling services leads to the status of openstackcontrolplane instance not updated to "Setup complete" and won't change with this.

pkg/openstack/common.go
```diff
@@ -7,12 +7,32 @@ import (
...
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// CheckDeleteDisabled - Delete the object which in turn will clean the sub resources
+func CheckDeleteDisabled(ctx context.Context, helper *helper.Helper, obj client.Object) (ctrl.Result, error) {
```
gibizer, line 18: I would not call it `Check` as it actively deletes not just checks for a condition. It also does not check for Disabled. So I would call this simply `Delete()`

pkg/openstack/common.go
```diff
@@ -7,12 +7,32 @@ import (
...
+	key := client.ObjectKeyFromObject(obj)
+	if err := helper.GetClient().Get(ctx, key, obj); err != nil {
+		if k8s_errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+	// Delete the object
+	if obj.GetDeletionTimestamp().IsZero() {
```
gibizer, line 27: Why we need the timestamp check here? Is it here to avoid queueing up multiple delete requests for an object that is already being deleted?

pkg/openstack/common.go
```diff
@@ -7,12 +7,32 @@ import (
...
+		if err := helper.GetClient().Delete(ctx, obj); err != nil {
+			return ctrl.Result{}, err
```
gibizer, line 29: `err` can be NotFound here as well (I think) but we should ignore that error.

pkg/openstack/common.go
```diff
@@ -7,12 +7,32 @@ import (
...
+		}
+	}
+	return ctrl.Result{}, nil
```
gibizer, line 32: Here we know that the delete succeeded. I think we should remove the service specific condition from the OpenStackControlPlane.Status.Condition list not to leave stale status about this deleted service there.

pkg/openstack/common.go
gibizer: > need to reconcile

I start to feel like an k8s operator :)

Thanks @abays for spotting the parallel tracks. 

I think we can adjust this patch to to solve the problem in the other PR as well. See above by comment about removing the service specific condition from the condition list when the service is deleted.

pkg/openstack/cinder.go
```diff
@@ -17,17 +17,19 @@ import (
...
 			Name:      "cinder",
 			Namespace: instance.Namespace,
 		},
 	}
 
+	if !instance.Spec.Cinder.Enabled {
+		if res, err := CheckDeleteDisabled(ctx, helper, cinder); err != nil {
+			return res, err
+		}
+	}
```
gibizer, line 31: If the service is disabled then we delete it at L28 but then here we continue and reach the CreateOrPatch call at L34 that will recreate the service. I think we either have a service enabled and then proceed to L34 or disabled and disabled and only do `CheckDeleteDisabled` but not proceed to L34.

pkg/openstack/common.go
abays: It seems like there is some common ground between these two PRs, specifically with regards to one of the latest comments from @gibizer:

```
I can see one solution that solves all problems:

    Solution1: When a service is disabled then delete the service CR and after that succeeded, delete the service specific condition from OpenStackControlPlane.Status.Condition.
```

The whole comment [1] can provide more context.  It seems like some people might have different opinions as to what should happen to a service when it is disabled.

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/162#issuecomment-1408224224

pkg/openstack/keystone.go
```diff
@@ -17,17 +17,19 @@ import (
...
-
 	keystoneAPI := &keystonev1.KeystoneAPI{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "keystone", //FIXME (keystone doesn't seem to work unless named "keystone")
 			Namespace: instance.Namespace,
 		},
 	}
 
+	if !instance.Spec.Keystone.Enabled {
+		if res, err := CheckDeleteDisabled(ctx, helper, keystoneAPI); err != nil {
```
gibizer, line 28: Deleting Keystone while still having other services using it (like cinder) will not work as KeystoneService and KeystoneEndpoint CRs will be blocked by the service finalizers. I think this is why we have the staged delete logic in https://github.com/openstack-k8s-operators/openstack-operator/blob/f2d183dd6292dcca4050e59ca3d7bd6cc5534545/controllers/core/openstackcontrolplane_controller.go#L263-L282 So I think we need some check here too to avoid this situation.

gibizer (changes requested):

pkg/openstack/common.go
rabi: Yeah I can see what you mean now.

pkg/openstack/common.go
rabi: Yes

pkg/openstack/common.go
rabi: I can add that, but we just checked that it exists before calling delete.

pkg/openstack/common.go
rabi: Yeah, that sounds reasonable. I did not think about it earlier.

pkg/openstack/cinder.go
rabi: Looks like we need to return if there is no error.  But strangely this works fine with my testing with no re-creation of service.  I'll recheck it.

pkg/openstack/keystone.go
rabi: IIUC you're talking about service dependencies. I think we should do that in the validation webhook and keep the lifecycle logic simple for services. I guess the commit message talks about it.

pkg/openstack/common.go
gibizer: Thanks. It probably won't hurt. :)

pkg/openstack/common.go
gibizer: as L20 and L28 is two separate API requests I believe it can happen that the first sees the object but second does not see it any more. It is probably a scenario with low probability but having a check for NotFound here is also cheap. I would add that check.

pkg/openstack/keystone.go
gibizer: Yepp it is about dependencies between the services. I.e. keystone cannot be deleted while nova is actively using KeystoneService CRs. 

Hm. So a validation webhook can reject a keystone disable request if nova is enabled. That part is good. And I guess that will be implemented in `ValidateServices` above. 

What if an single update is coming with enabled=False for both keystone and nova (and assuming no other keystone dependencies is enabled)? Will it be rejected with an error than nova needs to be disabled first? Or will it be allowed? 

If the latter then the edge case, where the update request contains enabled=False for all the services is logically equivalent for deleting all sub resources during the OpenStackControl plane delete. So I think the two delete scenario ( 1) disable all services 2) delete OpenStackControlPlane CR) has the same dependency handling needs. So I'm wondering if we could reuse parts of `reconcileDelete` for the disable service code path too.

Hm, even if we do the former then two update requests can come in closely after each other (nova disabled first then keystone) making it possible that nova is not deleted (even if it is already disabled) when keystone is disabled.

rabi: > What if an single update is coming with enabled=False for both keystone and nova 
That would be allowed by the webhook.

> where the update request contains enabled=False for all the services
That's a simple scenario where we would just allow it in the webhook without checking dependencies. We can make the validations as simple/complicated as we want.

> Hm, even if we do the former then two update requests can come in closely after each other
I assume the requests would be queued one after the other for the OpenStackControlPlane CR and would be handled accordingly and we should not have any issues right?

pkg/openstack/common.go
rabi: I have pushed a PR[1] to remove a condition in lib-common to use here.

[1] https://github.com/openstack-k8s-operators/lib-common/pull/156

pkg/openstack/common.go
rabi: Yeah, different API requests but they are sequential and Get is not asynchronous like delete. I can add it though.

pkg/openstack/keystone.go
gibizer: > > where the update request contains enabled=False for all the services
> 
> That's a simple scenario where we would just allow it in the webhook without checking dependencies. We can make the validations as simple/complicated as we want.

OK, but that would mean delete all the subresources. Which is very similar to the case when OpenStackControlPlane is deleted. So either the code in `reconcileDelete` is over-complicated for that scenario or the the currently proposed independent per service delete is too simplistic for it. @abays you authored the controller deletion of subresources in `reconcileDelete` so I guess you have a better view on it.

rabi: > OK, but that would mean delete all the subresources

I don't think we need to. To me it looks like CheckDeleteSubresources is doing things we don't need to and can be simplified.  reconcileDelete of service instances would called when you make a delete api call and that should be sufficient.

abays: Currently, you can delete (or, I suppose, disable) any of these [1] resources in any order you please.  What's important in the teardown logic, however, is that Keystone [2] is then removed only _after_ those resources have been deleted.  And finally, after Keystone and the rest are gone, you are free to remove MariaDB [3].  If you don't at least follow the order of [1] and then [2], the finalizers on the Keystone resources will prevent tear-down from completing.  

Keep in mind that this logic referenced here (and in above comments) is only for the _complete_ teardown of the `OpenStackControlPlane`.  Deleting individual services would be different, but any attempt to disable the Keystone service (which would entail deleting `KeystoneAPI`, `KeystoneService`, `KeystoneEndpoints`, etc, I am assuming) would be blocked by the finalizers on the Keystone service if other OpenStack services exist and have placed finalizers on the Keystone service resources.  And of course disabling and deleting the MariaDB service will torpedo everything.  For whatever reason MariaDB currently does not have the same level of dependency integrity protection via finalizers that Keystone does (it has them on `MariaDBDatabase`, but not on `MariaDB`, which can be deleted and hose everything).

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/534f09255ae43a7e7c48e0d118d81d37bed50380/controllers/core/openstackcontrolplane_controller.go#L267-L274
[2] https://github.com/openstack-k8s-operators/openstack-operator/blob/534f09255ae43a7e7c48e0d118d81d37bed50380/controllers/core/openstackcontrolplane_controller.go#L278
[3] https://github.com/openstack-k8s-operators/openstack-operator/blob/534f09255ae43a7e7c48e0d118d81d37bed50380/controllers/core/openstackcontrolplane_controller.go#L281

rabi: > Deleting individual services would be different, but any attempt to disable the Keystone service (which would entail deleting KeystoneAPI, KeystoneService, KeystoneEndpoints, etc, I am assuming) would be blocked by the finalizers on the Keystone service if other OpenStack services exist and have placed finalizers on the Keystone service resources.

Yeah looks like we add a finalizer[1] on the KeystoneAPI for the service instance and that won't let the KeystoneAPI  disable/delete attempt to succeed (would wait) unless the dependent service instances are disabled/deleted.  We probably need similar level of protection with MariaDB/MariaDBDatabase. I'll see if I can add that.

And you're right we need to do them in ordered fashion as you've implemented when tearing down OpenStackControlPlane instance, else api call to delete service instances won't return and the whole teardown would get stuck. 

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/master/controllers/keystoneservice_controller.go#L181-L190

rabi: I've proposed a patch go add finalizes[1] on MariaDB for services, so that it can't be deleted before the services. It It It would also require rbac changes in all operators. I've patch for keystone-operator[2] already proposed.

[1] https://github.com/openstack-k8s-operators/lib-common/pull/165
[2] https://github.com/openstack-k8s-operators/keystone-operator/pull/186

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -112,6 +112,8 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 		// Error reading the object - requeue the request.
 		return ctrl.Result{}, err
 	}
+	// Remove/Reset the ready condition if exists
+	instance.Status.Conditions.Remove(condition.ReadyCondition)
```
gibizer, line 116: So I guess we need this as now ReadyCondition can go from True to False when a new service is enabled. I'm wondering that it would be better to move this to the deferred function where we set the ReadyCondition to True. So the manipulation of the ReadyCondition would be in one place. We either set that to true there if IsReady() true or to false if IsReady() is false.

main.go
```diff
@@ -148,6 +149,12 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "TransportURL")
 		os.Exit(1)
 	}
+	if strings.ToLower(os.Getenv("ENABLE_WEBHOOKS")) != "false" {
```
gibizer, line 152: https://pkg.go.dev/strconv#ParseBool would be a more user friendly approach to allow multiple bool like string values

pkg/openstack/cinder.go
```diff
@@ -17,17 +17,21 @@ import (
...
 			Name:      "cinder",
 			Namespace: instance.Namespace,
 		},
 	}
 
+	if !instance.Spec.Cinder.Enabled {
+		if res, err := CheckDeleteDisabled(ctx, helper, cinder); err != nil {
+			return res, err
+		}
+		instance.Status.Conditions.Remove(corev1beta1.OpenStackControlPlaneCinderReadyCondition)
```
gibizer, line 31: +1

I think this allow us to make IsReady() simpler see my comment there.

apis/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -274,10 +274,18 @@ func init() {
...
+		(!instance.Spec.Ovn.Enabled ||
+			instance.Status.Conditions.IsTrue(OpenStackControlPlaneOVNReadyCondition)) &&
+		(!instance.Spec.Keystone.Enabled ||
+			instance.Status.Conditions.IsTrue(OpenStackControlPlaneKeystoneAPIReadyCondition)) &&
+		(!instance.Spec.Placement.Enabled ||
+			instance.Status.Conditions.IsTrue(OpenStackControlPlanePlacementAPIReadyCondition)) &&
+		(!instance.Spec.Glance.Enabled ||
+			instance.Status.Conditions.IsTrue(OpenStackControlPlaneGlanceReadyCondition)) &&
+		(!instance.Spec.Cinder.Enabled ||
+			instance.Status.Conditions.IsTrue(OpenStackControlPlaneCinderReadyCondition))
```
gibizer, line 290: What if we define this function as True if all the conditions in the list (except ReadyCondition) is true? 
As 
* the operator now removes the conditions of disabled services 
* initialize all the conditions to Unknown automatically
* all the conditions use True to report success

I think the below code has the same result with less complexity.

```golang
	for _, c := range instance.Status.Conditions {
		if c.Type == condition.ReadyCondition {
			continue
		}
		if c.Status != corev1.ConditionTrue {
			return false
		}
	}
	return true
```

pkg/openstack/common.go
gibizer: `ensureDeleted` would work for me to encode that this function works if the service is already deleted. Still the function does not depend on the service being disabled so I would not add that to the name. But we can agree to disagree here.

apis/core/v1beta1/openstackcontrolplane_types.go
rabi: > What if we define this function as True if all the conditions in the list (except ReadyCondition) is true?

Is it not possible that the condition is not added to Status.Conditions inspite of the service being enabled? If that's not possible then we can simplify it. But I thought to keep it explicit as before. 

> initialize all the conditions to Unknown automatically
Why should we do that, unless the the service has to be reconciled?

> all the conditions use True to report success

controllers/core/openstackcontrolplane_controller.go
rabi: Currently ReadyCondition once true is left as True unless there is some error in the subsequent reconciliation. So this resets the condition at the beginning.

pkg/openstack/cinder.go
abays: Could we set this back to `Unknown` as an alternative to removing it outright?

rabi: I think "removing the ready status" when the service does not exist/disabled, rather than setting it to 'Unknown' makes more sense. Also, we can look at implementing it in many different ways, but I would probably get something in that's working and then followup with whatever optimization we want to do.

apis/core/v1beta1/openstackcontrolplane_types.go
rabi: > I think the reasoning to do that is to let the use know what are the steps still needed to reach the Ready status.

IMO, initial reconciliation is different where the *ReadyCondition could be set to 'Unknown'. I don't think we should set all service conditions 'Unknown' for before every reconciliation. If the service is to be reconciled then only their status would change.

pkg/openstack/cinder.go
gibizer: Yeah I agree with @rabi. A service condition with Unknown here means the service needs to be deployed but the deploy hasn't been started yet. When the service is disabled we should not really add the service condition to the list as it is misleading.

apis/core/v1beta1/openstackcontrolplane_types.go
gibizer: Yeah it is messy. 

We have to differentiate between two cases when we say setting a service condition to Unknown.
1) Setting it temporarily at the beginning of the reconcile and letting the actual run of the Reconcile function update that Unknown state based on the current state of the service. 
2) Setting it to Unknown at the beginning *and persisting that Unknown to the API* via Update or Patch.

I agree that we should not do 2) for each Reconcile run as that will cause service conditions to flip between states rapidly. But I believe we can do 1) and let the Reconcile run do the recalculation each time and if that recalculation result in the same condition state as before persisted then the deferred Patch call will be no-op anyhow.

abays: Option 1 sounds reasonable

hack/create_self_signed_cert.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+set -ex
+
+TMPDIR=${TMPDIR:-"/tmp/k8s-webhook-server/serving-certs"}
+SERVICE=${SERVICE:-"openstackcontrolplane"}
```
dprince, line 5: Should this just be 'openstack' (for the openstack-operator). As we will have other CRD here as well.

Or do we instead need a certificate per CRD within the operator?

config/manager/manager.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
   labels:
-    control-plane: controller-manager
+    control-plane: openstack-controller-manager
```
dprince, line 5: I thought the idea was we could use this to search for all the operator controllers. If each operator has an individual name here wouldn't that change our use case here?

dprince: okay. I see in the commit message now this is strictly required to make webhooks work across all operators. we'll need to adopt similar label updates across the board then

main.go
dprince: @gibizer that may be but as we already have prior art for this code in our operators we should probably call stuff like this out in a separate issue right?

config/manager/manager.yaml
dprince: Also see this was discussed on the glance-operator. so ignore me here

apis/core/v1beta1/openstackcontrolplane_types.go
rabi: A few additional things to consider if we do option 1.

- if you initialize status temporarily without persisting for every reconcile, what conditional  would you use for the initial feedback[1] for first reconcile (currently it checks `instance.Status.Conditions == nil`)?

- Are we going to change how we update/change status in every operator[2]?

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/controllers/core/openstackcontrolplane_controller.go#L170
[2] https://github.com/openstack-k8s-operators/glance-operator/blob/master/controllers/glance_controller.go#L147

main.go
gibizer: I can file a separate issue for it as it is definitely not a bug just an improvement. (However I start to feel that we never have the time in the future to go and work on those improvement stories.)

apis/core/v1beta1/openstackcontrolplane_types.go
gibizer: > A few additional things to consider if we do option 1.
> 
>     * if you initialize status temporarily without persisting for every reconcile, what conditional  would you use for the initial feedback[1] for first reconcile (currently it checks `instance.Status.Conditions == nil`)?

I think we can use == nil still for the initial feedback as that will be only true until the end of the first Reconcile. In the rest of the reconciles we never observer == nil and do reinit of the Conditions without returning right away and persisting it.

> 
>     * Are we going to change how we update/change status in every operator[2]?

I think this case is special as setting enable=false in openstack-operator for a service the conditions for that service becomes invalid and we want to remove that condition. While in other operator I don't see similar cases. There each condition is added at start and has a meaning throughout of the existence of the CR.

> 
> 
> [1] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/controllers/core/openstackcontrolplane_controller.go#L170 [2] https://github.com/openstack-k8s-operators/glance-operator/blob/master/controllers/glance_controller.go#L147

rabi: > I think we can use == nil still for the initial feedback as that will be only true until the end of the first Reconcile

That would make the code flow not so nice.  (check status empty, update the status and return, else update status and proceed). Anyways.

> I think this case is special as setting enable=false in openstack-operator for a service the conditions for that service becomes invalid and we want to remove that condition.

I don't see it as different. We're creating a bunch of resources (standard vs custom) and updating the status. What we're saying is for the openstackctlplane CR, status of every resource it's composed of is unknown in the beginning of reconcile. It can also apply to others too.

gibizer: I rest my case here. I think we went into a rabbit hole (sorry for that) and it is becoming unproductive. I will stop pushing for changes in the implementation itself and I will try to focus on having an acceptable external behavior. On that point my only goal would be not to have misleading information in the Status.Conditions after a service is disabled. If the current implementation has that behavior then I'm OK with it.

pkg/openstack/ovn.go
```diff
@@ -28,6 +25,14 @@ func ReconcileOVN(ctx context.Context, instance *corev1beta1.OpenStackControlPla
 			},
 		}
 
+		if !instance.Spec.Ovn.Enabled {
+			if res, err := CheckDeleteDisabled(ctx, helper, OVNDBCluster); err != nil {
+				return res, err
+			}
+			instance.Status.Conditions.Remove(corev1beta1.OpenStackControlPlaneOVNReadyCondition)
+			return ctrl.Result{}, nil
```
abays, line 33: This should be `continue` instead of a `return`.  Otherwise other `OVNDBCluster` and `OVNNorthD` resources will not be deleted when OVN is disabled.

Makefile
```diff
@@ -115,8 +115,11 @@ build: generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
 	go build -o bin/csv-merger cmd/csv-merger/csv-merger.go
 
+generate-cert:
+	/bin/bash hack/create_self_signed_cert.sh
+
 .PHONY: run
-run: manifests generate fmt vet ## Run a controller from your host.
+run: manifests generate fmt vet generate-cert ## Run a controller from your host.
```
dprince, line 122: could we just leave generate-cert as a standalone target and not wire it into the 'manifests' target. Reason I ask is this may break the self-hosted Renovate job (which runs a 'make manifests' when it updates the PRs). I don't think we need/want to generate certs when creating the Renovate PRs basically.

dprince (changes requested):

Makefile
dprince: AH, nevermind. I misread this. It is the 'run' task this is wired into. That is probably fine I think.

apis/core/v1beta1/openstackcontrolplane_types.go
rabi: @gibizer Appreciate your review and feedback. We're not going into a rabbit hole. We can disagree on certain things and still can find a compromise to move forward. My point was to have a common pattern  across operators. Also, I feel we can work and merge things iteratively. Anyway, I've updated the patch as you wanted. We can revisit it if required later.

pkg/openstack/ovn.go
rabi: Nice catch, thanks!

abays (commented):
+1 from me

dprince (approved):

PR #159, opened 2023-01-23
[sample]Cleanup cell separation

gibizer:
Now that nova switched to cell separation by default the cell separation sample file can be cleaned up to contain only the necessary information and default the rest.

Depends-On: https://github.com/openstack-k8s-operators/nova-operator/pull/242 (merged)
Depends-On: #158 (merged)


gibizer:
I prefer #182 over this. So marking this as draft

gibizer:
#182 merged instead.

PR #161, opened 2023-01-24
Add IMAGEREGISTRY var to pin-custom-bundle-dockerfile

viroel:
This patch adds IMAGEREGISTRY var to pin-custom-bundle-dockerfile.sh to solve the issue with user's image available in a different registry and namespace. IMAGENAMESPACE already allows us to customize namespace.
With this new var we don´t need the previous workaround[1] to skip customized content added to custom-bundle.Dockerfile.

[1] https://github.com/openstack-k8s-operators/openstack-operator/commit/dff94253689ca6b3f77db8d0dbbd9d3d5eb6afbf


stuggi (approved):
/lgtm

PR #162, opened 2023-01-25
Don't set status or reconcile disabled services

bshephar:
We have some optional services that don't necessarily be enabled to deploy a cluster. For example Glance and Cinder. Currently, both of these will try to update the status even if they aren't enabled.

This ultimately prevents us from reporting a successful deployment. As such, this patch ensures that the statuses and reconciliation of these objects are only called if the service is indeed enabled.


bshephar:
This patch specifically addresses the issue I raised here:
https://issues.redhat.com/browse/OSP-21711

However, we may want to look at the discussion about some kind of data structure that will hold a slice of enabled services. Such a data structure might be able to simply this implementation as we would be able to iterate over the list and just call reconcile and status updates for those services without multiple conditional `if instance.Spec.Blah.Enabled`.

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -248,6 +245,16 @@ func (r *OpenStackControlPlaneReconciler) reconcileNormal(ctx context.Context, i
 		return ctrlResult, nil
 	}
 
+	// Reconcile optional services when enabled. If the service is explicitly disabled,
+	// we shouldn't try to reconcile it.
+	if instance.Spec.Cinder.Enabled {
```
kajinamit, line 250: The same is already done in ReconcileCinder function. If we implement this in this top-level then we probably have get the check of Enabled out from individual functions.

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -159,10 +159,14 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
...
 			condition.UnknownCondition(corev1beta1.OpenStackControlPlanePlacementAPIReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlanePlacementAPIReadyInitMessage),
 			condition.UnknownCondition(corev1beta1.OpenStackControlPlaneGlanceReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlaneGlanceReadyInitMessage),
-			condition.UnknownCondition(corev1beta1.OpenStackControlPlaneCinderReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlaneCinderReadyInitMessage),
 			condition.UnknownCondition(corev1beta1.OpenStackControlPlaneNovaReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlaneNovaReadyInitMessage),
 		)
 
+		// Initialize optional service conditions if enabled
+		if instance.Spec.Cinder.Enabled {
+			cl = append(cl, *condition.UnknownCondition(corev1beta1.OpenStackControlPlaneCinderReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlaneCinderReadyInitMessage))
+		}
```
kajinamit, line 168: I'm not quite familiar with the standard but does it make sense to present the condition with "disabled" instead of not exposing it at all ?

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -159,10 +159,14 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 			condition.UnknownCondition(corev1beta1.OpenStackControlPlaneKeystoneAPIReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlaneKeystoneAPIReadyInitMessage),
 			condition.UnknownCondition(corev1beta1.OpenStackControlPlanePlacementAPIReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlanePlacementAPIReadyInitMessage),
 			condition.UnknownCondition(corev1beta1.OpenStackControlPlaneGlanceReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlaneGlanceReadyInitMessage),
-			condition.UnknownCondition(corev1beta1.OpenStackControlPlaneCinderReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlaneCinderReadyInitMessage),
 			condition.UnknownCondition(corev1beta1.OpenStackControlPlaneNovaReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlaneNovaReadyInitMessage),
 		)
 
+		// Initialize optional service conditions if enabled
+		if instance.Spec.Cinder.Enabled {
```
kajinamit, line 166: I've seen a few users not using even nova (just for keystone and swift, for example). Also, some users would want to disable ovs/ovn when they use 3rd party SDN.

So I'm wondering whether it makes sense to make sure all services can be disabled (with some dependency validation as webhook ?)

kajinamit (changes requested):

controllers/core/openstackcontrolplane_controller.go
bshephar: I think my preference would be to just leave it out of the status entirely. Few reasons for that, but mainly because the purpose of the status is to report the outcome of our reconcile attempts. Since we shouldn't really try to reconcile a disabled service at all, there's not really a reason to include that service in the status report. That's at least my opinion on the decision to not include it.

controllers/core/openstackcontrolplane_controller.go
bshephar: Do we anticipate a minimum number of services deployed to constitute a viable, supportable OSP deployment? If so, maybe we can just group core and optional services for now. Until we have a better system for managing this.

I would ultimately like to stop handling this on a service-by-service basis and implement something a little more dynamic. But that is still a work in progress for me while I figure out the best implementation to propose.

controllers/core/openstackcontrolplane_controller.go
bshephar: Yeah, true. I should add that to this PR as well.

controllers/core/openstackcontrolplane_controller.go
kajinamit: Fair enough. Let me keep this open in case others want to chime in

controllers/core/openstackcontrolplane_controller.go
kajinamit: > Do we anticipate a minimum number of services deployed to constitute a viable, supportable OSP deployment?

In case we support a use case with only swift and keystone then probably the core would be only keystone, mariadb and memcached ?
I'd rather avoid defining "the must items" strictly and leave the selections to users. That would be more consistent with what we expose now (all services have enabled/disabled flag)

kajinamit (commented):
Thanks for updating. I want to add feedback by others about the remaining two comments, so will keep these two open now. These would not be a strong blocker if other folks agree with the current implementation.

controllers/core/openstackcontrolplane_controller.go
stuggi: I don't think I have a preference for this, at least at the moment. Just wondering if we should have the same across all the services.

controllers/core/openstackcontrolplane_controller.go
stuggi: I agree, we should probably have the same for all the services as they have enabled/disabled parameter

controllers/core/openstackcontrolplane_controller.go
stuggi, line 250: this is basically not really required, right? the current approach would do the same. and just have the condition to initialize the unknown state if enabled, or have I missed something?

pkg/openstack/cinder.go
```diff
@@ -17,10 +17,6 @@ import (
 
 // ReconcileCinder -
 func ReconcileCinder(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, helper *helper.Helper) (ctrl.Result, error) {
-	if !instance.Spec.Cinder.Enabled {
```
bshephar, line 20: Since this is an established design pattern, I acknowledge that we might not have the appetite to change this at this point. The alternative I would suggest is that we instead update the Status of disabled services to report "Ready" with the message "$serviceName is disabled".

That would allow the overall status of the OpenStackControlPlane to move to Ready, while maintaining the established design pattern. It might just be a little confusing for users though if we're reporting Ready for a service that isn't deployed.

controllers/core/openstackcontrolplane_controller.go
bshephar: Yeah, this isn't necessary since we check for the enabled status in `pkg/openstack/cinder.go`. But I removed it from cinder.go to avoid calling reconcile at all. I just commented on that, acknowledging it might not be a desirable approach.
https://github.com/openstack-k8s-operators/openstack-operator/pull/162#discussion_r1088641451

I'll update the PR and remove this to leave the established pattern in place. ...incoming update.

controllers/core/openstackcontrolplane_controller.go
bshephar: Ok, no worries. Let's do it for all services.

controllers/core/openstackcontrolplane_controller.go
bshephar: I think I still preference leaving it out to avoid confusion. I'll push another update to this PR with a implementation proposal.

pkg/openstack/cinder.go
bshephar: Yeah, I'll revert this and leave it as is.

stuggi (commented):
thanks, +1 from me

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -149,20 +149,30 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 	//
 	if instance.Status.Conditions == nil {
```
gibizer, line 150: All the below condition initialization only runs if Conditions is nil. So it only run once. So if a service was first enabled then disabled then status of that service will not disappear from the condition list. It also means that if you disable a service which has a non Ready status, then OpenStackControlPlane will never reach Ready status. Is it OK? 
I think the current behavior of openstack-operator is that we stop reconciling a service when it is disabled but that also means we stop looking at the status of that service to incorporate it into the status of the OpenStackControlPlane. So if your goal here is to make OpenStackControlPlane Ready with disabled service then you might need to consider this case as well.

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -213,35 +223,35 @@ func (r *OpenStackControlPlaneReconciler) reconcileNormal(ctx context.Context, i
...
-	ctrlResult, err = openstack.ReconcileNeutron(ctx, instance, helper)
+	ctrlResult, err = openstack.ReconcileNova(ctx, instance, helper)
 	if err != nil {
 		return ctrl.Result{}, err
 	} else if (ctrlResult != ctrl.Result{}) {
 		return ctrlResult, nil
 	}
 
-	ctrlResult, err = openstack.ReconcileNova(ctx, instance, helper)
+	ctrlResult, err = openstack.ReconcileCinder(ctx, instance, helper)
```
gibizer, line 254: Is there a specific reason why cinder is moved to the last?

controllers/core/openstackcontrolplane_controller.go
stuggi: > All the below condition initialization only runs if Conditions is nil. So it only run once. So if a service was first enabled then disabled then status of that service will not disappear from the condition list. It also means that if you disable a service which has a non Ready status, then OpenStackControlPlane will never reach Ready status. Is it OK? 

good spot, missed that

bshephar: This is a good point. Should probably add handling for that scenario. I'll look into that and push a new revision.

controllers/core/openstackcontrolplane_controller.go
bshephar: In the earlier patchsets, I had added handling for it being disabled in this file. It's position is just a remnant of that previous patchset, I'll move it back in my next revision

pkg/openstack/cinder.go
```diff
@@ -18,6 +18,7 @@ import (
 // ReconcileCinder -
 func ReconcileCinder(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, helper *helper.Helper) (ctrl.Result, error) {
 	if !instance.Spec.Cinder.Enabled {
+		instance.Status.Conditions.MarkTrue(corev1beta1.OpenStackControlPlaneCinderReadyCondition, "Cinder Service is Disabled")
```
bshephar, line 21: Implementing @kajinamit suggestion here. So in cases where the service is disabled, we just set the Ready status with a message to indicate the service is disabled.

bshephar:
Ok, so latest patchset. If the service is enabled initially, we will have a reported status. If the service is then disabled, then we will just report Ready, with a message about the service being disabled:

```
❯ oc patch OpenStackControlPlane/openstack -p \
'{"spec":{"cinder": {"enabled": false}}}' --type='merge'
```

```
❯ oc get openstackcontrolplane openstack -o json | jq .spec.cinder.enabled
false
```

Which results in:
```
❯ oc get openstackcontrolplane openstack -o json | jq ".status.conditions[1]"
{
  "lastTransitionTime": "2023-01-30T04:48:34Z",
  "message": "OpenStackControlPlane Cinder Service is Disabled",
  "reason": "Ready",
  "status": "True",
  "type": "OpenStackControlPlaneCinderReady"
}
```


The things I would like input on are:

1. Do we want to implement as proposed here, where if services are initially disabled, they are just left out from the Status entirely. If they are enabled, they initially appear. Once disabled, we just leave the status there, report Ready but provide a message about the service now being disabled.

2. Do we want to remove all handling for enabled / disabled services from the Controller itself and just let the reconcileService() functions report "Ready" "$serviceName is Disabled".

gibizer:
> The things I would like input on are:
> 
>     1. Do we want to implement as proposed here, where if services are initially disabled, they are just left out from the Status entirely. If they are enabled, they initially appear. Once disabled, we just leave the status there, report Ready but provide a message about the service now being disabled.

We have multiple connected behavior with different motivations / end goals. So I think it worth teasing them apart a bit to understand the motivations.

* Goal1: do not deploy services that are not needed by the deployment. So there could be disabled services in OpenStackControlPlane when it is *created*.
* Goal2: allow gradually enable services during adoption. I.e. most of the services (even mandatory ones) might be disabled at first when OpenStackControlPlane is *created*. But then later on will be *enabled* probably one by one.
* Goal3: OpenStackControlPlane.Status.Condition.Ready should reflect the status of the existing deployment compared against what is requested in OpenStackControlPlane.Spec. So if some service is *disabled* in Spec then the Ready condition can *ignore* the status of that service as that service is not part of what is requested to be ready.

I'm not sure about the next goal but I found use cases that make sense at least to me.
* Goal4: allow disabling an already enabled service. The use case could be either:
  * Goal4.1: After a successful deployment the end user decides that some service are not needed as they are unused and they want to get rid of them to decrease the unnecessary complexity of the control plane.
  * Goal4.2: An adoption step enabled a service but failed to adopting it and the upgrade engineer tries to re-try the adoption step by going back to the starting state of it. I.e. disabling the service.

We made some implementation decision before that I cannot connect to the above goals:
* Impl1: if a service is disabled we stop reconciling it and stop updating our status based on its status but we don't delete the service CR. So the service controller might eventually deploys the service and the service CR go to Ready independently what the OpenStackControlPlane.Status shows.
* Impl2: The OpenStackControlPlane.Status.Condition.Ready is only set to True if all the other conditions in the list are True.

I see the following problems:
* Problem1: A service can be disabled in OpenStackControlPlane but fully running in the deployment. So Goal4.1 cannot really be achieved simply by disabling an unneeded service.
* Problem2: The status of the service specific condition in OpenStackControlPlane.Status.Condition diverging from the condition in the service CR.
* Problem3: It is probably a nit, but with the proposal "provide a message about the service now being disabled" above we are duplicating information. The enabled=False field in the Spec is blindly reflected in the message of the condition. This condition value either need to signify something that changed in the real deployment (see Problem1) or we don't need it as the Spec value is enough.

I can see one solution that solves all problems:

1. Solution1: When a service is disabled then delete the service CR and after that succeeded, delete the service specific condition from OpenStackControlPlane.Status.Condition.

If we can ignore Goal4 then I see one additional possible solution:

2. Solution2: When a service is disabled, stop reconciling it as today but continue monitoring the status of the service CR to update the service specific condition in OpenStackControlPlane.Status.Condition. However modify the logic that calculates the OpenStackControlPlane.Status.Condition.Ready condition to ignore the value of the service specific condition of each disabled service. This resolves Problem2 and Problem3 but does not resolve Problem1 so Goal4 is not achieved.

gibizer:
> 
>     1. Solution1: When a service is disabled then delete the service CR and after that succeeded, delete the service specific condition from OpenStackControlPlane.Status.Condition.

We might be able just do that as part of https://github.com/openstack-k8s-operators/openstack-operator/pull/155

bshephar:
> > ```
> > 1. Solution1: When a service is disabled then delete the service CR and after that succeeded, delete the service specific condition from OpenStackControlPlane.Status.Condition.
> > ```
> 
> 
> We might be able just do that as part of #155

Thanks for the detailed feedback. I agree that Rabi's PR addresses the cleanup of the already enabled services. So, we can address the Status problem here in this PR, while the other one can handle the implementation of how we reconcile services that get disabled.

My ideal solution would be to have a data structure that holds the enabled services. Then iterate over that list to set the conditions. I just haven't quite got the implementation right for that yet. In lieu of that desired state, I tried to group the dependent services to set their conditions if they are enabled. So without writing if statements for each and every service, I think that grouping more-or-less covers Goal2 right? Where, we don't set the status unless the service is enabled. Using Cinder again as the example. If it were to be enabled later, then we would ultimately end up setting the condition once the this point in the ReconcileCinder function: 
https://github.com/openstack-k8s-operators/openstack-operator/blob/master/pkg/openstack/cinder.go#L66-L73

For Goal3. I think that would be easily achievable if we had a `Status.EnabledServices` list like I mentioned above as well. We could iterate over the instance definition to pull out the services that are enabled, append them to the list and then use that simplify the implementations of the other operations. Such as the the implementation here:
https://github.com/openstack-k8s-operators/openstack-operator/blob/master/controllers/core/openstackcontrolplane_controller.go#L180

Could be simplified by doing something like:

```
for _, svc := range instance.Status.EnabledServices {
    reconcileService(svc)
}
```

Similar things could be done for the overall Ready status. Rather than statically defining each:
https://github.com/openstack-k8s-operators/openstack-operator/blob/master/apis/core/v1beta1/openstackcontrolplane_types.go#L276-L283


But I'll need some more time, preferably a different PR to implement that ideal state if that works for you here?

gibizer:
> > > ```
> > > 1. Solution1: When a service is disabled then delete the service CR and after that succeeded, delete the service specific condition from OpenStackControlPlane.Status.Condition.
> > > ```
> > 
> > 
> > We might be able just do that as part of #155
> 
> Thanks for the detailed feedback. I agree that Rabi's PR addresses the cleanup of the already enabled services. So, we can address the Status problem here in this PR, while the other one can handle the implementation of how we reconcile services that get disabled.
> 
> My ideal solution would be to have a data structure that holds the enabled services. Then iterate over that list to set the conditions. I just haven't quite got the implementation right for that yet. In lieu of that desired state, I tried to group the dependent services to set their conditions if they are enabled. So without writing if statements for each and every service, I think that grouping more-or-less covers Goal2 right? Where, we don't set the status unless the service is enabled. Using Cinder again as the example. If it were to be enabled later, then we would ultimately end up setting the condition once we get to the this point in the ReconcileCinder function: https://github.com/openstack-k8s-operators/openstack-operator/blob/master/pkg/openstack/cinder.go#L66-L73

Yeah, lets focus on not adding condition for a disable service during condition init. I think #155 can cover the rest. 
Yes, if Cinder is later enabled then the reconciliation will add the status back eventually at the place you linked to. 

> 
> For Goal3. I think that would be easily achievable if we had a `Status.EnabledServices` list like I mentioned above as well. We could iterate over the instance definition to pull out the services that are enabled, append them to the list and then use that to simplify the implementations of the other operations. Such as the the implementation here: https://github.com/openstack-k8s-operators/openstack-operator/blob/master/controllers/core/openstackcontrolplane_controller.go#L180
> 
> Could be simplified by doing something like:
> 
> ```
> for _, svc := range instance.Status.EnabledServices {
>     reconcileService(svc)
> }
> ```
> 
> Similar things could be done for the overall Ready status. Rather than statically defining each: https://github.com/openstack-k8s-operators/openstack-operator/blob/master/apis/core/v1beta1/openstackcontrolplane_types.go#L276-L283
> 
> But I'll need some more time, preferably a different PR to implement that ideal state if that works for you here?

Sure, we can take this separately. 

I can convinced to have `instance.Status.EnabledServices` if it has an extra meaning over `instance.Spec.[Cinder|Glance|...].Enabled`. For example if we want to differentiate between instance.Spec.Cinder.Enabled is set to `true` by the user to request the deployment of cinder and `instance.Status.EnabledService[Cinder]` being `true` meaning the deployment of Cinder has been succeeded. Otherwise I would not expose the EnabledService on our API. Instead I would have it as an internal variable that is filled based on `instance.Spec.[Cinder|Glance|...].Enabled` at the start of `Reconcile` and used to ease the iteration of the enabled service as you suggested above.

controllers/core/openstackcontrolplane_controller.go
bshephar: So, ultimately the outcome here is that we will end up marking the service as Ready, with a message in the status field that the Service is disabled.

I understand @gibizer point around having a message to explain it's disabled, when we can also see it's disabled in the spec. I think probably from a user perspective, they might be wondering why the service has been marked as Ready, and having the message there probably helps to clarify. So that is implemented in the cinder.go `ReconcileCinder()` function, during the service enabled check.

PR #163, opened 2023-01-26
Add Ironic API to ControlPlane CR

steveb:
It is currently disabled by default.


controllers/core/openstackcontrolplane_controller.go
```diff
@@ -160,6 +162,7 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 			condition.UnknownCondition(corev1beta1.OpenStackControlPlaneKeystoneAPIReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlaneKeystoneAPIReadyInitMessage),
 			condition.UnknownCondition(corev1beta1.OpenStackControlPlanePlacementAPIReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlanePlacementAPIReadyInitMessage),
 			condition.UnknownCondition(corev1beta1.OpenStackControlPlaneGlanceReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlaneGlanceReadyInitMessage),
+			condition.UnknownCondition(corev1beta1.OpenStackControlPlaneIronicReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlaneIronicReadyInitMessage),
```
bshephar, line 165: I believe this will prevent the openstackcontrolplane CR from reporting Ready if we deploy with Ironic disabled:
https://issues.redhat.com/browse/OSP-21711

Because we initialise the status as Unknown, and then don't run anything in the ReconcileIronic function, so nothing will ever update the status from Unknown. This is my initial proposal for handling disabled services, although I would like to work towards a better implementation that prevents us needing to handle things on a operator by operator basis eventually:

https://github.com/openstack-k8s-operators/openstack-operator/pull/162/commits/aeaf42bb4782aef17a5bc55567bfa1a819e80023

pkg/openstack/ironic.go
```diff
@@ -0,0 +1,75 @@
...
+	ironicv1 "github.com/openstack-k8s-operators/ironic-operator/api/v1beta1"
+	corev1beta1 "github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// ReconcileIronic -
+func ReconcileIronic(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, helper *helper.Helper) (ctrl.Result, error) {
+	if !instance.Spec.Ironic.Enabled {
+		return ctrl.Result{}, nil
```
bshephar, line 21: The other option would be that we add handling for the Status update in this section before we return here. But imo, I prefer not even calling ReconcileIronic if the service is disabled. Which necessitates that we handle the status initialisation differently during the initial Reconcile function above.

controllers/core/openstackcontrolplane_controller.go
steveb: Done, thanks

pkg/openstack/ironic.go
steveb: Done

abays (commented):
+1 from me

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -164,6 +166,11 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 			condition.UnknownCondition(corev1beta1.OpenStackControlPlaneNovaReadyCondition, condition.InitReason, corev1beta1.OpenStackControlPlaneNovaReadyInitMessage),
 		)
 
+		// Initialize optional service conditions if enabled
+		if instance.Spec.Ironic.Enabled {
```
bshephar, line 170: +1 on this. It will ensure the overall deployment doesn't get stuck reporting Ironic not ready if the service is disabled. Thanks for updating the implementation here.

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -249,6 +256,15 @@ func (r *OpenStackControlPlaneReconciler) reconcileNormal(ctx context.Context, i
 		return ctrlResult, nil
 	}
 
+	if instance.Spec.Ironic.Enabled {
```
bshephar, line 259: Apologies for doing this to you. But the general consensus on the implementation for the other change was the keep the current design standard. So the preference from the wider team is that we just call reconcile here and leave the `if instance.Spec.Ironic.Enabled {` check within in the `ReconcileIronic()` function. Which is what you had originally done.

We should probably change that back before we merge this to satisfy the concerns that were raised on:
https://github.com/openstack-k8s-operators/openstack-operator/pull/162

pkg/openstack/ironic.go
```diff
@@ -0,0 +1,71 @@
...
+
+	ironicv1 "github.com/openstack-k8s-operators/ironic-operator/api/v1beta1"
+	corev1beta1 "github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// ReconcileIronic -
+func ReconcileIronic(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, helper *helper.Helper) (ctrl.Result, error) {
+	ironic := &ironicv1.Ironic{
```
bshephar, line 20: So, adding the check back in here:

```
if !instance.Spec.Ironic.Enabled {
    return ctrl.Result{}, nil
}
```

bshephar (changes requested):
Just one small change request to satisfy some consistency concerns that were raised on the other PR. Then I think we're right to merge it.

controllers/core/openstackcontrolplane_controller.go
steveb: Chur

controllers/core/openstackcontrolplane_controller.go
steveb: Done

dprince (approved):

PR #166, opened 2023-02-01
[WIP]Add generate_crd.sh script to generate CR with custom service containers

raukadah:
Currently meta operator uses core_v1beta1_openstackcontrolplane.yaml to deploy podified control plane. We have hardcoded openstack services containers.

generate_crd.sh scripts takes the above CR and parameterized the services containers. So that user can provide custom path.

It will be useful for CI testing when contents are stored in different registeries with different hashes.

In order to run:./hack/generate_crd.sh. It will generate core_v1beta1_openstackcontrolplane_generated.yaml cr file.

Signed-off-by: Chandan Kumar <raukadah@gmail.com>

### DIFF
--- a/hack/generate_crd.sh
+++ b/hack/generate_crd.sh
@@ -1,0 +1,122 @@
+#!/bin/bash
+
+set -ex
+
+CONFIG_DIR=${CONFIG_DIR:-"../config/samples"}
+CRD_NAME=${CRD_NAME:-"core_v1beta1_openstackcontrolplane_generated.yaml"}
+
+# Containers Vars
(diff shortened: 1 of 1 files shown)


stuggi (commented):
afaik ci uses install_yamls, right? wouldn't it be better to have that integrated there and use the existing way via the env overwrites like e.g.  KEYSTONEAPI_IMG ? we'd probably need a script there to customize the ctlplane CR via kustomize, like in https://github.com/openstack-k8s-operators/install_yamls/blob/master/scripts/gen-service-kustomize.sh#L70 ?

PR #167, opened 2023-02-02
Integrate infra-operator

dprince:
Includes support for new OpenStackClient CR

Drops transportURL from this operator as it now lives in the infra-operator


dprince:
The CI failure here is valid. The containerImage for the OpenStackClient CR isn't getting set thus causing the OpenStackControlPlane CR to fail to reconcile. I will push a fix.

gibizer:
We need nova and cinder to adapt to this:
* https://github.com/openstack-k8s-operators/nova-operator/pull/249 (merged)
* https://github.com/openstack-k8s-operators/cinder-operator/pull/129 (open)

stuggi:
> We need nova and cinder to adapt to this:
> 
> * [switch to infra-operator for rabbitmqv1 nova-operator#249](https://github.com/openstack-k8s-operators/nova-operator/pull/249) (merged)
> * [switch to infra-operator for rabbitmqv1 cinder-operator#129](https://github.com/openstack-k8s-operators/cinder-operator/pull/129) (open)

and https://github.com/openstack-k8s-operators/neutron-operator/pull/109

abays:
The Prow failure here is caused by this error:

```
error: evalsymlink failure on '/alabama/install_yamls/out/openstack/openstack/cr' : lstat /alabama/install_yamls/out/openstack/openstack/cr: no such file or directory
```

Part of `make deploy_openstack` in `install_yamls` is to first attempt to clean-up any old deployment [1].  This results in an error if there isn't a previous deployment when this command is executed:

```
oc kustomize /alabama/install_yamls/out/openstack/openstack/cr | oc delete --ignore-not-found=true -f -
```

The referenced directory does not exist and the command exits with a return code of 1, which Prow treats accordingly.  One solution would be to add `|| true` to this `install_yamls` line [2]:

```
oc kustomize ${DEPLOY_DIR} | oc delete --ignore-not-found=true -f -
```

I can imagine other operators will run into this same problem if we try to deploy them via `install_yamls` in Prow, as they all follow this pattern.

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/82d452c3a4c4495c984af04aa50518f09b1916e6/Makefile#L250
[2] https://github.com/openstack-k8s-operators/install_yamls/blob/82d452c3a4c4495c984af04aa50518f09b1916e6/Makefile#L265

gibizer:
As far as I see nova, cinder and neutron dependencies has been merged so this should work now.

gibizer:
the deploy stops at:
```
+ timeout 1200 xargs -d '\n' -I '{}' sh -c 'echo testing condition={}; oc wait openstackcontrolplane.core.openstack.org/openstack --for=condition={} --timeout=-1s'
testing condition=OpenStackControlPlaneClientReady
```

and the operator reports:
```
            {
                "lastTransitionTime": "2023-02-10T15:14:47Z",
                "message": "OpenStackControlPlane Client in progress",
                "reason": "Requested",
                "severity": "Info",
                "status": "False",
                "type": "OpenStackControlPlaneClientReady"
            },
```

gibizer:

maybe it is just a slow deploy?

pkg/openstack/openstackclient.go
```diff
@@ -0,0 +1,87 @@
...
+			condition.SeverityWarning,
+			corev1beta1.OpenStackControlPlaneClientReadyErrorMessage,
+			err.Error()))
+		return ctrl.Result{}, err
+	}
+	if op != controllerutil.OperationResultNone {
+		helper.GetLogger().Info(fmt.Sprintf("OpenStackClient %s - %s", openstackclient.Name, op))
+	}
+
+	if openstackclient.Status.Conditions.IsTrue(condition.DeploymentReadyCondition) {
```
gibizer, line 76: This is not in line with the conditions the OpenstackClient CRD exposes. Either this needs to check clientv1beta1.OpenStackClientReadyCondition or more over check condition.ReadyCondition

https://github.com/openstack-k8s-operators/infra-operator/blob/6e3aad14c3aacb00e343bc1777970b4e6c4a9b8d/controllers/client/openstackclient_controller.go#L245-L248

https://github.com/openstack-k8s-operators/infra-operator/blob/6e3aad14c3aacb00e343bc1777970b4e6c4a9b8d/controllers/client/openstackclient_controller.go#L105

gibizer (changes requested):
There is a condition naming / usage mismatch between the OpenStackClient in infra-operator and what the openstack-operator uses to detect when the client is ready. This causing the end2end job failure.

pkg/openstack/openstackclient.go
dprince: Hi @gibizer. I found similar independently. Also adjusted the Makefile to ensure I don't break 'make run' with the new way I pass in the openstackclient container.

dprince:
passing CI, but now a conflict. let me rebase

gibizer (approved):
I have only a cleanup nit that can be fixed separately.

PR #168, opened 2023-02-03
Support exposing rabbitmq instance via metallb

stuggi:
introduce externalEndpoints parameters per rabbitmq instance to configure metallb LoadBalancer service using the pre-created metallb IPAddressPool and L2Advertisement.

E.g. to expose the api on an internalapi network which
- its metallb ipaddresspool is configured in an osp-internalapi object
- and request a specific loadbalancerIP which is shared with other services (list of IPs for multi stack ipv4/ipv6)

```
  rabbitmq:
    enabled: true
    templates:
      rabbitmq:
        replicas: 1
        externalEndpoint:
          ipAddressPool: osp-internalapi
          loadBalancerIPs:
          - 172.17.0.202
          sharedIP: true
          sharedIPKey: ""
      rabbitmq-cell1:
        replicas: 1
```

Depends-On: openstack-k8s-operators/lib-common#160


apis/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -164,7 +164,42 @@ type RabbitmqSection struct {
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=""
+	// SharedIPKey specifies the sharing key which gets set as the annotation on the LoadBalancer service.
+	// Services which share the same VIP must have the same SharedIPKey. Defaults to the IPAddressPool if
+	// SharedIP is true, but no SharedIPKey specified.
+	SharedIPKey string `json:"sharedIPKey"`
+
+	// +kubebuilder:validation:Optional
+	// LoadBalancerIPs, request given IPs from the pool if available. Using a list to allow dual stack (IPv4/IPv6) support
+	LoadBalancerIPs []string `json:"loadBalancerIPs"`
```
dprince, line 202: What is our gameplan on how we'll provide these to services? Like I create an OSControlPlane resource... what is going to fill in the loadBalancerIPs for me? Or do I just have to specify them manually?

And if there is a collision (two services specify the same IPs) first one to reconcile wins I guess. And then it the 2nd service would fail to reconcile completely until they are edited to be different?

stuggi: > What is our gameplan on how we'll provide these to services? Like I create an OSControlPlane resource... what is going to fill in the loadBalancerIPs for me? Or do I just have to specify them manually?

The LoadBalancerIPs is optional, if you don't set it, the next free one on the pool get picked. right now if you want a fixed one you have to manually specify them. You also have to manually create the metallb pool

> And if there is a collision (two services specify the same IPs) first one to reconcile wins I guess. And then it the 2nd service would fail to reconcile completely until they are edited to be different?

The default is to request shared IPs using `sharedIP` which defaults to true. If no LoadBalancerIPs get specified they'll use the same IP. 
If services request non shared IPs ( `sharedIP` == false) and request the same LoadBalancerIP, then one will win and the other service won't get an IP. In a follow up we should add validation via web hook to check 
- for `sharedIP` false, LoadBalancerIPs are unique if specified
- for rabbitmq in specific, at least right now, LoadBalancerIPs need to be unique across rabbitmq instances, or we need to pick different ports.

abays (commented):
+1 from me as far as I understand it

stuggi:
just rebased

dprince (approved):

PR #169, opened 2023-02-03
Add build_root image configuration for prow

viroel:
ci-operator is the tool that handles the process of building images and running workflows in Prow jobs. By adding this configuration in the repo, we can replace the base root image (which will serve as base for most of the jobs) along with code changes, when upgrading dependencies (golang for instance)
This file has no effect for now, only after updating ci-operator config on openshift/release, in another PR.

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,0 +1,4 @@
+build_root_image:
+  name: tools
+  namespace: openstack-k8s-operators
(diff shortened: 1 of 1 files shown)


viroel:
More info at https://docs.ci.openshift.org/docs/architecture/ci-operator/#build-root-image

dprince (approved):

PR #173, opened 2023-02-07
Update sample config for Nova service containers

raukadah:
Signed-off-by: Chandan Kumar <raukadah@gmail.com>

### DIFF
--- a/config/samples/core_v1beta1_openstackcontrolplane.yaml
+++ b/config/samples/core_v1beta1_openstackcontrolplane.yaml
@@ -91,5 +91,11 @@ spec:
         containerImage: quay.io/tripleozedcentos9/openstack-nova-api:current-tripleo
       schedulerServiceTemplate:
         containerImage: quay.io/tripleozedcentos9/openstack-nova-scheduler:current-tripleo
-      conductorServiceTemplate:
-        containerImage: quay.io/tripleozedcentos9/openstack-nova-conductor:current-tripleo
+      #cellTemplates:
+      #  cell0:
+      #    conductorServiceTemplate:
+      #      containerImage: quay.io/tripleozedcentos9/openstack-nova-conductor:current-tripleo
+      #    metadataServiceTemplate:
+      #      containerImage: quay.io/tripleozedcentos9/openstack-nova-metadata:current-tripleo
+      #    noVNCProxyServiceTemplate:
+      #      containerImage: quay.io/tripleozedcentos9/openstack-nova-novncproxy:current-tripleo
--- a/config/samples/core_v1beta1_openstackcontrolplane.yaml
+++ b/config/samples/core_v1beta1_openstackcontrolplane.yaml
@@ -87,3 +87,9 @@ spec:
   nova:
     template:
       secret: osp-secret
+      apiServiceTemplate:
+        containerImage: quay.io/tripleozedcentos9/openstack-nova-api:current-tripleo
+      schedulerServiceTemplate:
+        containerImage: quay.io/tripleozedcentos9/openstack-nova-scheduler:current-tripleo
+      conductorServiceTemplate:
+        containerImage: quay.io/tripleozedcentos9/openstack-nova-conductor:current-tripleo


config/samples/core_v1beta1_openstackcontrolplane.yaml
```diff
@@ -87,3 +87,9 @@ spec:
   nova:
     template:
       secret: osp-secret
+      apiServiceTemplate:
+        containerImage: quay.io/tripleozedcentos9/openstack-nova-api:current-tripleo
+      schedulerServiceTemplate:
+        containerImage: quay.io/tripleozedcentos9/openstack-nova-scheduler:current-tripleo
+      conductorServiceTemplate:
+        containerImage: quay.io/tripleozedcentos9/openstack-nova-conductor:current-tripleo
```
gibizer, line 95: This will not work as nova has a more nested structure due to cells.

```yaml
$ oc get -o yaml OpenStackControlPlane/openstack | yq -y '.spec.nova'
enabled: true
template:
  apiDatabaseInstance: openstack
  apiDatabaseUser: nova_api
  apiMessageBusInstance: rabbitmq
  apiServiceTemplate:
    containerImage: quay.io/tripleozedcentos9/openstack-nova-api:current-tripleo
    replicas: 1
  cellTemplates:
    cell0:
      cellDatabaseInstance: openstack
      cellDatabaseUser: nova_cell0
      cellMessageBusInstance: unused
      conductorServiceTemplate:
        containerImage: quay.io/tripleozedcentos9/openstack-nova-conductor:current-tripleo
        replicas: 1
      hasAPIAccess: true
  keystoneInstance: keystone
  passwordSelectors:
    apiDatabase: NovaAPIDatabasePassword
    cellDatabase: NovaCell0DatabasePassword
    service: NovaPassword
  schedulerServiceTemplate:
    containerImage: quay.io/tripleozedcentos9/openstack-nova-scheduler:current-tripleo
    replicas: 1
  secret: osp-secret
  serviceUser: nova
```

gibizer (changes requested):
I'm not sure why you need to define something here that is the default.

raukadah:
> I'm not sure why you need to define something here that is the default.

Just like other services, I wanted to define it in samples openstackcontrolplane CR so that they can easily create their kustomization file by referencing that. That's why I was doing that.

config/samples/core_v1beta1_openstackcontrolplane.yaml
raukadah: I have also added the nova cells config related to container image in comments so that one can easily refer that and replace it.

gibizer:
> > I'm not sure why you need to define something here that is the default.
> 
> Thank you @gibizer for reviewing it. Just like other services, I wanted to define it in samples openstackcontrolplane CR so that they can easily create their kustomization file by referencing that. That's why I was doing that.

I believe kustomize can be used even if the field is not mentioned in the sample.

raukadah:
> > > I'm not sure why you need to define something here that is the default.
> > 
> > 
> > Thank you @gibizer for reviewing it. Just like other services, I wanted to define it in samples openstackcontrolplane CR so that they can easily create their kustomization file by referencing that. That's why I was doing that.
> 
> I believe kustomize can be used even if the field is not mentioned in the sample.

Thank you @gibizer for reviewing it.
in my crc openstack dev env, I tried with https://github.com/openstack-k8s-operators/install_yamls/pull/81 patch only and hit by following error while running make openstack_deploy:
```
error: replace operation does not apply: doc is missing path: /spec/nova/template/apiServiceTemplate/containerImage: missing value
```
I have already mentioned the same here also: https://github.com/openstack-k8s-operators/install_yamls/pull/81#issuecomment-1423937130 

Then I applied this patch also and it works fine. I think it does the kustomization over this https://github.com/openstack-k8s-operators/install_yamls/blob/master/Makefile#L23 which reference to sample cr file.

I hope it helps or may be I am missing something in my env.

abays:
> > > > I'm not sure why you need to define something here that is the default.
> > > 
> > > 
> > > Thank you @gibizer for reviewing it. Just like other services, I wanted to define it in samples openstackcontrolplane CR so that they can easily create their kustomization file by referencing that. That's why I was doing that.
> > 
> > 
> > I believe kustomize can be used even if the field is not mentioned in the sample.
> 
> Thank you @gibizer for reviewing it. in my crc openstack dev env, I tried with [openstack-k8s-operators/install_yamls#81](https://github.com/openstack-k8s-operators/install_yamls/pull/81) patch only and hit by following error while running make openstack_deploy:
> 
> ```
> error: replace operation does not apply: doc is missing path: /spec/nova/template/apiServiceTemplate/containerImage: missing value
> ```
> 
> I have already mentioned the same here also: [openstack-k8s-operators/install_yamls#81 (comment)](https://github.com/openstack-k8s-operators/install_yamls/pull/81#issuecomment-1423937130)
> 
> Then I applied this patch also and it works fine. I think it does the kustomization over this https://github.com/openstack-k8s-operators/install_yamls/blob/master/Makefile#L23 which reference to sample cr file.
> 
> I hope it helps or may be I am missing something in my env.

Kustomize `replace` works with a missing field if the field is the only element missing in the path (i.e. `/spec/some_missing_field`).  However, `replace` will fail if the path contains multiple missing tiers (i.e. `/spec/some_missing_field/some_missing_field2`).

raukadah:
Thank you @gibizer @abays for helping me to understand the stuff. I am going to close this pr and submit a new pull request with patchesStrategicMerge in install_yamls repo.

PR #175, opened 2023-02-10
Create ConfigMap to propagate global options

kajinamit:
This introduces the openstack-deployment ConfigMap, which is used to propagate global information such as enabled services.


controllers/core/openstackcontrolplane_controller.go
```diff
@@ -177,7 +186,50 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
...
+	customData := map[string]string{}
+
+	enabledServices := []string{}
+	// TODO: Add all supported services.
+	if instance.Spec.Nova.Enabled {
+		enabledServices = append(enabledServices, "nova")
+	}
+	if instance.Spec.Cinder.Enabled {
+		enabledServices = append(enabledServices, "cinder")
+	}
```
kajinamit, line 205: We might want to add a helper function like https://github.com/openstack-k8s-operators/lib-common/pull/169

PR #178, opened 2023-02-13
Allows to pass base container as build-arg

cjeanner:
At least, using buildah 1.28.0, it seems to be stable enough to pass the images as proper build-arg, allowing to provide a clean interface to override the ones we want.

With this patch exposing all of the images as build arguments, we can even override the golang base container.

Proper usage is, for instance:
```
buildah bud -t my-test:latest \
        --build-arg KEYSTONE_BUNDLE=custom.local/openstack-k8s/keystone-operator-bundle:latest .
```


dprince:
@cjeanner thanks. I had tried doing this a long time ago and it didn't work with podman at the time. Seems to be good now though (I'm on FC37 now) so lets do it.

PR #179, opened 2023-02-14
Remove redundant getters

rabi:
We don't need those for public fields.

### DIFF
--- a/controllers/core/openstackcontrolplane_controller.go
+++ b/controllers/core/openstackcontrolplane_controller.go
@@ -48,26 +48,6 @@ import (
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 )
 
-// GetClient -
-func (r *OpenStackControlPlaneReconciler) GetClient() client.Client {
-	return r.Client
-}
-
-// GetKClient -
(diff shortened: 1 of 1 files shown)


controllers/core/openstackcontrolplane_controller.go
```diff
@@ -47,26 +47,6 @@ import (
...
-
-// GetLogger -
-func (r *OpenStackControlPlaneReconciler) GetLogger() logr.Logger {
-	return r.Log
-}
-
-// GetScheme -
-func (r *OpenStackControlPlaneReconciler) GetScheme() *runtime.Scheme {
-	return r.Scheme
-}
```
abays, line 68: @stuggi Are these just essentially copy-pastes from our old pattern in OSPdO?  Just wondering if we still use this pattern somewhere in any of the new operators.

stuggi: yes, those are left overs from the older ReconcilerCommon and are now no longer needed with the newer Helper

dprince:
yes, lgtm but it needs a rebase now that transport URL lives in the infra-operator

abays (approved):
/lgtm

PR #181, opened 2023-02-14
update csv-merger to include webhookdefinitions

dprince:
This PR updates the csv-merger so that it merges webhookdefinitions from the  CSV files for other operators during bundle creation.

Also bumps the dependencies for the glance-operator which now uses validating webhooks


dprince:
This should pass (again) once we merge or fix the issue identified here: https://github.com/openstack-k8s-operators/openstack-operator/pull/184

rabi:
/lgtm

Thanks Dan!

OLM generates multiple self-signed certs and secrets; one per operator. But I think it's fine for the time being. We can probably look at using same self-signed cert (either user provided/generated) for all operator webhooks in the future.

rabi (approved):

PR #184, opened 2023-02-14
pin-custom-bundle-dockerfile.sh: avoid dup tags

dprince:
This updates the hack/pin-custom-bundle-dockerfile.sh so it avoids duplicate tags which can get returned from quay.io in some situations.

### DIFF
--- a/hack/pin-custom-bundle-dockerfile.sh
+++ b/hack/pin-custom-bundle-dockerfile.sh
@@ -27,7 +27,6 @@ for MOD_PATH in $(go list -m -json all | jq -r '. | select(.Path | contains("ope
       fi
   fi
 
-  SHA=$(curl -s ${REPO_CURL_URL}/$BASE-operator-bundle/tag/ \
-        | jq -r .tags[].name | grep $REF)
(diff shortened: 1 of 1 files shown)


dprince:
Context here is today I merged in support for the dataplane-operator. And when the bundle gets built that operators bundle repo on quay.io seems to return multiple tags (something we have not seen before). Perhaps something to look into but this shouldn't hurt.

dprince:
++ curl -s https://quay.io/api/v1/repository/openstack-k8s-operators/dataplane-operator-bundle/tag/
++ jq -r '.tags[].name'
++ grep 0d4f4a6a874f
+ SHA='0d4f4a6a874f39bdf9c41cf18799b5f205c1196b
0d4f4a6a874f39bdf9c41cf18799b5f205c1196b'
+ sed -i custom-bundle.Dockerfile.pinned -e 's|quay.io/openstack-k8s-operators/dataplane-operator-bundle.*|quay.io/openstack-k8s-operators/dataplane-operator-bundle:0d4f4a6a874f39bdf9c41cf18799b5f205c1196b
0d4f4a6a874f39bdf9c41cf18799b5f205c1196b|'
sed: -e expression #1, char 160: unterminated `s' command

abays (approved):
/lgtm

PR #185, opened 2023-02-15
make run: Set ENABLE_WEBHOOKS=false

dprince:
Also makes the generate-cert target option as it is only useful for really advanced local webhook testing which requires proxying remote API trafic through the local webhook server

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,12 +115,10 @@ build: generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
 	go build -o bin/csv-merger cmd/csv-merger/csv-merger.go
 
-generate-cert:
-	/bin/bash hack/create_self_signed_cert.sh
-
 .PHONY: run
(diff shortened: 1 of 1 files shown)


rabi:
Actually it's possible to proxy the webhook traffic and use the locally generated certificate[1] with 'make run'. Though I've not yet managed to get it working. 

How do we expect the users to use 'make run' target? Is it that you edit the csv to remove the openstack-operator-controller-manager deployment? If you remove the webhookdefinitions along with that (which you anway have to do I guess) would make it work .  I'm fine to use ENABLE_WEBHOOKS=false in 'make run' by default now. However, we should probably leave the cert-generation target and cert generation so that one can use it if they can proxy webhook traffic and debug webhooks.

My intent is to get it working for debugging.

[1] https://book.kubebuilder.io/cronjob-tutorial/running.html#running-and-deploying-the-controller

rabi (approved):

PR #191, opened 2023-02-16
Force syncing the version in go.mod files

gibizer:
It is easy to forget to bump a dependency to the same version both in the go.mod and in the ./apis/go.mod files. Such difference can cause that in different environment make manifests generates different CRD definition based on which dependency version it uses from which go.mod files.

We saw this causing that local pre-commit run in a repo that has a local go.work file produces different CRD defs than a CI run that has no go.work file present as it is being a clean git clone.

This patch adds go work sync to the gowork make target run in pre-commit to force syncing the go.mod files in different directories. This will ensure that if the go.mod file are out of sync the pre-commit will make them in sync and fail on the diff generated.


stuggi (approved):
/lgtm

PR #193, opened 2023-02-16
Add IMAGEBASE var to pin-custim-bundle-dockerfile

viroel:
Adds IMAGEBASE var to allow user to provide a base operator name, which should have its image namespace and/or registry replaced by IMAGENAMESPACE/IMAGEREGISTRY.
Issue: Checking only the GITHUB user does not cover a scenario where a PR is proposed from the same ORG and USER, but from a different branch[1].

[1] https://github.com/openstack-k8s-operators/keystone-operator/pull/180


stuggi (approved):
/lgtm

PR #196, opened 2023-02-20
Add basic service dependency validation

rabi:
This adds basic validation for enabled services. This can/would be further improved (ex. validating when cinder backend for glance is enabled etc).

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_webhook.go
+++ b/apis/core/v1beta1/openstackcontrolplane_webhook.go
@@ -17,6 +17,8 @@ limitations under the License.
 package v1beta1
 
 import (
+	"fmt"
+
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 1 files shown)


apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -61,9 +63,54 @@ func (r *OpenStackControlPlane) ValidateDelete() error {
...
+		return r.Spec.Mariadb.Enabled
+	case "Glance":
+		return r.Spec.Mariadb.Enabled && r.Spec.Keystone.Enabled
+	case "Cinder":
+		return r.Spec.Mariadb.Enabled && r.Spec.Rabbitmq.Enabled && r.Spec.Keystone.Enabled
+	case "Placement":
+		return r.Spec.Mariadb.Enabled && r.Spec.Keystone.Enabled
+	case "Neutron":
+		return (r.Spec.Mariadb.Enabled && r.Spec.Rabbitmq.Enabled &&
+			r.Spec.Keystone.Enabled && r.Spec.Ovn.Enabled && r.Spec.Ovs.Enabled)
```
kajinamit, line 78: We probably should omit ovn and ovs now, because ovn/ovs might not be used when 3rd party SDN such as contrail is used.

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -61,9 +63,54 @@ func (r *OpenStackControlPlane) ValidateDelete() error {
...
+	case "Nova":
```
kajinamit, line 79: nova would require glance as well.

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -61,9 +63,54 @@ func (r *OpenStackControlPlane) ValidateDelete() error {
 	return nil
 }
 
+func (r *OpenStackControlPlane) checkDepsEnabled(name string) bool {
+	switch name {
+	case "KeyStone":
```
kajinamit, line 68: should be Keystone

kajinamit (changes requested):

apis/core/v1beta1/openstackcontrolplane_webhook.go
rabi: Yeah removed them. Though I was looking at the overcloud ovn default.

apis/core/v1beta1/openstackcontrolplane_webhook.go
rabi: Done

apis/core/v1beta1/openstackcontrolplane_webhook.go
rabi: Done

kajinamit (approved):
The change looks good to me know. It sees something is wrong with the test infra.

```
error: Post "https://api.oko-07.oooci.ccitredhat.com:6443/apis/build.openshift.io/v1/namespaces/openstack/buildconfigs": dial tcp 38.102.83.93:6443: connect: connection refused
```

PR #197, opened 2023-02-20
Fix ironic-oprator storageClass override

hjensas:
Since commit - https://github.com/openstack-k8s-operators/ironic-operator/commit/0b66e35fed7b1868d93af15aa984437adade0288 the storage Class should be set on the Ironic Spec not on the Conductror spec.


hjensas:
closing, this was fixed in a different patch already.

PR #200, opened 2023-02-21
Update Neutron validation dependencies

karelyatin:
Dependency on ovn was missing, this adds it.

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_webhook.go
+++ b/apis/core/v1beta1/openstackcontrolplane_webhook.go
@@ -76,7 +76,7 @@ func (r *OpenStackControlPlane) checkDepsEnabled(name string) bool {
 		return r.Spec.Mariadb.Enabled && r.Spec.Keystone.Enabled
 	case "Neutron":
 		return (r.Spec.Mariadb.Enabled && r.Spec.Rabbitmq.Enabled &&
-			r.Spec.Keystone.Enabled)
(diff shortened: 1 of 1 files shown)


apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -76,7 +76,7 @@ func (r *OpenStackControlPlane) checkDepsEnabled(name string) bool {
 		return r.Spec.Mariadb.Enabled && r.Spec.Keystone.Enabled
 	case "Neutron":
 		return (r.Spec.Mariadb.Enabled && r.Spec.Rabbitmq.Enabled &&
-			r.Spec.Keystone.Enabled)
+			r.Spec.Keystone.Enabled && r.Spec.Ovn.Enabled)
```
rabi, line 79: We discussed that in that PR and removed it as as there maybe third-party sdn[1]

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/196#discussion_r1111556251

karelyatin: ohkk me not sure how third party sdn going to work, but as of now neutron CR wouldn't work without ovn.

rabi: Are we not going to support any other mechanism drivers with nextgen? I guess the check here would based on mechanism drivers in CustomServiceConfig and the neutron-operator code also has to change accordingly(?).

PR #201, opened 2023-02-21
Reconcile OpenStackDataPlaneNode

fao89:
Depends on:
- https://github.com/openstack-k8s-operators/dataplane-operator/pull/48
- https://github.com/openstack-k8s-operators/dataplane-operator/pull/49


pkg/openstack/dataplane.go
```diff
@@ -0,0 +1,64 @@
...
+	if !instance.Spec.DataPlaneNode.Enabled {
+		return ctrl.Result{}, nil
+	}
+
+	dataPlaneNode := &dataplanev1beta1.OpenStackDataPlaneNode{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "dataplanenode", //FIXME
+			Namespace: instance.Namespace,
+		},
+	}
```
fao89, line 29: I'm not so sure if I need to add more initial values from: https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanenode_deployment.yaml

@slagle wdyt?

dprince: within reason I would add what is needed here to streamline a demo of this workflow for now. If it really only applies to our Q1 demo and we might have better ways of setting defaults in the future perhaps comment them to address those concerns. If we run the demo via the OpenShift web console I would also add that the default config/sample for each CR is going to get slurped into alm-examples for the CSV and would thus get applied as default values in the UI there. But that only applies in the OpenShift web console I think

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -303,6 +313,7 @@ func (r *OpenStackControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) err
 		Owns(&neutronv1.NeutronAPI{}).
 		Owns(&novav1.Nova{}).
 		Owns(&ironicv1.Ironic{}).
+		Owns(&dataplanev1beta1.OpenStackDataPlaneNode{}).
```
dprince, line 316: I don't think the ControlPlane CR should own these.

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -224,6 +226,13 @@ func (r *OpenStackControlPlaneReconciler) reconcileNormal(ctx context.Context, i
 		return ctrlResult, nil
 	}
 
+	ctrlResult, err = openstack.ReconcileDataplaneNode(ctx, instance, helper)
```
dprince, line 229: Why are we reconciling Dataplane nodes as part of the controlplane? I thought these were meant to be managed as part of the DataPlane CR?

dprince (changes requested):

controllers/core/openstackcontrolplane_controller.go
fao89: oh, my bad! I was following the pattern of the other operators and didn't realize it was the controlplane

fao89:
I'm closing this, as I realize now the approach is way different than I was thinking

PR #203, opened 2023-02-22
Configure mutating webhook

rabi:
Add defaults for ironic storageClass.

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_webhook.go
+++ b/apis/core/v1beta1/openstackcontrolplane_webhook.go
@@ -67,3 +67,22 @@ func (r *OpenStackControlPlane) ValidateServices() error {
 	return nil
 
 }
+
+// +kubebuilder:webhook:path=/mutate-core-openstack-org-v1beta1-openstackcontrolplane,mutating=true,failurePolicy=fail,sideEffects=None,groups=core.openstack.org,resources=openstackcontrolplanes,verbs=create;update,versions=v1beta1,name=mopenstackcontrolplane.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Defaulter = &OpenStackControlPlane{}
+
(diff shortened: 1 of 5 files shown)


apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -67,3 +67,22 @@ func (r *OpenStackControlPlane) ValidateServices() error {
...
+
+var _ webhook.Defaulter = &OpenStackControlPlane{}
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type
+func (r *OpenStackControlPlane) Default() {
+	openstackcontrolplanelog.Info("Before calling webhook")
+	// Default storageClass for ironic
+	if r.Spec.Ironic.Enabled {
+		r.Spec.Ironic.Template.StorageClass = r.Spec.StorageClass
+		for _, c := range r.Spec.Ironic.Template.IronicConductors {
```
hjensas, line 81: Since we have storage class in the top level Ironics now and do a similar transfer to each conductor group in IronicCounductors in ironic-operator we should not need to do the iteration here? i.e L80 should be enough.

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -67,3 +67,22 @@ func (r *OpenStackControlPlane) ValidateServices() error {
...
+			if c.StorageClass == "" {
+				c.StorageClass = r.Spec.StorageClass
```
hjensas, line 83: Does this assign actually work?
I was doing something similar in ironic-operator, and looks like c is a copy of the element in the array - so assigning to it does not update the value. I had to use the loop index like example below.
```

for idx, c := range spec.IronicConductors {
	// StorageClass
	if c.StorageClass == "" {
		spec.IronicConductors[idx].StorageClass = spec.StorageClass
	}
```

PR #205, opened 2023-02-23
Simplify delete reconcile

rabi:
I don't see any issues with deletion with proper finalizers and owner references atm. Also there are more resources like ironic/openstackclient added that are not yet included.

### DIFF
--- a/controllers/core/openstackcontrolplane_controller.go
+++ b/controllers/core/openstackcontrolplane_controller.go
@@ -27,7 +27,6 @@ import (
 	keystonev1 "github.com/openstack-k8s-operators/keystone-operator/api/v1beta1"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
-	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	mariadbv1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
 	neutronv1 "github.com/openstack-k8s-operators/neutron-operator/api/v1beta1"
 	novav1 "github.com/openstack-k8s-operators/nova-operator/api/v1beta1"
@@ -42,7 +41,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


rabi:
@abays This works without any step ordering and explicit api calls. Let me know if you see any issues.

abays:
> @abays This works without any step ordering and explicit api calls. Let me know if you see any issues.

With the finalizers we added for MariaDB and Keystone sub-resources, this might be fine to remove.  However, it could also potentially reintroduce races that you might not notice when testing this PR a few times.  Let me try a few things and get back to you on that.

abays:
> > reintroduce races that you might not notice when testing this PR a few times
> 
> I think, those won't happen unless we use ownerrefs and finalizers properly in the future. I guess it's better for it to break and we fix rather than doing non-k8s style controlled cleanup(?).

I'm not seeing any of the snags we were hitting before, so I think the finalizers I mentioned are doing their intended jobs.  We'll just have to keep an eye out for other dependencies between service operators, especially if they introduce finalizers between one another.  But I think this is good for now.

abays:
@dprince Could you take a look at this and/or try it when you get a chance?  It works in my environments.  I think we can indeed remove this teardown logic now that we have proper finalizers in place for Keystone and MariaDB.

dprince (approved):

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -133,13 +131,9 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
...
 
-	// If we're not deleting this and the service object doesn't have our finalizer, add it.
-	if instance.DeletionTimestamp.IsZero() && controllerutil.AddFinalizer(instance, helper.GetFinalizer()) {
-		return ctrl.Result{}, err
-	}
-
 	if !instance.DeletionTimestamp.IsZero() {
-		return r.reconcileDelete(ctx, instance, helper)
+		r.Log.Info("Reconciled OpenStackControlPlane delete successfully")
+		return ctrl.Result{}, err
```
gibizer, line 136: Based on my tests we never reach this code path now. Probably because there is no finalizer on the OpenstackControlPlane resource, the delete of the resource is automatic and even if the reconciler is called it will get NotFound from the API why trying to read the instance. As we are not logging there the `make openstack_cleanup_deploy` target actually logs nothing in the openstack-operator.

gibizer: Proposed cleanup and log move in https://github.com/openstack-k8s-operators/openstack-operator/pull/229

rabi: Hmm.. look like if custom resource does not have any finalizers, the controller will not be involved in the deletion/finalization process. I thought metdata deletionTimestamp would anyway be set and it's possible for controller reconcile to run before it's been garbage collected.

PR #208, opened 2023-02-23
Use field.ErrorList in ValidateServices()

hjensas:
Set up an error list field, and append issues to the list when validating service dependencies. There is no reason to stop on the first error.

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_webhook.go
+++ b/apis/core/v1beta1/openstackcontrolplane_webhook.go
@@ -19,7 +19,10 @@ package v1beta1
 import (
 	"fmt"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
@@ -46,14 +49,40 @@ var _ webhook.Validator = &OpenStackControlPlane{}
 func (r *OpenStackControlPlane) ValidateCreate() error {
 	openstackcontrolplanelog.Info("validate create", "name", r.Name)
 
-	return r.ValidateServices()
+	var allErrs field.ErrorList
+
+	allErrs = append(allErrs, r.ValidateServices()...)
+
+	if len(allErrs) != 0 {
+		return apierrors.NewInvalid(
+			schema.GroupKind{
+				Group: "openstackcontrolplane.openstack.org",
+				Kind:  "OpenstackControlPlane",
+			},
+			r.Name, allErrs)
+	}
+
+	return nil
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (r *OpenStackControlPlane) ValidateUpdate(old runtime.Object) error {
 	openstackcontrolplanelog.Info("validate update", "name", r.Name)
 
-	return r.ValidateServices()
+	var allErrs field.ErrorList
+
+	allErrs = append(allErrs, r.ValidateServices()...)
+
+	if len(allErrs) != 0 {
+		return apierrors.NewInvalid(
+			schema.GroupKind{
+				Group: "openstackcontrolplane.openstack.org",
+				Kind:  "OpenstackControlPlane",
+			},
+			r.Name, allErrs)
+	}
+
+	return nil
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type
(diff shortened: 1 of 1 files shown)


apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -46,14 +49,42 @@ var _ webhook.Validator = &OpenStackControlPlane{}
 func (r *OpenStackControlPlane) ValidateCreate() error {
 	openstackcontrolplanelog.Info("validate create", "name", r.Name)
 
-	return r.ValidateServices()
+	var allErrs field.ErrorList
+
+	allErrs = append(allErrs, r.ValidateServices()...)
+
+	if len(allErrs) != 0 {
+		return apierrors.NewInvalid(
```
rabi, line 57: Should we use newInvalid() or newConflict()/newApplyConflict() instead?  It's not invalid value per se.

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -46,14 +49,42 @@ var _ webhook.Validator = &OpenStackControlPlane{}
...
+	}
+
+	if len(allErrs) != 0 {
+		return apierrors.NewInvalid(
+			schema.GroupKind{
+				Group: "openstackcontrolplane.openstack.org",
+				Kind:  "OpenstackControlplane",
+			},
+			r.Name, allErrs)
+	}
```
rabi, line 85: Isn't this the same code for both ValidateCreate() and ValidateUpdate() and can be in ValdateServices()? We don't use old object for update.

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -81,39 +112,127 @@ func (r *OpenStackControlPlane) checkDepsEnabled(name string) bool {
...
 
-	return nil
+	if r.Spec.Ironic.Enabled && !r.checkDepsEnabled("Ironic-JsonRPC") {
+		path = field.NewPath("spec").Child("Ironic").Child("enabled")
+		dependencies = "Mariadb and Keystone"
+		allErrs = append(
+			allErrs,
+			field.Invalid(
+				path,
+				r.Spec.Nova.Enabled,
```
rabi, line 218: r.Spec.Ironic.Enabled ?

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -81,39 +112,127 @@ func (r *OpenStackControlPlane) checkDepsEnabled(name string) bool {
...
+				fmt.Sprintf(errorMsg, "Ironic", dependencies),
+			),
+		)
+	}
+
+	if (r.Spec.Ironic.Enabled && r.Spec.Ironic.Template.RPCTransport == "oslo") && !r.checkDepsEnabled("Ironic-JsonRPC") {
```
rabi, line 224: r.checkDepsEnabled("Ironic")?

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -81,39 +112,127 @@ func (r *OpenStackControlPlane) checkDepsEnabled(name string) bool {
...
+		path = field.NewPath("spec").Child("Ironic").Child("enabled")
+		dependencies = "Mariadb, Rabbitmq and Keystone"
+		allErrs = append(
+			allErrs,
+			field.Invalid(
+				path,
+				r.Spec.Nova.Enabled,
```
rabi, line 231: r.Spec.Ironic.Enabled ?

apis/core/v1beta1/openstackcontrolplane_webhook.go
hjensas: Done.

apis/core/v1beta1/openstackcontrolplane_webhook.go
hjensas: Done

apis/core/v1beta1/openstackcontrolplane_webhook.go
hjensas: I imagine we will run other validations than `ValidateServices`, each validation function returns a list of errors and append. We can change it to `ValidateServices()` and then have it call a `ValidateServiceDependencies()` and add other validations in `ValidateServices()` as they get introduced. ?

apis/core/v1beta1/openstackcontrolplane_webhook.go
hjensas: Using the list, the idea is that the apierrors raised will be from all validations, including validations in individual operators. For example here we append errors from ironic-operator validator - https://github.com/openstack-k8s-operators/openstack-operator/blob/92bcbf201c3bb6b399fc1b605f80cedaae185904/apis/core/v1beta1/openstackcontrolplane_webhook.go#L54-L67

Each error in "allErrs" can be of different type, conflict, forbidden, invalid etc.
 I think "Invalid" for each dependency check might not be the best error, but I think we may want to iterate and add some custom error types?

At this level a NewInvalid seems appropriate? It is the user input that is invalid?
But for individual errors in "allErrs" we can set individual types.

rabi: The way newInvaid() is rendered in the response[1]  can be confusing when it says the field is invalid. "it's valid field value but it conflicts with other values". Probably having our own custom StatusError would be better.


[1] https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/admission/validator.go#L70
[2] https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L145

apis/core/v1beta1/openstackcontrolplane_webhook.go
rabi: OK, I was thinking that we would probably not do any validation specific for update/patch and compare with old object. But possibly there could be a few like changing the ml2 driver etc.

apis/core/v1beta1/openstackcontrolplane_webhook.go
hjensas: This is what kubectr/oc outputs dependency errors in multiple services:

```
The OpenstackControlplane "openstack" is invalid: 
* spec.Keystone.enabled: Invalid value: true: Keystone service dependencies are not enabled, needs: Mariadb
* spec.Neutron.enabled: Invalid value: true: Neutron service dependencies are not enabled, needs: Mariadb, Rabbitmq and Keystone
* spec.Nova.enabled: Invalid value: true: Nova service dependencies are not enabled, needs: Mariadb, Rabbitmq, Keystone, Placement, Neutron and Glance
* spec.Ironic.enabled: Invalid value: true: Ironic service dependencies are not enabled, needs: Mariadb and Keystone
make: *** [Makefile:280: openstack_deploy] Error 1
```

I think `The OpenstackControlplane "openstack" is invalid` is not bad, but each individual field we `Invalid value` is not so good, I don't se a better alternative in the package i.e would be good to follow up with custom error - MissingDependency or something?

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -46,14 +49,42 @@ var _ webhook.Validator = &OpenStackControlPlane{}
...
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (r *OpenStackControlPlane) ValidateUpdate(old runtime.Object) error {
 	openstackcontrolplanelog.Info("validate update", "name", r.Name)
 
-	return r.ValidateServices()
+	var allErrs field.ErrorList
+
+	if err := r.ValidateServices(); err != nil {
+		allErrs = append(allErrs, err...)
+	}
```
rabi, line 76: nit: Create and update has different syntax , here `err` is a `field.ErrorList`. We can probably sync that if you're updating again.

rabi (approved):

rabi (approved):
/lgtm

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -46,14 +49,40 @@ var _ webhook.Validator = &OpenStackControlPlane{}
...
-	return r.ValidateServices()
+	var allErrs field.ErrorList
+
+	allErrs = append(allErrs, r.ValidateServices()...)
+
+	if len(allErrs) != 0 {
+		return apierrors.NewInvalid(
+			schema.GroupKind{
+				Group: "openstackcontrolplane.openstack.org",
+				Kind:  "OpenstackControlplane",
```
dprince, line 80: I don't think it really matters bit I think the last P should be capitalized here right (to fully match the CRD)? OpenStackControlPlane

dprince (approved):

apis/core/v1beta1/openstackcontrolplane_webhook.go
hjensas: I had to rebase, and change the MariaDB/Galera check to add its error to the list so I took the opportunity to change to capital P here as well.

hjensas:
Closing, https://github.com/openstack-k8s-operators/openstack-operator/pull/220 implements the same.

PR #206, opened 2023-02-23
validation webhook: run ironic Validators

hjensas:
Depends-On: https://github.com/openstack-k8s-operators/ironic-operator/pull/141

Jira: [OSP-22630](https://issues.redhat.com//browse/OSP-22630)

PR #220, opened 2023-03-06
Validation and mutation webhook call-out pattern

abays:
Introduces pattern of webhooks for service operator validations and defaults which call into the specific mutation and validation logic of the individual operators (in this first PR, for Glance and Ironic).

### DIFF
--- a/cmd/csv-merger/csv-merger.go
+++ b/cmd/csv-merger/csv-merger.go
@@ -29,6 +29,7 @@ import (
 
 	csvv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -123,7 +124,6 @@ func main() {
 			Status:     csvBase.Status}
 
 		installStrategyBase := csvBase.Spec.InstallStrategy.StrategySpec
-		webhookdefinitions := csvBase.Spec.WebhookDefinitions
 
 		for _, csvFile := range csvs {
 			if csvFile != "" {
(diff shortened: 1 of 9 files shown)


cmd/csv-merger/csv-merger.go
```diff
@@ -140,6 +140,32 @@ func main() {
...
+						installStrategyBase.DeploymentSpecs[0].Spec.Template.Spec.Containers[1].Env = append(
+							// OpenStack Operator controller-manager container env vars
+							installStrategyBase.DeploymentSpecs[0].Spec.Template.Spec.Containers[1].Env,
+							// Service Operator controller-manager container env vars
+							container.Env...,
+						)
+						// We found the controller-manager container, so no need to continue iterating
+						break
+					}
+				}
```
dprince, line 167: The above part looks good. In addition to that however don't we also need to append an extra ENV to all service operators (glance-operator, etc.) so that we set ENABLE_WEBHOOKS=false for each as when they are installed alongside of openstack-operator we don't want the local webhooks enabled there.

dprince (changes requested):

cmd/csv-merger/csv-merger.go
abays: Good call.  I will add that.

rabi:
I  had a patch to add defaulting webhooks[1] marked WIP as I was working around adding omitempty tags across all operators. I'll abandon mine as this was already received some reviews..

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/203

gibizer (commented):
looks OK to me

rabi:
Can we also add the defaulting webhook for local testing [1]?

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/hack/configure_local_webhook.sh

cmd/csv-merger/csv-merger.go
abays: @dprince Should we also remove the `webhookDefinitions` incorporation?  If we're setting `ENABLE_WEBHOOKS=false` in the service operator's deployment's env vars, then we wouldn't want there to be `ValidatingWebhookConfiguration`s or `MutatingWebhookConfiguration`s defined for it.  Otherwise those will still try to call the webhook server (which is not running) and throw an error whenever someone creates/update the service operator's CR(s).

abays:
> Can we also add the defaulting webhook for local testing [1]?
> 
> [1] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/hack/configure_local_webhook.sh

Yes, good idea.  Let me add that too.

cmd/csv-merger/csv-merger.go
dprince: Yes, sounds good. lets remove them

main.go
```diff
@@ -147,6 +147,16 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Defaults for service operators
+
+	// Acquire environmental defaults and initialize Glance and GlanceAPI defaults with them
+	glanceDefaults := glancev1.GlanceDefaults{
```
dprince, line 153: this could get messy down the road. may want to consider putting it in a function down the road, but lets go with it for now

dprince (approved):

hjensas:
Looks good to me, looks like this covers what I try to do with PRs: #208 and #206 - I will close those PR's.

/approve

hjensas (approved):

PR #221, opened 2023-03-06
Allow deploying the control plane with Galera instead of MariaDB

dciabrin:
A new Galera subresource is declared, disabled by default. It exposes similar specs as the MariaDB CR (including adoption).

MariaDB must be disabled in the controlplane CR if Galera is enabled. This is enforced by webhooks.


abays (commented):
+1 from me

dprince (approved):

PR #222, opened 2023-03-07
Default Galera to enabled=false

abays:
The API processing treats boolean values set to `false` as if they are empty, so it triggers the kubebuilder `default` annotation to set it to `true`.  Doing so causes both MariaDB and Galera to be enabled, which hits a validation error.

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_types.go
+++ b/apis/core/v1beta1/openstackcontrolplane_types.go
@@ -169,7 +169,7 @@ type MariadbSection struct {
 // GaleraSection defines the desired state of Galera services
 type GaleraSection struct {
 	// +kubebuilder:validation:Optional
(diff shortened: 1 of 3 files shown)


apis/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -169,7 +169,7 @@ type MariadbSection struct {
 // GaleraSection defines the desired state of Galera services
 type GaleraSection struct {
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default=true
+	// +kubebuilder:default=false
```
bogdando, line 172: according to https://github.com/openstack-k8s-operators/openstack-operator/pull/223, omitempty should be removed as well

PR #223, opened 2023-03-08
Remove omitempty from <ServiceOperator>.enabled

gibizer:
In the following definition
```golang
type <Service> struct {
	// +kubebuilder:default=true
	// Enabled - Whether <Service> service should be deployed and managed
	Enabled bool `json:"enabled,omitempty"`
```
omitempty causes that if the enabled is set to false then it is serialized to json then the json output will not have an enabled field as false is the empty golang value for the bool type. Then if that json is deserialized then the missing enabled field will be defaulted to true based on the kubebuilder:default. This means the above definition does not allow setting enabled to false.

Setting enable to false worked before so I think the turning on of the webhook triggers the extra deserialization / serialization step in the input processing causing the issue.

This patch removes the omitempty from the enabled fields to prevent the issue.

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_types.go
+++ b/apis/core/v1beta1/openstackcontrolplane_types.go
@@ -111,7 +111,7 @@ type KeystoneSection struct {
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=true
 	// Enabled - Whether Keystone service should be deployed and managed
-	Enabled bool `json:"enabled,omitempty"`
+	Enabled bool `json:"enabled"`
 
 	// +kubebuilder:validation:Optional
 	// Template - Overrides to use when creating the Keystone service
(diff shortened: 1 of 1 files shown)


gibizer:
/hold
I need to test this manually to see if this resolve the issue as we don't have automated test coverage for this feature.

gibizer:
/unhold

I tested this manually and with this PR enabled can be set to false again.  I only tested the Nova.enabled but I assume it works the same for the rest of the services. (If you want you can use the build `OPENSTACK_IMG=quay.io/gibi/openstack-operator-index:v0.0.52 make openstack` to test this further)

dprince (approved):

apis/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -171,7 +171,7 @@ type GaleraSection struct {
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=false
```
bogdando, line 172: technically, we can keep omitempty for this one, as it defaulted to the bool type nil value?

gibizer: technically yes.

PR #226, opened 2023-03-10
Add Horizon to OpenStackControlPlane

bshephar:
This change adds the necessary pieces to deploy the Horizon operator.


bshephar:
I believe the operator-build-deploy job failure will be resolved by this patch:
https://github.com/openstack-k8s-operators/horizon-operator/pull/58

gibizer:
/test golangci
golangci-lint runs in pre-commit as well and it passed there. (yes we eventually remove the duplication of jobs)

gibizer (commented):
When I tried to deploy this I saw that the horizon controller_manager restarts with
```
W0313 08:40:08.044118       1 reflector.go:424] pkg/mod/k8s.io/client-go@v0.26.2/tools/cache/reflector.go:169: failed to list *v1beta1.Memcached: memcacheds.memcached.openstack.org is forbidden: User "system:serviceaccount:openstack:horizon-operator-controller-manager" cannot list resource "memcacheds" in API group "memcached.openstack.org" at the cluster scope
```

Do you see that if you try to deploy horizon alone or is it specific to this integration patch that causing the missing permission?

gibizer:
> When I tried to deploy this I saw that the horizon controller_manager restarts with
> 
> ```
> W0313 08:40:08.044118       1 reflector.go:424] pkg/mod/k8s.io/client-go@v0.26.2/tools/cache/reflector.go:169: failed to list *v1beta1.Memcached: memcacheds.memcached.openstack.org is forbidden: User "system:serviceaccount:openstack:horizon-operator-controller-manager" cannot list resource "memcacheds" in API group "memcached.openstack.org" at the cluster scope
> ```
> 
> Do you see that if you try to deploy horizon alone or is it specific to this integration patch that causing the missing permission?

I think it is due to https://github.com/openstack-k8s-operators/horizon-operator/issues/62

bshephar:
The only possible thing I can see for the golangci error is that it's related to a bug that was fixed in 1.46.2 and we're running these tests with 1.46.0:
```
INFO[2023-03-13T08:24:19Z] Tagging ci/golangci-lint:v1.46.0 into pipeline:golangci-lint.
```

A similar issue appears is reported here:
https://github.com/golangci/golangci-lint/issues/2851

Which was apparently fixed by:
https://github.com/1uf3/execinquery/pull/13/commits/bef5bea9ead17efccc61651464cfa4cf492c5eb9

With the fix being listed in 1.46.2:
https://github.com/golangci/golangci-lint/blob/master/CHANGELOG.md#v1462

When we execute `make golangci` locally, it is pulling the latest version of `golangci-lint`:
```
+ http_download_curl /tmp/tmp.awuaaPCkPc/golangci-lint-1.51.2-linux-amd64.tar.gz https://github.com/golangci/golangci-lint/releases/download/v1.51.2/golangci-lint-1.51.2-linux-amd64.tar.gz
+ local_file=/tmp/tmp.awuaaPCkPc/golangci-lint-1.51.2-linux-amd64.tar.gz
```

So, my assumption is that the issue being reported in the golangci job is related to an issue with the `golangci-lint` version in-use. I would be interested to know if it's at all possible to bump this version?

gibizer:
> golangci-lint runs in pre-commit as well and it passed there. (yes we eventually remove the duplication of jobs)

I was mistaken, here in openstack-operator we don't run golangci-lint in pre-commit :/

gibizer:
> So, my assumption is that the issue being reported in the golangci job is related to an issue with the `golangci-lint` version in-use. I would be interested to know if it's at all possible to bump this version?

The job uses a golangci-lint specific container:

https://github.com/openshift/release/blob/ef2dca28318a6f0bd6257813165ffcdefb062e9d/ci-operator/config/openstack-k8s-operators/openstack-operator/openstack-k8s-operators-openstack-operator-master.yaml#L6-L9

but I'm not sure where that is coming from

viroel:
/test golangci

golangci-lint bumped to v1.48.0 in [1]

[1] https://github.com/openshift/release/pull/37271

gibizer (commented):
looks good to me

fmount:
As per our conversation, this PR is currently blocked due to the bundle size limit that generates a `request too large` to `etcd`, preventing OLM to move forward w/ the `InstallPlan`.
The strategy we agreed is to move to a logical split of the openstack components in at least three different bundles: `openstack-core`, `openstack-storage`, `openstack-extra(s)`, and @dprince is working on it.
The patch [1] describes what we've done with manila, that was just an example to prove that we were able to successfully deploy the `OpenStackControlPlane` as long as we have `>1 Job` deploying the static resources (`CRDs`, `rbac`, etc): the current work is a generalized way to pack a set of operators together, in a way that we can scale (in a long term view) without hitting the size limit again. 

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/243

fmount (changes requested):

config/samples/core_v1beta1_openstackcontrolplane.yaml
```diff
@@ -114,6 +114,10 @@ spec:
       databaseInstance: openstack
       containerImage: quay.io/tripleozedcentos9/openstack-neutron-server:current-tripleo
       secret: osp-secret
+  horizon:
```
stuggi, line 117: please also update the network isolation samples

custom-bundle.Dockerfile
```diff
@@ -13,6 +13,7 @@ ARG ANSIBLEEE_BUNDLE=quay.io/openstack-k8s-operators/openstack-ansibleee-operato
 ARG DATAPLANE_BUNDLE=quay.io/openstack-k8s-operators/dataplane-operator-bundle:latest
 ARG NOVA_BUNDLE=quay.io/openstack-k8s-operators/nova-operator-bundle:latest
 ARG IRONIC_BUNDLE=quay.io/openstack-k8s-operators/ironic-operator-bundle:latest
+ARG HORIZON_BUNDLE=quay.io/openstack-k8s-operators/horizon-operator-bundle:latest
```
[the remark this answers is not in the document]
fmount: @stuggi with split we're going to have the `dependencies.yaml` [1], where a version can be specified for each bundle. 
In theory, grouping operators in an `openstack-storage` bundle means that the specific operator version (let's say manila for instance) is hidden and only visible/specified within the `openstack-storage` bundle.
The other key change is creating an index where >1 bundle are added/concatenated [2] (this will trigger within OLM the additional job).
Of course where you see `manila` in [1] [2] in a follow up will be a bundle that represents a group of operators (e.g. core, storage, extra).
@dprince to add more in case I'm overlooking something relevant.

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/dependencies.yaml#L5
[2] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/.github/create_opm_index.sh#L13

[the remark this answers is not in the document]
stuggi: > @stuggi with split we're going to have the `dependencies.yaml` [1], where a version can be specified for each bundle. In theory, grouping operators in an `openstack-storage` bundle means that the specific operator version (let's say manila for instance) is hidden and only visible/specified within the `openstack-storage` bundle. The other key change is creating an index where >1 bundle are added/concatenated [2] (this will trigger within OLM the additional job). Of course where you see `manila` in [1] [2] in a follow up will be a bundle that represents a group of operators (e.g. core, storage, extra). @dprince to add more in case I'm overlooking something relevant.
> 
> [1] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/dependencies.yaml#L5 [2] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/.github/create_opm_index.sh#L13

@fmount ack, thanks for the pointer to the dependencies.yaml , I foregot about. Now I remember it was mentioned in the call we had last week. So if I got it correct, we'd have to update [2] to support pulling a custom bundle from e.g. my quay pointing to a specific version. not sure how this is done via the dependencies.yaml, but I think it is to early at this point and we'll see in the next weeks

config/samples/core_v1beta1_openstackcontrolplane.yaml
bshephar: ack. Done

custom-bundle.Dockerfile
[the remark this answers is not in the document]
bshephar: I've restored this file now that we're going to split it out.

custom-bundle.Dockerfile
```diff
@@ -48,6 +49,7 @@ FROM $DATAPLANE_BUNDLE as dataplane-bundle
 FROM $NOVA_BUNDLE as nova-bundle
 FROM $IRONIC_BUNDLE as ironic-bundle
 FROM $OPENSTACK_STORAGE_BUNDLE as openstack-storage-bundle
+FROM $HORIZON_BUNDLE as horizon-bundle
```
fmount, line 52: @dprince are we ok to merge the `horizon` bundle in the `openstack-operator`?

fmount (changes requested):

config/samples/core_v1beta1_openstackcontrolplane.yaml
fmount: @bshephar not sure I see the network-isolation related samples updated in this patch

fmount:
From the `build-deploy` job:

- The operator looks healthy [1]
- It's an optional component (enabled is false by default): we don't see any pod here [2], which is expected

I had a minor comment on the `samples/` area for network-isolation, otherwise looks good.


[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_openstack-operator/226/pull-ci-openstack-k8s-operators-openstack-operator-master-openstack-operator-build-deploy/1643379112682721280/artifacts/openstack-operator-build-deploy/openstack-k8s-operators-deploy/build-log.txt

[2] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_openstack-operator/226/pull-ci-openstack-k8s-operators-openstack-operator-master-openstack-operator-build-deploy/1643379112682721280/artifacts/openstack-operator-build-deploy/openstack-k8s-operators-gather/artifacts/pods/

fmount (approved):

dprince (approved):

PR #229, opened 2023-03-13
Cleanup delete code path

gibizer:
The PR #205 simplified the deletion code path. Now the OpenStackControlPlane CR deletion is fully automatic as there is no Finalizer on it. So the reconciler will get an NotFound when called to reconcile a deleted instance instead of seeing the instance with delete timestamp set. So that dead code is removed and logging in moved from it to the NotFound handler so that we still log something when the resource is deleted.


gibizer:
```
[33mWARN[0m[2023-03-13T10:42:20Z] Found 1 conditions for ClusterClaim:

  *[2023-03-13T09:42:23Z]NoClusters: No clusters in pool are ready to be claimed 
```

abays (approved):
/lgtm

PR #230, opened 2023-03-13
PoC: Add OpenStackDataPlane

fao89:
Depends on:
- https://github.com/openstack-k8s-operators/dataplane-operator/pull/77
- https://github.com/openstack-k8s-operators/dataplane-operator/pull/72


slagle:
I think this belongs in dataplane-operator for the time being. We already have this CRD there, and I think we can do this integration with keeping it all in dataplane-operator. @dprince what do you think?

apis/core/v1beta1/openstackdataplane_types.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// OpenStackDataPlaneSpec defines the desired state of OpenStackDataPlane
+type OpenStackDataPlaneSpec struct {
+
+	// +kubebuilder:validation:Optional
+	// Nodes - List of nodes
```
slagle, line 33: map?

apis/core/v1beta1/openstackdataplane_types.go
```diff
@@ -0,0 +1,109 @@
...
+	Nodes map[string]dataplanev1beta1.OpenStackDataPlaneNodeSpec `json:"nodes,omitempty"`
+	// +kubebuilder:validation:Optional
+	// Roles - List of roles
```
slagle, line 36: map?

apis/core/v1beta1/openstackdataplane_types.go
```diff
@@ -0,0 +1,109 @@
...
+	Roles map[string]dataplanev1beta1.OpenStackDataPlaneRoleSpec `json:"roles,omitempty"`
```
slagle, line 37: Maintaining the list of nodes and roles here wasn't exactly how I was thinking this would work, however I'm open to it. I thought we would just label the roles with the dataplane they are in like we do for nodes in a role.

With how you have it defined here, we could create the DataPlane all in one YAML doc, which might make for a nicer UX interface. Alternatively, you need to create the DataPlane, Role, and Nodes separately and then trigger a deploy.

Another thought I've had is to associate these items with other top level CRDs, DataPlaneBinding that would group roles in a dataplane, and DataPlaneRoleBinding that would group nodes in a role.

Either way, we need the DeployStrategy field here as well so we can trigger deploys on the DataPlane CR.

controllers/core/openstackdataplane_controller.go
```diff
@@ -0,0 +1,148 @@
...
+
+	ctrlResult, err := openstack.ReconcileDataPlane(ctx, instance, helper)
+	if err != nil {
+		return ctrl.Result{}, err
+	} else if (ctrlResult != ctrl.Result{}) {
+		return ctrlResult, nil
+	}
+
+	return ctrl.Result{}, nil
+}
```
slagle, line 139: we do need the deploy logic in this function. It should just be looping over the roles, and calling the Deploy func for each passing it the already generated role inventory.

apis/core/v1beta1/openstackdataplane_types.go
fao89: oh yeah, I forgot to update

apis/core/v1beta1/openstackdataplane_types.go
fao89: oh yeah, I forgot to update

apis/core/v1beta1/openstackdataplane_types.go
fao89: DataPlaneRoleBinding sounds more idiomatic to the kubernetes world, and the label approach is more consistent with the code we already have

controllers/core/openstackdataplane_controller.go
fao89: I was trying to follow the ControlPlane approach, but yeah, it makes sense to call the deploy func

fao89:
> I think this belongs in dataplane-operator for the time being. We already have this CRD there, and I think we can do this integration with keeping it all in dataplane-operator. @dprince what do you think?

in case we keep everything in the dataplane-operator, the openstack/dataplane operator integration would be just the CSV merging? I'm wondering what would be the DoD of OSP-22547

slagle:
> > I think this belongs in dataplane-operator for the time being. We already have this CRD there, and I think we can do this integration with keeping it all in dataplane-operator. @dprince what do you think?
> 
> in case we keep everything in the dataplane-operator, the openstack/dataplane operator integration would be just the CSV merging? I'm wondering what would be the DoD of [OSP-22547](https://issues.redhat.com//browse/OSP-22547)

It's making sure the csv is merged in and the CRD's are exposed top level as was done here, https://github.com/openstack-k8s-operators/openstack-operator/commit/cf507340b4a72586fbdcd1b28f74080b6a73e559

fao89:
closing as it belongs to the dataplane-operator

PR #233, opened 2023-03-14
Fix netwok isolation sample

gibizer:
The current sample fails to apply due to missing ironic pieces:
```
  The OpenStackControlPlane "openstack-network-isolation" is invalid:
  spec.ironic.template.ironicConductors: Invalid value: "null":
  spec.ironic.template.ironicConductors in body must be of type array: "null"
```
So this patch copies over the ironic part of the default sample to the network isolation sample.


stuggi (approved):
/lgtm

PR #234, opened 2023-03-14
only exclude local ./apis module replace

stuggi:
The api module in the infra-operator is named apis, which right now gets excluded. This breaks pinning the infra operator to a custom version.


abays (approved):
/lgtm

PR #242, opened 2023-03-22
Add sample for network isolation using ceph for glance and cinder

stuggi:
Adds a sample to deploy the ctplane using isolated networks and ceph. The ceph configs get attached from ceph-conf-files secret. The _FSID_ in the cinder custome config needs to be replaces with the one from the ceph cluster:
```
rbd_secret_uuid=_FSID_
```

### DIFF
--- a/config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml
+++ b/config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml
@@ -1,0 +1,200 @@
+apiVersion: core.openstack.org/v1beta1
+kind: OpenStackControlPlane
+metadata:
+  name: openstack-network-isolation
+spec:
+  secret: osp-secret
+  storageClass: local-storage
+  extraMounts:
+    - name: v1
+      region: r1
+      extraVol:
+        - propagation:
+          - CinderVolume
+          - CinderBackup
+          - GlanceAPI
+          extraVolType: Ceph
+          volumes:
+          - name: ceph
+            projected:
+              sources:
+              - secret:
+                  name: ceph-conf-files
+          mounts:
+          - name: ceph
+            mountPath: "/etc/ceph"
+            readOnly: true
+  cinder:
+    template:
+      cinderAPI:
+        containerImage: quay.io/tripleozedcentos9/openstack-cinder-api:current-tripleo
+        externalEndpoints:
+        - endpoint: internal
+          ipAddressPool: internalapi
+          loadBalancerIPs:
+          - 172.17.0.80
+      cinderBackup:
+        containerImage: quay.io/tripleozedcentos9/openstack-cinder-backup:current-tripleo
+        customServiceConfig: |
+          [DEFAULT]
+          backup_driver = cinder.backup.drivers.ceph.CephBackupDriver
+          backup_ceph_pool = backups
+          backup_ceph_user = openstack
+        networkAttachments:
+        - storage
+      cinderScheduler:
+        containerImage: quay.io/tripleozedcentos9/openstack-cinder-scheduler:current-tripleo
+      cinderVolumes:
+        ceph:
+          containerImage: quay.io/tripleozedcentos9/openstack-cinder-volume:current-tripleo
+          customServiceConfig: |
+            [DEFAULT]
+            enabled_backends=ceph
+            [ceph]
+            volume_backend_name=ceph
+            volume_driver=cinder.volume.drivers.rbd.RBDDriver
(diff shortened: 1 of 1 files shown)


config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml
```diff
@@ -0,0 +1,216 @@
...
+            [DEFAULT]
+            enabled_backends=ceph
+            [ceph]
+            volume_backend_name=ceph
+            volume_driver=cinder.volume.drivers.rbd.RBDDriver
+            rbd_ceph_conf=/etc/ceph/ceph.conf
+            rbd_user=openstack
+            rbd_pool=volumes
+            rbd_flatten_volume_from_snapshot=False
+            rbd_secret_uuid=_FSID_
```
fultonj, line 41: Eventually, you should be able to remove line 41 `rbd_secret_uuid=_FSID_` because of this patch:

  https://review.opendev.org/c/openstack/cinder/+/860090

@Akrog @ASBishop We need to wait for Antelope before we can do this, right?

Oddly, I left it out and cinder worked for me with zed.

  https://github.com/fultonj/zed/blob/main/edpm/kustomize/kustomization.yaml#L39

I don't understand why yet however, so I think we should keep this as @stuggi has it. I was using zed

    image: quay.io/tripleozedcentos9/openstack-cinder-volume:current-tripleo

fultonj (commented):
I support merging this example +2

config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml
```diff
@@ -0,0 +1,216 @@
...
+          networkAttachments:
+          - storage
+          replicas: 0 # backend needs to be configured
+      extraMounts:
```
fmount, line 45: @stuggi is there a particular reason to define these `extraMounts` here instead of having them in the top-level `spec` ?
If you define them at `openstack-operator` level you can use the same syntax to propagate the volumes to both `Cinder` and `Glance` (but you define them only once):

```
   - propagation:
              - CinderVolume
              - CinderBackup
              - GlanceAPI
```

fmount (changes requested):

config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml
stuggi: @fmount ah no, just forgot about the top-level one. will change it, thanks!

config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml
ASBishop: @fultonj Not sure of what you were testing when you wrote "cinder worked for me with zed," but bear in mind the rbd_secret_uuid is required for nova's side of things. Without that option, you can create volumes but nova won't be able to attach them.

At present, that patch is only available in Antelope. I'll ask around and see how cinder people feel about a stable backport.

config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml
```diff
@@ -0,0 +1,200 @@
...
+          [DEFAULT]
+          backup_driver = cinder.backup.drivers.ceph.CephBackupDriver
+          backup_ceph_pool = backups
+          backup_ceph_user = openstack
+        networkAttachments:
+        - storage
+      cinderScheduler:
+        containerImage: quay.io/tripleozedcentos9/openstack-cinder-scheduler:current-tripleo
+      cinderVolumes:
+        volume1:
```
ASBishop, line 48: Since "volume1" was just the original name that predates recent ceph support, could this be changed to "ceph"?

config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml
fmount: Thanks, +1 on my side, we can merge this sample as long as it's ok for the other storage folks!

fultonj:
One nit about a comment, not the PR itself.
```
  sed -i "s/FSID/${FSID}/" $OPENSTACK_CTLPLANE
```
vs
```
  sed -i "s/_FSID_/${FSID}/" $OPENSTACK_CTLPLANE
```
You want the second so you don't end up with a line like this:
```
            rbd_secret_uuid=_b07ec152-274d-4c8c-b939-32362336ee9f_
```

fmount (approved):

PR #246, opened 2023-03-23
Disable cinder backup & volume

Akrog:
Cinder backup and volume services require the storage backend configuration to work:

- Backup won't complain on start but it will fail on operations.
- Volume will be in CrashLoopBackOff forever.

This patch changes config samples that don't have configured backends by setting the replicas of the services to 0.


dprince:
@Akrog thanks! looks good to me

dprince (approved):

PR #247, opened 2023-03-23
Openstack k8s operators, drop alm-examples for now

dprince:


fao89:
we can drop them on dataplane, we just need to remove the samples from here: https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/kustomization.yaml

fao89 (approved):

fao89:
I believe this PR won't be needed, it seems the issue is having variables inside the alm-examples: https://github.com/openstack-k8s-operators/dataplane-operator/pull/124#discussion_r1146790848

PR #249, opened 2023-03-24
Fix network isolation CR by adding manila

fao89:
```
$make openstack_deploy
oc kustomize /home/faguiard/project/install_yamls/out/openstack/openstack/cr | oc apply -f -
The OpenStackControlPlane "openstack-network-isolation" is invalid: spec.manila.template.manilaShares: Invalid value: "null": spec.manila.template.manilaShares in body must be of type object: "null"
make: *** [Makefile:338: openstack_deploy] Error 1
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ custom-bundle.Dockerfile.pinned
 
 # editor and IDE paraphernalia
 .idea
+.vscode
 *.swp
 *.swo
 *~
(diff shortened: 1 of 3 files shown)


stuggi (approved):
/lgtm

karelyatin (commented):
i see there are more samples like galera,seperate cell, collapsed cell those don't need similar change?

dprince (approved):

gibizer (approved):
Thanks.

PR #253, opened 2023-03-27
Fix Cinder templates in samples

abays:
Right now we are seeing this error in CI:

```
+ oc kustomize /alabama/install_yamls/out/openstack/openstack/cr/
+ oc apply -f -
The OpenStackControlPlane "openstack" is invalid: 
* spec.cinder.template.databaseInstance: Required value
* spec.cinder.template.secret: Required value
```

This is due to the fact that `databaseInstance` and `secret` are not included in the OpenStack operator `config/samples` `Cinder` `template` blocks.  Other service operators either have none, one or both of these fields included, depending on whether or not those fields are required in their respective CRDs.  This of course leads to a question of what the proper optional/required designation should ultimately be, such that we can properly achieve consistency across all the service operators.  But for now we're patching the samples here to fix CI.

### DIFF
--- a/config/samples/core_v1beta1_openstackcontrolplane.yaml
+++ b/config/samples/core_v1beta1_openstackcontrolplane.yaml
@@ -60,6 +60,8 @@ spec:
         containerImage: quay.io/tripleozedcentos9/openstack-glance-api:current-tripleo
   cinder:
     template:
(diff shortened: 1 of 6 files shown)


fmount (approved):
looks right, thanks!

PR #254, opened 2023-03-27
Expose NovaMetadata to internalapi

gibizer:
This PR change the network isolation sample to expose nova-metadata service automatically on the internalapi network so that the ovn-metadata agent can access it from the compute.


fao89 (commented):
please add it to: https://github.com/openstack-k8s-operators/openstack-operator/blob/master/config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml

gibizer:
> please add it to: https://github.com/openstack-k8s-operators/openstack-operator/blob/master/config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml

good point. Done.

fao89 (approved):

PR #257, opened 2023-03-29
Bump nova- and dataplane-operator

gibizer:
We landed couple of related patches there. So this patch bumps the versions together.

We cannot rely on renovate to bump these deps as bumping all the deps together pushes us over the bundle size limit as seen in the test of #255


fao89 (approved):
thank you!

abays (approved):
/lgtm

PR #262, opened 2023-04-03
Connect ovs pod to the tenant network

slawqo:
In samples with network isolation enabled, lets connect ovs pods to the tenant network, instead of the internalapi. OVS pod can communicate with OVN DBs using internal OCP network provided to the PODs and tenant network attached to the pod using multus will be then used to establish tunnels with EDPM nodes.

Related: #[OSP-20056](https://issues.redhat.com//browse/OSP-20056)


bshephar (approved):
Sounds like a good plan

stuggi (approved):
/lgtm

PR #278, opened 2023-04-20
Set k8s binary as an option in order to be overriden by OCP bin

matbu:
While installing openstack-operator on Openshift cluster, it can be necessary to override the kubectl binary by the Openshift bin.


stuggi:
what I usually do is `alias kubectl=oc`

PR #279, opened 2023-04-20
Updated config samples for galera deployment

dciabrin:
Sync the galera config sample to match the default deployment. Add a deployment sample for 3-nodes galera clusters.


abays (approved):
/lgtm

PR #282, opened 2023-04-21
Container img defaults for Keystone and MariaDB

abays:
Wire-in defaulting logic for Keystone and MariaDB now that we're using the latest versions of them (where `containerImage` is no longer defaulted via `kubebuilder` annotations, but instead is handled via defaulting webhooks).


ASBishop (commented):
+1, fixed the issue with my deployment

gibizer (approved):
looks good

PR #283, opened 2023-04-24
[samples]Remove defaulted containerImage fields

gibizer:
We have defaulting webhook to set the containerImage fields so to actually test that webhook in the e2e job this patch removes the explicit values from the containerImage fields from our samples for those services where the webhook integration is already done in openstack-operator.

### DIFF
--- a/config/samples/core_v1beta1_openstackcontrolplane.yaml
+++ b/config/samples/core_v1beta1_openstackcontrolplane.yaml
@@ -7,28 +7,23 @@ spec:
   storageClass: local-storage
   keystone:
     template:
-      containerImage: quay.io/tripleozedcentos9/openstack-keystone:current-tripleo
       databaseInstance: openstack
       secret: osp-secret
   mariadb:
     templates:
       openstack:
-        containerImage: quay.io/tripleozedcentos9/openstack-mariadb:current-tripleo
         storageRequest: 500M
       openstack-cell1:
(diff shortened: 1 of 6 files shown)


gibizer:
Does sub CRs in Glance fails to default?

```
+ oc kustomize /alabama/install_yamls/out/openstack/openstack/cr/
+ oc apply -f -
The OpenStackControlPlane "openstack" is invalid: 
* spec.glance.template.glanceAPIInternal.apiType: Unsupported value: "": supported values: "internal", "external"
* spec.glance.template.glanceAPIExternal.apiType: Unsupported value: "": supported values: "internal", "external"
```

gibizer:
> Does sub CRs in Glance fails to default?
> 
> ```
> + oc kustomize /alabama/install_yamls/out/openstack/openstack/cr/
> + oc apply -f -
> The OpenStackControlPlane "openstack" is invalid: 
> * spec.glance.template.glanceAPIInternal.apiType: Unsupported value: "": supported values: "internal", "external"
> * spec.glance.template.glanceAPIExternal.apiType: Unsupported value: "": supported values: "internal", "external"
> ```

It is probably fixed by https://github.com/openstack-k8s-operators/glance-operator/pull/212 but we need to bump that glance version into openstack-operator first.

gibizer:
the latest glance-operator is now used in openstack-operator so this PR is rebased and expected to be green.

abays (approved):
/lgtm

PR #285, opened 2023-04-25
Update to latest Glance

abays:
Update OpenStack operator to use the latest Glance (where `APIType` is removed from `GlanceAPITemplate`  and moved instead to `GlanceAPI.Spec`)


gibizer (approved):
Looks OK to me. The samples has no apitType defined so nothing to remove from there.

abays:
This was accomplished via another PR and is no longer neeed

PR #286, opened 2023-04-26
Add openstack-baremetal-operator to openstack-operator bundle

rabi:
We already have dataplane-operator that needs it for baremetal provisioning. We probably should create separate bundle for dataplane-operator that includes it in the future.


abays (approved):
/lgtm

PR #287, opened 2023-04-26
Add support for using local registry for bundle build

raukadah:
registry:2 container image is used to spin a local registry.
It supports v2 api.
    
The Quay registry supports both v1 and v2 api. While using v2 api
on quay registry, it does not list all the tags which breaks the
scripts. That's why we have not moved the existing api to v2.
    
registry:2 supports only v2. for making sure both quay and local
registry works together.
    
This patch adds LOCAL_REGISTRY flag to set the v2 api.
By setting LOCAL_REGISTRY=1, v2 api for local registry
can be used.

### DIFF
--- a/hack/pin-custom-bundle-dockerfile.sh
+++ b/hack/pin-custom-bundle-dockerfile.sh
@@ -5,6 +5,7 @@ DOCKERFILE=${DOCKERFILE:-"custom-bundle.Dockerfile"}
 IMAGENAMESPACE=${IMAGENAMESPACE:-"openstack-k8s-operators"}
 IMAGEREGISTRY=${IMAGEREGISTRY:-"quay.io"}
 IMAGEBASE=${IMAGEBASE:-}
+LOCAL_REGISTRY=${LOCAL_REGISTRY:-0}
 
 cp "$DOCKERFILE" "${DOCKERFILE}.pinned"
 
@@ -23,14 +24,25 @@ for MOD_PATH in $(go list -m -json all | jq -r '. | select(.Path | contains("ope
     REPO_URL="quay.io/openstack-k8s-operators"
     if [[ "$GITHUB_USER" != "openstack-k8s-operators" || "$BASE" == "$IMAGEBASE" ]]; then
         if [[ "$IMAGENAMESPACE" != "openstack-k8s-operators" || "${IMAGEREGISTRY}" != "quay.io" ]]; then
-            REPO_CURL_URL="https://${IMAGEREGISTRY}/api/v1/repository/${IMAGENAMESPACE}"
             REPO_URL="${IMAGEREGISTRY}/${IMAGENAMESPACE}"
+            # Quay registry v2 api does not return all the tags that's why keeping v1 for quay and v2
+            # for local registry
+            if [[ ${LOCAL_REGISTRY} -eq 1 ]]; then
+                REPO_CURL_URL="${IMAGEREGISTRY}/v2/${IMAGENAMESPACE}"
+            else
+                REPO_CURL_URL="https://${IMAGEREGISTRY}/api/v1/repository/${IMAGENAMESPACE}"
+            fi
         else
             REPO_CURL_URL="https://quay.io/api/v1/repository/${GITHUB_USER}"
             REPO_URL="quay.io/${GITHUB_USER}"
         fi
     fi
 
-    SHA=$(curl -s ${REPO_CURL_URL}/$BASE-operator-bundle/tag/ | jq -r .tags[].name | sort -u | grep $REF)
+    if [[ ${LOCAL_REGISTRY} -eq 1 && "$BASE" == "$IMAGEBASE" ]]; then
+        SHA=$(curl -s ${REPO_CURL_URL}/$BASE-operator-bundle/tags/list | jq -r .tags[] | sort -u | grep $REF)
+    else
+        SHA=$(curl -s ${REPO_CURL_URL}/$BASE-operator-bundle/tag/ | jq -r .tags[].name | sort -u | grep $REF)
+    fi
+
     sed -i "${DOCKERFILE}.pinned" -e "s|quay.io/openstack-k8s-operators/${BASE}-operator-bundle.*|${REPO_URL}/${BASE}-operator-bundle:$SHA|"
 done


viroel (changes requested):
Hey Chandan, some questions regarding the use of LOCAL_REGISTRY
This script is starting to get more and more complex, we might need to rework this or add proper docs.

raukadah:
@viroel The Quay v2 api does not returns all the tags. If there are less than 50 tags. It returns correctly. But if there are more than 50 tags. It does not return correctly. It is kind of limitation from quay registry side.
I need to make this change in such a way to handle both cases.
```
❯ curl -s https://quay.io/v2/openstack-k8s-operators/cinder-operator-bundle/tags/list | jq -r .tags
[
  "v0.0.1",
  "v0.0.20",
  "ec0b73ab54463be2bab19a4e75ed4da322cd2cab",
  "20062a9a070563c962667d3497ea51b15ccef753",
  "46d3f2e2acb7cac0f00f89235def9fd573a54406",
  "7665a222fcfb6c869c202dc0805a3fd2f18dab70",
  "974e334b286457983f0ee56b77d21881118e77fc",
  "416335d89758779fe13f5de9ee7182f3c68d0874",
  "30664865dad09c090f2dd1862ee3a75429a581c7",
  "7729d141f19014ee0cbbf93eafc2bbdf3fef8984",
  "3b458b6b498d445d8db9e6844b3cef934a33ec32",
  "b7ac98f03fe57e7c0641c02a6f483345004e0ebf",
  "3b8eabe8c2492a832eb82b6b6a97178806e6c4f4",
  "d711bbb1bc97c5957086b990dafd947c5adfa142",
  "4775f09b51dbacd32914a86fdb5db11969135fd2",
  "3d3cbfa539a64d182f71ab1fe52279f18ece9ac8",
  "5ff276d7d8dec6ba9dd850138fca5abfc7d473c0",
  "27f036911829b8675a1708749e0cb841eaa5d6d6",
  "f0ce5ca4f47fb7621373e517920da6759940495e",
  "c391cf8ee423d2e49ccb8751677de401b9f9dafa",
  "37f7e1b2e5855be076f5fb27aee365e36d4707ce",
  "a6cfe836a9a37ce8289f46d26b4ccd02637f5e49",
  "cb242352e51c3ed386c5d2904d5303712c695c7b",
  "2d51c05f38f5cf365f4121dcd44495d4fca3b14e",
  "fa0d531d4d41b0e86c5947dbe3881a70d13c9a59",
  "d8b5cd2360b885bd533f2cb492a104ec2e0f00d3",
  "77b8278c88f1327ff4ccc050f360a33d246b4c47",
  "d8533e93acd92d5a253c1024558098978ec7df8d",
  "731d6beb57ede1896db35e83e3fc84dd3c7a8d84",
  "86505b36274dd3c5a328c0a77e436af62b7c5de8",
  "1c41c6a308449c3669a21f1d1823fe10f865634d",
  "b3459cc22166e9956a492d68dc920fb534ac2b5f",
  "55d4fb9fc656f596baa9e18998813dba2eed92cf",
  "cc2b18c2708cf30bbcb0be28cb81d959e46f7b1c",
  "688f272d10c1ede337e1854819dab408fdc27377",
  "db7bb5a4cdebade8241919a374046e1d8e68be98",
  "34c04d0f0b6a2d5589fe017adae5e0c59d6f05d8",
  "c21e19041ea856fa1b967961db37d437a95d701b",
  "d2af3e4ce866398e0cd457d054b28b7795810000",
  "02ab8b0e4b839af3abefd8760b49672ba9b955eb",
  "a9a8efadf3c3f263ac860ffde8eee054fd940d39",
  "0f82fd6ff45f4c508641e815ea2ff88971d5c93f",
  "db203212132398b434c585bb1975f77ce6074757",
  "64a1fc7bde1393b1776ed4283129ae617ea26e61",
  "d03dea1b2755c1796b6e0f5d41938c1880ceffc6",
  "60bdac1a3f2db7cd1ea1e8a43a3309bc3b4209e7",
  "bcf8790de2f7c0bc9d776cfe5b269ed40db79b3d",
  "aa5c5cbc2a596e1b38e022c1e8e50b1c4f0c60e5",
  "72f7cd5631939f19f05c6266656406b88f8d5f1f",
  "0f27b56154d5cdb6f5ffb66a7eb5e52057b257d5"
]
```

raukadah:
@viroel Based on my testing: https://github.com/openstack-k8s-operators/dataplane-operator/pull/167, here is the logs https://review.rdoproject.org/zuul/buildset/747bfbb9f08b4f979313112172bbc89f and https://logserver.rdoproject.org/67/167/30999236d1ee09f544871a4065bc4ec772ef6b93/github-check/dataplane-operator-content-provider/db15184/local_registry.log
```
{"repositories": ["openstack-k8s-operators/dataplane-operator", "openstack-k8s-operators/dataplane-operator-bundle", "openstack-k8s-operators/dataplane-operator-index", "openstack-k8s-operators/openstack-operator", "openstack-k8s-operators/openstack-operator-bundle", "openstack-k8s-operators/openstack-operator-index", "openstack-k8s-operators/openstack-operator-storage-bundle"]}
```

hack/pin-custom-bundle-dockerfile.sh
```diff
@@ -23,14 +24,25 @@ for MOD_PATH in $(go list -m -json all | jq -r '. | select(.Path | contains("ope
...
+                REPO_CURL_URL="https://${IMAGEREGISTRY}/api/v1/repository/${IMAGENAMESPACE}"
+            fi
         else
             REPO_CURL_URL="https://quay.io/api/v1/repository/${GITHUB_USER}"
             REPO_URL="quay.io/${GITHUB_USER}"
         fi
     fi
 
-    SHA=$(curl -s ${REPO_CURL_URL}/$BASE-operator-bundle/tag/ | jq -r .tags[].name | sort -u | grep $REF)
+    if [[ ${LOCAL_REGISTRY} -eq 1 && "$BASE" == "$IMAGEBASE" ]]; then
```
viroel, line 41: I didn't get why you need `&& "$BASE" == "$IMAGEBASE" ` here, since this script runs against all modules. This would limit it to replace only IMAGEBASE.
We will also need to change IMAGEBASE to be a list, in order to test multiple operators build, before generating the custom bundle. But can be added in another patch.

viroel: Btw, the idea behind IMAGEBASE var is to cover the scenario where bots propose PRs from OKO org. That scenario was not not working since GITHUB_USER is "openstack-k8s-operators" in those patches.

viroel (commented):
The logic is more clear now Chandan, thanks for that. Just a question inline.

hack/pin-custom-bundle-dockerfile.sh
raukadah: @viroel Thank you for reviewing it. I have used ""$BASE" == "$IMAGEBASE"" to set get proper SHA value otherwise if I donot use. SHA logic fails if we set LOCAL_REGISTRY to 1 and use v2 api. It fails the bundle-build process.
I am not sure how can i avoid that. need help here.

viroel: Yeah, that makes sense. Since all modules other modules will be on reg v1, and LOCAL_REGISTRY will be set 1 in all iterations.

viroel (commented):
We can see script outputs here:
https://logserver.rdoproject.org/67/167/30999236d1ee09f544871a4065bc4ec772ef6b93/github-check/dataplane-operator-content-provider/db15184/ci-framework/logs/ci_make_013_openstack-operator_-_call_bundle-build.log

and 

https://logserver.rdoproject.org/67/167/30999236d1ee09f544871a4065bc4ec772ef6b93/github-check/dataplane-operator-content-provider/db15184/local_registry.log

stuggi (commented):
looks good to me. @dprince can you also have a look

dprince:
I'm good with it. Agree complexity here is going up but this is useful and still easy enough to follow.

dprince (approved):

PR #293, opened 2023-05-04
Change suggested-namespace to openstack-operators

dprince:
This change will set the suggested namespace to the openstack-operators which will make the default namespace when installed by olm be named 'openstack-operators'. This will act as a default in the OpenShift web console.


config/webhook/manifests.yaml
```diff
@@ -25,6 +25,7 @@ webhooks:
     resources:
     - openstackcontrolplanes
   sideEffects: None
+  namespaceSelector: {}
```
dprince, line 28: @abays so I figured out if I patch the validatingwebhookconfiguration after installation and set namespaceSelector: {} it works as in openstack services can install into the 'openstack' namespace but the operators themselves reside in the 'openstack-operators' namespace. Ditto required for the mutatingwebhookconfiguration. Setting it this way in the config/webhook/manifests doesn't seem to override how OLM configures it. Still looking for a solution here. See also: https://docs.openshift.com/container-platform/4.12/rest_api/extension_apis/mutatingwebhookconfiguration-admissionregistration-k8s-io-v1.html (namespaceSelector on that page)

Alternative would be I guess we label our openstack namespace and see if we can get it working that way.

dprince: I am dropping these lines. I think it might be targetNamespaces that is restricting us and causing the nameSelectors

abays (approved):
/lgtm

PR #295, opened 2023-05-05
Add Nova defaulting mechanism

abays:
Adds Nova to the defaulting webhook mechanism, as Nova defaulting logic is now available.


main.go
```diff
@@ -259,6 +259,9 @@ func setupServiceOperatorDefaults() {
 
 	neutronv1.SetupNeutronAPIDefaults(neutronAPIDefaults)
 
+	// Nova
+	novav1.SetupDefaults()
+
```
abays, line 264: We will replicate this simplified approach for the other operators as we move forward

PR #298, opened 2023-05-09
Drop containerImages from samples

dprince:


abays (approved):
Looks beautiful to me

PR #301, opened 2023-05-10
Refine ironic template in config samples

steveb:
This change does the following:
- Add missing service config for new service ironicNeutronAgent
- Remove remaining storageClass attributes


abays (approved):
/lgtm

PR #304, opened 2023-05-12
use v1 api to list tags

SeanMooney:


hack/pin-custom-bundle-dockerfile.sh
```diff
@@ -39,7 +39,8 @@ for MOD_PATH in $(go list -m -json all | jq -r '. | select(.Path | contains("ope
     fi
 
     if [[ ${LOCAL_REGISTRY} -eq 1 && "$BASE" == "$IMAGEBASE" ]]; then
-        SHA=$(curl -s ${REPO_CURL_URL}/$BASE-operator-bundle/tags/list | jq -r .tags[] | sort -u | grep $REF)
+        SHA=$(curl -s ${REPO_CURL_URL}/$BASE-operator-bundle/tags/list | tee tags | jq -r .tags[] | sort -u | grep $REF)
+        cat tags
```
SeanMooney, line 43: there is something wrong with the local registry support
this fails to find the nova-opertor and i expect it would fail for any operators built and pushed to the local regisry 

im wondering is there an auth issue since based on https://docs.docker.com/registry/spec/api/#get-tags this looks right so im teeing the responce to tag and cating it to see what the responce has

SeanMooney: oh its not taking this branch its using the other barnch 
https://review.rdoproject.org/zuul/build/762415ad6f434a07a5bfb912beadc42a/log/ci-framework/logs/ci_make_011_openstack-operator_-_call_bundle.log#314-315

hack/pin-custom-bundle-dockerfile.sh
```diff
@@ -36,13 +36,14 @@ for MOD_PATH in $(go list -m -json all | jq -r '. | select(.Path | contains("ope
...
             REPO_URL="quay.io/${GITHUB_USER}"
         fi
+        if [[ ${LOCAL_REGISTRY} -eq 1 ]]; then
+            SHA=$(curl -s ${REPO_CURL_URL}/$BASE-operator-bundle/tags/list | jq -r .tags[] | sort -u | grep $REF)
+        else
+            SHA=$(curl -s ${REPO_CURL_URL}/$BASE-operator-bundle/tag/ | jq -r .tags[].name | sort -u | grep $REF)
+        fi
     fi
 
-    if [[ ${LOCAL_REGISTRY} -eq 1 && "$BASE" == "$IMAGEBASE" ]]; then
```
SeanMooney, line 41: ok the base check is required 


the problem is basically that when using the content provider job we are using a local regiesty but then "BASE" == "$imageBASE" is not true for the nova operator.

and sha end ups being empty because we take the wrong path.

but if i move the around the it fails on the cinder operator because it is not using the local registry.

SeanMooney: perhapse instead i need to check "$GITHUB_USER" != "openstack-k8s-operators" ||

on the PR where the content provider job is running the $GITHUB_USER when processing the nova operator (since the pr was against nova) is actually my user or rather the user of the source of the pr rhater then the repo its being submitted too.

SeanMooney: if we look at the first run that got this far before i modified the code

https://review.rdoproject.org/zuul/build/bb2d1bddb52f49a9b9da2d343234aa29/log/ci-framework/logs/ci_make_011_openstack-operator_-_call_bundle.log#295-316

we can see it took the wrong branch and used the v1 API 
and that BASE=nova and GITHUB_USER=SeanMooney

we can also see where it check if we were uign the local regiesty
 if [[ ${LOCAL_REGISTRY} -eq 1 && "$BASE" == "$IMAGEBASE" ]]; then
+ [[ 1 -eq 1 ]]
+ [[ nova == '' ]]

so for some reason IMAGEBASE was not set

we can see it starts unset 

https://review.rdoproject.org/zuul/build/bb2d1bddb52f49a9b9da2d343234aa29/log/ci-framework/logs/ci_make_011_openstack-operator_-_call_bundle.log#23

and is never assigned too
so "$BASE" == "$IMAGEBASE" is not working its always false
so it need to be updated with something else.

PR #306, opened 2023-05-15
Move ironic template defaulting to the webhook

hjensas:
Setting the values for Secret, DatabaseInstance and StorageClass in pkg/ironic makes the API result listing the spec for OpenstackControlplane not match actual settings used.

Also, defaulting in the pkg does not work as intended since the webhook call to ironic-operator explicitly set values such as the StorageClass for each conductor in "ironicConductors" - i.e setting StorageClass in the top level "ironics" in pkg/ironic never apply to the conductors. This caused invalid "" be defined for conductors when actual intent was to default it based on what was defined for top level ironics.

This change moves the defaulting of top level ironics properties to the webhook in openstack-controlplane, the defaulting happens before calling the Default() function in the ironic-operator.


son-vyas:
All jobs look green. LGTM!

PR #312, opened 2023-05-16
fix local registry support for PRs

SeanMooney:
This change updates the pin custom bundle script
to use the same critia to determin if the v2 or v1
endpoint should be used when generating the resouce url
jq expression.

Related: [OSPRH-97](https://issues.redhat.com//browse/OSPRH-97)


raukadah (approved):
/approve

stuggi (approved):
/lgtm

PR #314, opened 2023-05-17
Allow 'internal' Nova webhooks to remain enabled

abays:
We have discovered a context in which the current webhook pattern used across our operators is not currently viable.  This happens when the DataPlane operator creates a `NovaExternalCompute` CR in the context of an overall deployment via OpenStack operator.  A problem then arises because...

1. The Nova operator is deployed via OpenStack operator and thus the Nova operator webhooks are disabled
2. The DataPlane operator creates a `NovaExternalCompute` CR
3. Because the `NovaExternalCompute` CR comes from the DataPlane operator and not the OpenStack operator, it avoids the OpenStack operator's defaulting webhooks
4. Without the OpenStack operator's defaulting webhooks firing, no default images can be set on the `NovaExternalCompute` CR
5. And also, since the Nova operator webhooks are disabled, no default images are set via that mechanism as well

The stop-gap solution proposed in this PR is to always have the "internal" Nova webhooks enabled for the Nova operator.  That is:

1. The `ENABLE_WEBHOOKS=false` flag will _not_ be set on the bundled Nova operator deployment, such that the Nova controller-manager pod will still run its webhook server
2. The webhook cert volume for the Nova operator deployment will _not_ be removed, such that the Nova controller-manager service will work properly
3. The `MutatingWebhookConfiguration` definitions from the Nova operator CSV will be merged with the OpenStack operator webhook definitions, such that they are created during OpenStack operator installation via OLM

We will then take time to think of a potential better solution.

### DIFF
--- a/cmd/csv-merger/csv-merger.go
+++ b/cmd/csv-merger/csv-merger.go
@@ -208,30 +208,34 @@ func main() {
 				if container.Name == "manager" {
 					envVarList = append(envVarList, container.Env...)
 
-					// Now we also need to turn off any "internal" webhooks that belong to the service
-					// operator, as we are now using "external" webhooks that live in the OpenStack
-					// operator.  These "external" webhooks will eventually call the mutating/validating
-					// logic that was previously housed within the "internal" Service Operator webhook
-					// logic.
-					container.Env = append(container.Env,
-						v1.EnvVar{
-							Name:  "ENABLE_WEBHOOKS",
-							Value: "false",
-						},
-					)
+					// FIXME: HACK: Allow NovaExternalCompute CRs that are created from the DataPlane operator
+					//              to be properly defaulted via Nova webhooks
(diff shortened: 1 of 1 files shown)


rabi (commented):
/lgtm

Probably one more reason for my earlier argument to keep the defaulting webhooks in respective operators :smile: . We can probably merge this and clear up the blocked edpm jobs and then discuss the way forward.

gibizer (commented):
Thanks @abays for the great summary of the situation. I agree to unblock the EDPM job by this change.

gibizer:
> /lgtm
> 
> Probably one more reason for my earlier argument to keep the defaulting webhooks in respective operators smile . We can probably merge this and clear up the blocked edpm jobs and then discuss the way forward.

I'm not against running the webhooks both in the openstack-operator and in the service operators too. Sure we duplicate the execution of the defaulting / validation logic but we can have the safety of validation even if somebody creates a service CR independently from the OpenstackControlPlane CR.

PR #323, opened 2023-05-24
Add kuttl tests

jparoly:


kuttl-test.yaml
```diff
@@ -0,0 +1,25 @@
...
+# 2. An OCP 4.10+ CRC cluster with Podified Operators has been deployed
+# 3. CLI user has access to $KUBECONFIG
+# 4. The environment variable INSTALL_YAMLS is set to the the path of the
+#    install_yamls repo
+
+apiVersion: kuttl.dev/v1beta1
+kind: TestSuite
+reportFormat: JSON
+reportName: kuttl-test-openstack
+namespace: openstack
```
abays, line 21: I think we're trying to use unique namespaces going forward for all KUTTL tests [1].  If we could get that working here, that would be awesome.  Attempting this would also require changing [2] to not `oc apply` against the `openstack` namespace.

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/769ebc6d0776ca4edf06a2d2e0d9eedcf3906cbb/kuttl-test.yaml#L20
[2] https://github.com/openstack-k8s-operators/openstack-operator/pull/323/files#diff-d760e3bc41b0c740ee3a351bc381a27357a90ef627adaddd61937eaed30020a2

cescgina: @abays yes, that's right, thanks for pointing it out

tests/kuttl/tests/deployments/01-deploy-openstack.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      oc apply -n openstack -f ../../../../config/samples/core_v1beta1_openstackcontrolplane.yaml
```
cescgina, line 5: As @abays pointed out, we should replace `-n openstack` with `-n $NAMESPACE`, so we can run the tests in a different namespace

tests/kuttl/common/assert-sample-deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+        ovndbcluster-nb:
+          replicas: 1
+          dbType: NB
+          storageRequest: 10G
+        ovndbcluster-sb:
+          replicas: 1
+          dbType: SB
+          storageRequest: 10G
+      ovnNorthd:
+        replicas: 1
```
abays, line 78: We should probably rebase this PR to get the latest OVN changes to the `config/samples` and then update the assert sample here to include this in the OVN section:  https://github.com/openstack-k8s-operators/openstack-operator/blob/151dfb03e1c29757911c05ef708e6c5ce8b33333/config/samples/core_v1beta1_openstackcontrolplane_collapsed_cell.yaml#L63-L67

tests/kuttl/common/assert-sample-deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+  ovs:
+    template:
+      external-ids:
+        system-id: "random"
+        ovn-bridge: "br-int"
+        ovn-encap-type: "geneve"
```
abays, line 84: After rebase, also remove this section, since it no longer exists.

tests/kuttl/tests/deployments/02-assert-collapsed-cell.yaml
abays, line 1: I think step 2 here should actually be a separate test.  We've already deployed an `OpenStackControlPlane` in the namespace in step 1.  It doesn't really make sense to deploy another one in the same namespace while the other one is still there.

jparoly: > I think step 2 here should actually be a separate test. We've already deployed an `OpenStackControlPlane` in the namespace in step 1. It doesn't really make sense to deploy another one in the same namespace while the other one is still there.

When you say "separate test" do you mean a separate folder for the test or use a different namespace or something else?

abays: I would use a separate folder if KUTTL is smart enough to treat each sub-folder within the `deployments` folder as a separate test.  Just don't know off of the top of my head.

cescgina: I think that the right structure is to have two folders at the `deployments` folder, so we have two separate test cases, one for each use case

jparoly: done... correctly I think

tests/kuttl/common/assert-sample-deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+          replicas: 1
+          dbType: NB
+          storageRequest: 10G
+        ovndbcluster-sb:
+          replicas: 1
+          dbType: SB
+          storageRequest: 10G
+      ovnNorthd:
+        replicas: 1
+  ovnController:
```
abays, line 79: This needs to be indented like lines 68 and 77

tests/kuttl/tests/deployments/02-assert-collapsed-cell.yaml
abays: Another issue with this is that both deployment tests will be run in succession via `make openstack_kuttl`.  What I am seeing is that the first `basic-deployment` test succeeds, but it leaves the `OpenStackControlPlane` (and all its derivative resources) deployed.  Then the `collapsed` test runs and the resources from the previous test seem to collide with those (causing errors and failing the test).  KUTTL actually would remove resources from the previous test by deleting the namespace _if_ it generates a random namespace, but since we provide the namespace it skips clean-up.  So we need to handle this in some manner, even if it's adding a clean-up step to each test.

tests/kuttl/common/assert-sample-deployment.yaml
abays: @jparoly Not sure if you saw this one about the indentation

jparoly: done, thank you

abays: Still needs further indenting.  `ovnController` should be flush with `ovnDBCluster` and  `ovnNorthd`.

tests/kuttl/tests/deployments/02-assert-collapsed-cell.yaml
abays: @jparoly See also this comment about the tests colliding with one another.  For the meantime, it might be best to just delete the `OpenStackControlPlane` as a clean-up step in each test.  Then we'll need to think about how to properly address this going forward, as it's a problem that all KUTTL tests will encounter once there is more than one test.

jparoly: missed that comment. ok will try out the delete. Please let me know how it looks and if there is something else needed.

tests/kuttl/common/assert-sample-deployment.yaml
jparoly: oy, thanks, sorry about that

tests/kuttl/tests/collapsed/01-assert-collapsed-cell.yaml
```diff
@@ -0,0 +1,112 @@
...
+  name: openstack-collapsed-cell
+spec:
+  secret: osp-secret
+  keystone:
+    template:
+      databaseInstance: openstack
+      secret: osp-secret
+  mariadb:
+    template:
+      storageRequest: 500M
```
abays, line 13: The `core_v1beta1_openstackcontrolplane_collapsed_cell.yaml` sample does not have a `template` for `mariadb`, so this will not match.

tests/kuttl/tests/basic-deployment/02-cleanup.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+delete:
+- apiVersion: core.openstack.org/v1beta1
+  kind: OpenStackControlPlane
+  name: openstack
+  namespace: $NAMESPACE
```
abays, line 7: I had to remove all of these `namespace: $NAMESPACE` lines in all the `02-cleanup.yaml` and `02-errors-cleanup.yaml` files in both test dirs, otherwise these steps did not actually clean-up the `OpenStackControlPlane`.

tests/kuttl/tests/collapsed/02-cleanup.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+delete:
+- apiVersion: core.openstack.org/v1beta1
+  kind: OpenStackControlPlane
+  name: openstack
```
abays, line 6: `s/name: openstack/name: openstack-collapsed-cell`

tests/kuttl/tests/collapsed/02-errors-cleanup.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: core.openstack.org/v1beta1
+kind: OpenStackControlPlane
+metadata:
+  name: openstack
```
abays, line 4: `s/name: openstack/name: openstack-collapsed-cell`

abays (approved):
/lgtm

PR #330, opened 2023-06-01
Config samples for galera deployment with network isolation

dciabrin:
Sync the galera config samples to add support for network isolation with 1-node galera clusters.


abays (approved):
/lgtm

PR #340, opened 2023-06-07
Don't *merge* OS operator bundles, only ENVs

dprince:
This PR updates the openstack-operator bundle build so that it only merges ENV vars up from the operators, not the bundles themselves.

The "storage bundle" has been deleted. Storage operators are now required just like everything else.

Additionally, the catalog-build and github workflows have been updated to include "pinned" bundle images based on the contents of the go.mod file


hack/pinned-bundle-images.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+set -e
```
dprince, line 2: I would like to fashion a way to share the logic here with pin-custom-bundle-dockerfile.sh as it is largely the same. Perhaps can do that in a follow up.

.github/create_bundle.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+a!/bin/bash
```
abays, line 1: This looks like a typo

abays:
https://github.com/openshift/release/pull/40184 merged

PR #347, opened 2023-06-14
Delete galera pvcs in kuttl test

cescgina:
A kuttl test case is deploying openstack with galera but not deleting
its pvcs. That causes the CI job to hang when it tries to delete the
create pvs. This change explicitely deletes the 3 galera pvcs.

### DIFF
--- a/tests/kuttl/tests/galera-3replicas/02-cleanup.yaml
+++ b/tests/kuttl/tests/galera-3replicas/02-cleanup.yaml
@@ -4,3 +4,9 @@ delete:
 - apiVersion: core.openstack.org/v1beta1
   kind: OpenStackControlPlane
   name: openstack
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- script: |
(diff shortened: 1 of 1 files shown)


tests/kuttl/tests/galera-3replicas/02-cleanup.yaml
```diff
@@ -4,3 +4,9 @@ delete:
 - apiVersion: core.openstack.org/v1beta1
   kind: OpenStackControlPlane
   name: openstack
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- script: |
+    oc delete -n $NAMESPACE pvc mysql-db-openstack-galera-0 mysql-db-openstack-galera-1 mysql-db-openstack-galera-2
```
abays, line 12: I know it's probably impossible for this situation to occur, but what it makes sense to include `--ignore-not-found` just in case these are somehow missing?

cescgina: It's unlikely, but it won't hurt, so I added just in case

abays (approved):
/lgtm

PR #349, opened 2023-06-16
Donot trigger dependent jobs on irrelevant files changes

raukadah:
This patch adds the irrelevant files list to edpm and baremetal jobs to avoid the job freeze issue.


raukadah:
recheck

raukadah:
recheck

PR #353, opened 2023-06-20
Update Heat CR definition in samples

kajinamit:
Some spec keys are now hidden since the recent refactoring. This updates the definitions and leaves only minimum set.

### DIFF
--- a/config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml
+++ b/config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml
@@ -163,48 +163,12 @@ spec:
   heat:
     enabled: false
     template:
-      customServiceConfig: "# add your customization here"
       databaseInstance: openstack
-      databaseUser: "heat"
-      rabbitMqClusterName: rabbitmq
-      debug:
-        bootstrap: false
-        dbSync: false
-        service: false
       heatAPI:
-        customServiceConfig: "# add your customization here"
-        databaseUser: "heat"
-        debug:
-          bootstrap: false
-          dbSync: false
-          service: false
-        passwordSelectors:
-          service: AdminPassword
-          database: AdminPassword
         replicas: 1
-        resources: {}
-        secret: "osp-secret"
-        serviceUser: ""
       heatEngine:
-        customServiceConfig: "# add your customization here"
-        databaseUser: "heat"
-        debug:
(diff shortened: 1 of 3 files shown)


bshephar (dismissed):

config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml
```diff
@@ -163,48 +163,14 @@ spec:
...
-          database: AdminPassword
         replicas: 1
-        resources: {}
-        secret: "osp-secret"
-        serviceUser: ""
-      passwordSelectors:
-        service: AdminPassword
-        database: AdminPassword
-      preserveJobs: false
+      rabbitMqClusterName: rabbitmq
```
bshephar, line 172: I think this is the one pre-commit is complaining about.

bshephar: Yeah, easier to see the indentation issue in this view:
https://github.com/openstack-k8s-operators/openstack-operator/blob/aa26499c3d330c898c90b0e3223ccb8eb0d90c49/config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml#LL163C1-L173C25

config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml
```diff
@@ -208,48 +208,14 @@ spec:
...
-          database: AdminPassword
         replicas: 1
-        resources: {}
-        secret: "osp-secret"
-        serviceUser: ""
-      passwordSelectors:
-        service: AdminPassword
-        database: AdminPassword
-      preserveJobs: false
+      rabbitMqClusterName: rabbitmq
```
bshephar, line 217: And the same for this one:
https://github.com/openstack-k8s-operators/openstack-operator/blob/aa26499c3d330c898c90b0e3223ccb8eb0d90c49/config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml#L212-L217

kajinamit: Removed these. I initially intended to keep these but we don't have this key in the other CRs so we can probably remove it now.

PR #354, opened 2023-06-21
Enable memcached by default

kajinamit:
... because memcached is globally used as caching backend.

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_types.go
+++ b/apis/core/v1beta1/openstackcontrolplane_types.go
@@ -108,6 +108,7 @@ type OpenStackControlPlaneSpec struct {
 	Rabbitmq RabbitmqSection `json:"rabbitmq,omitempty"`
 
 	// +kubebuilder:validation:Optional
+	//+operator-sdk:csv:customresourcedefinitions:type=spec
 	// Memcached - Parameters related to the Memcached service
 	Memcached MemcachedSection `json:"memcached,omitempty"`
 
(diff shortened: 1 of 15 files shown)


apis/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -262,11 +262,13 @@ type RabbitmqSection struct {
 // MemcachedSection defines the desired state of Memcached services
 type MemcachedSection struct {
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default=false
+	// +kubebuilder:default=true
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
```
abays, line 266: I'm curious as to what this annotation is doing.  I haven't seen it before.  Is it something add by a newer version of `operator-sdk`?

kajinamit: IIUC these are code markers supported by operator-sdk. While the existing annotations work for crd, these work for csv. https://v0-19-x.sdk.operatorframework.io/docs/golang/legacy/references/markers/

These were previously added to the other part of specs so I'm adding the same annotations for consistency. I noticed I didn't regenerate csv properly after making the change, which is not fixed in the latest change.

If we want I can make it a separate change.

abays: Cool, I was just wondering.  I think it's fine to have it in this PR.

kajinamit:
recheck

abays:
@kajinamit Is this how OOO behaves as well?  You have to have Memcached to be able to use certain OpenStack services?

kajinamit:
> @kajinamit Is this how OOO behaves as well? You have to have Memcached to be able to use certain OpenStack services?

Yes. For a recent few releases all openstack services are deployed with caching enabled, which is backed by memcached.

abays (approved):
/lgtm

PR #364, opened 2023-06-26
Wire in webhook validations from underlying operators

kajinamit:
... to force these validations at meta operator layer as well.

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_webhook.go
+++ b/apis/core/v1beta1/openstackcontrolplane_webhook.go
@@ -61,7 +61,7 @@ func (r *OpenStackControlPlane) ValidateCreate() error {
 
 	var allErrs field.ErrorList
 	basePath := field.NewPath("spec")
-	if err := r.ValidateServices(basePath); err != nil {
+	if err := r.ValidateServices(basePath, false); err != nil {
 		allErrs = append(allErrs, err...)
 	}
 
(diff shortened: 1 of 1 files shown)


apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -166,56 +166,129 @@ func (r *OpenStackControlPlane) ValidateServices(basePath *field.Path) field.Err
 			err := field.Invalid(basePath.Child("keystone").Child("enabled"), r.Spec.Keystone.Enabled, depErrorMsg)
 			allErrs = append(allErrs, err)
 		}
+
+		if isUpdate {
+			if err := r.Spec.Keystone.Template.ValidateUpdate(basePath.Child("keystone").Child("template")); err != nil {
```
gibizer, line 171: This signature is problematic. The original ValidateUpdate signature from the webhook has the signature of  `ValidateUpdate(old runtime.Object)` so webhooks can compare old and new values during update validation. So we need to pass such old values from openstack-operator too to keep the same possibility

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -150,7 +150,7 @@ func (r *OpenStackControlPlane) checkDepsEnabled(name string) string {
 }
 
 // ValidateServices implements common function for validating services
-func (r *OpenStackControlPlane) ValidateServices(basePath *field.Path) field.ErrorList {
+func (r *OpenStackControlPlane) ValidateServices(basePath *field.Path, isUpdate bool) field.ErrorList {
```
gibizer, line 153: Booleans in function parameters that then used in ifs to change logic is a common code smell. This function probably need to be two separate functions, one for the create and one for the update codepath.

gibizer (changes requested):
I would really do this step by step so that when an service operator needs validation it can add it and hook it in.

PR #370, opened 2023-06-27
Ensure Memcached is deleted after being removed from spec

kajinamit:
OpenStackControlPlane CR allows defining multiple Memcached instances, and users can remove existing Memcached instances from sppec.

This change ensures the instances are deleted when they are removed from spec.


kajinamit:
We probably should add kuttl test for this scenario.

abays (approved):
/lgtm

abays:
recheck

PR #371, opened 2023-06-27
Don't use filter_tag_name with quay.rdoproject.org

rabi:
It does not work with some non quay.io registries. Also we push the built images to rdo to query them, so they would be fresh and be in the limites number of items returned with list.


raukadah:
Zuul content provider job is failing with following [error](https://logserver.rdoproject.org/71/371/18e500ed0c9956dc792d218583b30542152fea86/github-check/openstack-operator-content-provider/4071a69/ci-framework-data/logs/ci_make_006_openstack-operator_-_call_catalog-build.log)
```
Error: initializing source docker://quay.io/openstack-k8s-operators/rabbitmq-cluster-operator-bundle@sha256:49044ebeb92a5bfc3b597b3042d057b63f7c1cddfc5555f2d49780afe228ece0: reading manifest sha256:49044ebeb92a5bfc3b597b3042d057b63f7c1cddfc5555f2d49780afe228ece0 in quay.io/openstack-k8s-operators/rabbitmq-cluster-operator-bundle: manifest unknown
```

hack/pin-bundle-images.sh
```diff
@@ -46,6 +46,8 @@ for MOD_PATH in $(go list -mod=readonly -m -json all | jq -r '. | select(.Path |
 
     if [[ ${LOCAL_REGISTRY} -eq 1 && ( "$GITHUB_USER" != "openstack-k8s-operators" || "$BASE" == "$IMAGEBASE" ) ]]; then
         SHA=$(curl -s ${REPO_CURL_URL}/$BASE-operator-bundle/tags/list | jq -r .tags[] | sort -u | grep $REF)
+    elif [[ "${IMAGEREGISTRY}" != "quay.io" ]]; then
+        SHA=$(curl -s ${REPO_CURL_URL}/$BASE-operator-bundle/tag/ | jq -r .tags[].name | sort -u | grep $REF)
```
steveb, line 50: We should land this if it unblocks CI, but something like the following could be an alternative to filter_tag_name, and avoid using sort and grep:

  jq -r '.tags | map(select ( .name | contains("$REF")))[0].name'

rabi: I think that returns a "null" string and won't pass with [-z $SHA] check below, would leave it as is and can be worked on later, if required.

[ramishra@fedora ~]$ curl -s 'https://quay.io/api/v1/repository/openstack-k8s-operators/dataplane-operator-bundle/tag/' |  jq -r '.tags | map(select ( .name | contains("123456")))[0].name'
null

[ramishra@fedora ~]$ curl -s 'https://quay.io/api/v1/repository/openstack-k8s-operators/dataplane-operator-bundle/tag/' |  jq -r '.tags[].name' | sort -u | grep 123456
[ramishra@fedora ~]$

rabi:
Has to wait for https://github.com/openstack-k8s-operators/openstack-operator/pull/372

gibizer:
#372 landed

Sandeepyadav93:
recheck

stuggi (approved):
/lgtm

rabi:
recheck

karelyatin (approved):
/lgtm

abays:
recheck

PR #373, opened 2023-06-27
Reconcile MariaDB/Galera when disabled

kajinamit:
... to ensure the existing instances are purged. Note that this change does not remove the instances already removed from spec.


kajinamit:
recheck

PR #374, opened 2023-06-27
Make sure the requested database instance exists

kajinamit:
This makes sure the database instances requested by individual services are defined as part of the control plane. This helps users to avoid unexpectedly editing name of the database instances listed.

Note that this does not cover nova which requires multiple database instances.


kajinamit:
I probably should wait until https://github.com/openstack-k8s-operators/openstack-operator/pull/375 is merged.

abays:
Would this interfere with migrations in any way?

kajinamit:
> Would this interfere with migrations in any way?

No. This does not block the migration though this introduces an requirement to deploy 2 database instances during migration.

PR #379, opened 2023-06-28
Use default storage class for galera resource

dciabrin:
Rely on webhooks to populate a default storage class if not provided in the CR. This is useful in e.g. CI to provide a custom storage class during runs.


abays:
Just for the sake of clarity, it technically isn't the webhooks that do defaulting for `Galera.Spec.StorageClass` at the moment.  It's handled within the reconcile loop itself [1].

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/8a87bb3a4d557858f4ede09bdc1c29e907a3a2ba/pkg/openstack/galera.go#L111-L113

abays (approved):
/lgtm

PR #380, opened 2023-06-28
Fix cinderScheduler in config samples

ASBishop:
Fix the cinderScheduler configuration to ensure an actual replica is created. Even though its default replica count is 1, the service isn't properly initialized unless _something_ is specified in the CR.

### DIFF
--- a/config/samples/core_v1beta1_openstackcontrolplane_galera_network_isolation.yaml
+++ b/config/samples/core_v1beta1_openstackcontrolplane_galera_network_isolation.yaml
@@ -26,6 +26,8 @@ spec:
           ipAddressPool: internalapi
           loadBalancerIPs:
           - 172.17.0.80
+      cinderScheduler:
+        replicas: 1
       cinderBackup:
         networkAttachments:
         - storage
--- a/config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml
+++ b/config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml
@@ -26,6 +26,8 @@ spec:
           ipAddressPool: internalapi
           loadBalancerIPs:
           - 172.17.0.80
+      cinderScheduler:
+        replicas: 1
       cinderBackup:
         networkAttachments:
         - storage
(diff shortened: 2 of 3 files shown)


config/samples/core_v1beta1_openstackcontrolplane_galera_network_isolation.yaml
```diff
@@ -26,6 +26,8 @@ spec:
           ipAddressPool: internalapi
           loadBalancerIPs:
           - 172.17.0.80
+      cinderScheduler:
+        replicas: 1
```
abays, line 30: I'm perplexed as to how these samples could have even worked without an explicit `cinderScheduler` block.  According to `CinderSpec` struct, that field is required [1].  Maybe the defaulting webhook injects a zero-value and thus bypasses the constraint.  This is certainly something we've been concerned about elsewhere.

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/1a796805ec77fabb597eab76ff2f70723dc8183b/api/v1beta1/cinder_types.go#L100-L102

abays: So I realized something while I was thinking about this some more.  Our webhooks will inject the default images for all Cinder services [1], which will result in the various `Cinder*` blocks appearing for those services in the `CinderSpec`.  That's how it gets around the `required` constraint.  But what the webhooks do not do is set any of the other fields in the blocks (i.e. `replicas`).  Those would either be absent, have zero-values or perhaps have some `kubebuilder` default.

I am surprised though that this was ever accepted as valid YAML by the API:

```
  cinderScheduler:
  cinderBackup:
    replicas: 0
```

I would think `cinderScheduler:` would have to be `cinderScheduler: {}` to be parsable.

abays (approved):
/lgtm

config/samples/core_v1beta1_openstackcontrolplane_galera_network_isolation.yaml
ASBishop: fwiw, I recall seeing the behavior we're discussing prior to when we reorganized the structs (the nesting).

abays:
recheck

abays:
recheck

PR #383, opened 2023-06-30
Align to the validation webhook pattern

gibizer:
* split up the validation calls towards the service CRDs to have a separate ValidateCreate and ValidateUpdate call
* make sure that the service operators ValidateUpdate call gets the old version of the Spec struct so that they can compare

The pattern is documented in https://github.com/openstack-k8s-operators/docs/pull/47

Depends-on: https://github.com/openstack-k8s-operators/ironic-operator/pull/277 (merged)
Depends-on: https://github.com/openstack-k8s-operators/nova-operator/pull/428 (merged)


gibizer:
recheck
```
FAILED - RETRYING: [localhost]: Wait for OpenStack subscription creation (1 retries left).
```

PR #388, opened 2023-07-03
Cleanup pvcs and add status checks

jparoly:


tests/kuttl/tests/basic-deployment/02-cleanup.yaml
```diff
@@ -4,3 +4,8 @@ delete:
 - apiVersion: core.openstack.org/v1beta1
   kind: OpenStackControlPlane
   name: openstack
+commands:
+- script: |
+    oc delete --ignore-not-found=true -n $NAMESPACE pvc \
+    ansible-ee-logs \
```
jparoly, line 10: not sure if I need to delete ansible-ee-logs; it was left over one time locally

cescgina: I think that the `ansible-ee-logs` is created by `make crc_storage` so we don't need to delete it here, but I'm not completely sure about that

jparoly: removed it

tests/kuttl/common/errors_cleanup_openstack.yaml
abays, line 1: This file needs a newline at the end to pass `pre-commit`

tests/kuttl/tests/collapsed/01-assert-collapsed-cell.yaml
```diff
@@ -111,3 +104,57 @@ spec:
         service: CeilometerPassword
       secret: osp-secret
       serviceUser: ceilometer
+status:
```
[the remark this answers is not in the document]
jparoly: I'm seeing a bunch of other errors too.  Are the 3 tests running in parallel in CI?  That would be a problem I think.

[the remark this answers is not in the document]
cescgina: They should only run one at a time: https://github.com/openstack-k8s-operators/openstack-operator/blob/08801ea0fd19cfeebc2c6d6f9d520cc2e35397b6/kuttl-test.yaml#L23

config/samples/core_v1beta1_openstackcontrolplane.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: core.openstack.org/v1beta1
 kind: OpenStackControlPlane
 metadata:
-  name: openstack
+  name: openstack-basic
```
cescgina, line 4: One of the test failures is because there is a mismatch between the name of the `OpenstackControlPlane` here and in the assert file: https://github.com/openstack-k8s-operators/openstack-operator/blob/08801ea0fd19cfeebc2c6d6f9d520cc2e35397b6/tests/kuttl/common/assert-sample-deployment.yaml#L4

tests/kuttl/tests/collapsed/01-assert-collapsed-cell.yaml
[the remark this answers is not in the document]
cescgina: The problem is that because the test assertion fails, cleanup is not run, and then the following test cases can't deal with resources from previous deployment (rabbitmq seems particularly problematic)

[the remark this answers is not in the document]
jparoly: > Based on the results of the KUTTL test in CI, it looks like this list is out-of-date:
> 
> ```
> case.go:366: resource OpenStackControlPlane:openstack-kuttl-tests/openstack-collapsed-cell: .status.conditions: slice length mismatch: 13 != 18
> ```

I ran all tests locally manually, separated with `make crc_storage_cleanup` and `make crc_storage` and they all passed.

jparoly:
I do not understand the github-check failures, is this a known issue/flakiness we have seen?

jparoly:
Rabbitmq and placement seems to have problems but they also do not seem to be running in order or waiting for the last test to finish.

jparoly:
> Rabbitmq and placement seems to have problems but they also do not seem to be running in order or waiting for the last test to finish.

actually it seems it tries to run galera-3replicas before collapsed test... hmmm not sure why, adding a sleep

jparoly:
kuttl tests will fail until https://github.com/openstack-k8s-operators/install_yamls/pull/420 is merged

abays:
recheck

abays (approved):
/lgtm

PR #389, opened 2023-07-04
Update to last telemetry-operator version to workaround the disappare…

jlarriba:
…ance of playbooks in edpm-ansible


jlarriba:
recheck

jlarriba:
recheck

PR #398, opened 2023-07-19
Revert "Remove OpenStackClient CR"

jistr:
This reverts commit ccd18fcc69d3151af229df071faa825253bb264e, which caused a promotion pipeline failure. We can reintroduce it when we're ready to land the patch that reintroduces it in keystone operator.


kajinamit:
No we can't revert this. This is causing circular dependencies between keystone-operator and infra-operator and is blocking our release process. The plan was to move this to keystone-operator so that we have only the straight dependency.

PR #400, opened 2023-07-19
Configure renovate to pin k8s and controller-runtime for dev preview and pin ocp

stuggi:
- pin k8s and controller runtime to be in sync with lib-common and configure renovate to keep the pin
- pin ocp api version to a hash in sync with lib-common instead of the nonexistent v3.9.0+incompatible tag


stuggi:
recheck

PR #430, opened 2023-08-10
Sync with latest rabbitmq structs

dprince:
go get github.com/rabbitmq/cluster-operator/v2@v2.5.0
make manifests generate

Also updated go.mod to pin us to k8s.io 0.26.x for now as rabbitmq bumped us to 0.27


dasm:
recheck

dprince:
recheck

abays (approved):
/lgtm

apis/go.mod
```diff
@@ -97,6 +97,15 @@ require (
...
 
+replace (
+	// pin to k8s 0.26.x for now
+	k8s.io/api => k8s.io/api v0.26.9
+	k8s.io/apimachinery => k8s.io/apimachinery v0.26.9
+	k8s.io/client-go => k8s.io/client-go v0.26.9
+	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.14.6
+
+)
+
```
gibizer, line 108: So this forces go build to use 0.26.9 even if rabbitmqv2 declares that is needs 0.27. As we only use the struct from the rabbitmqv2 and does not call any function from the rabbitmqv2 module this is probably OK. If we would call some defaulting or validation functions from rabbitmqv2 then this can be problematic given that 0.27 forces signature changes in that area.

dprince: Eventually we'll bump up to 0.27 too so this should work out in the long term. The short term alternative is to stay on the old rabbitmq structs.

This has been posted for over a month... @gibizer do you really want to hold off on this further?

gibizer: Sorry I wasn't clear. I think this is OK as is, as we dont try to call any defaulting or validationg functions feom rabbitv2. So I had no intentio  to gold this up.

gibizer (approved):

PR #436, opened 2023-08-23
Use lib-common functionality to calculate ReadyCondition

kajinamit:
Recent lib-common provides the way to calculate condition based on sub conditions. This replaces the current implementation by that common one.


fao89 (approved):

abays (approved):
/lgtm

PR #440, opened 2023-08-25
Default service specs even if the service is disabled

gibizer:
Even if the service is disabled the service spec needs to be valid according to then schema so that it can be stored in etcd.

If there is a field like NovaSpec.NovaMetadataTemplate.Enabled that is defined as *bool but the default value of it is only provided by the NovaSpec defaulting webhook (as the default value is context dependent) then if the defaulting is not run on the NovaSpec then the field value remains null, and that fails the schema validation.

So this patch make the defaulting unconditional for every service spec.

Needed-by: https://github.com/openstack-k8s-operators/nova-operator/pull/460


apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -158,7 +158,7 @@ func (r *OpenStackControlPlane) checkDepsEnabled(name string) string {
 	case "Octavia":
 		if !((r.Spec.Mariadb.Enabled || r.Spec.Galera.Enabled) && r.Spec.Memcached.Enabled && r.Spec.Rabbitmq.Enabled &&
 			r.Spec.Keystone.Enabled && r.Spec.Neutron.Enabled && r.Spec.Glance.Enabled && r.Spec.Nova.Enabled &&
-		    r.Spec.Ovn.Enabled) {
+			r.Spec.Ovn.Enabled) {
```
gibizer, line 161: this was done by go vet automatically for me

dprince (approved):

PR #441, opened 2023-08-25
update the nova-operator and dataplane-operator tags

SeanMooney:
This chagne just updates teh nova and dataplane operator
api module tags to take account of the v0.1.1 tags.
While there are no CRD changes in the or other changes in the API
moduels its still good to keep these in sync but will not
break anything in this instance.


SeanMooney:
recheck

dprince (approved):

dprince:
I think we've already picked these up elsewhere. Please re-open if still needed

PR #451, opened 2023-08-29
Bind manila-share to the storage network

fmount:
The sample had a wrong networkAttachment in both scheduler and manilaShare. This patch is supposed to fix the NAD part of the template specs.

### DIFF
--- a/config/samples/core_v1beta1_openstackcontrolplane_galera_network_isolation.yaml
+++ b/config/samples/core_v1beta1_openstackcontrolplane_galera_network_isolation.yaml
@@ -127,13 +127,11 @@ spec:
         - internalapi
       manilaScheduler:
         replicas: 1
-        networkAttachments:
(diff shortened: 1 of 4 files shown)


fmount:
recheck

fmount:
I think the CI failure is out of scope here, @abays @dprince given this is just a sample file (with the right network fix), we could merge it, it should be safe enough, right?

abays:
recheck

@fmount We could probably just merge it, but unless it's urgent let's try one more time

fmount:
> recheck
> 
> @fmount We could probably just merge it, but unless it's urgent let's try one more time

Not urgent, I'm ok to try even more times :D

fmount:
> recheck
> 
> @fmount We could probably just merge it, but unless it's urgent let's try one more time

@abays safe to merge I guess at this point

PR #457, opened 2023-09-07
Create several services routes based on annotations on the svc endpoint

stuggi:
Creates the route for the keystoneapi, glance, placement, cinder, neutron, nova, heat, horizon, manila and swift based on the annotation set by the service operator on the their service.

The service operator sets for public endpoint the following ingress_create and ingress_name annotation on the service:
```
apiVersion: v1
kind: Service
metadata:
  annotations:
    core.openstack.org/ingress_create: "true"
    core.openstack.org/ingress_name: glance
```
* Based on this information and the label selector the openstack-operator passes into the service override on the component, the openstack-operator creates their routes and upsdates the component CR service override with the route hostname for the endpointURL
* with this the openstack-operator does not have to know the service name schema on the service operators
* The `core.openstack.org/ingress_name` annotation also allows the customization of the route prefix via the service override for the component. The default applied by the service operators is the service name specified in the service operator, e.g to get a route for cinder api  `my-cinder-openstack.apps-crc.testing`:
```
      cinderAPI:
        override:
          service:
            public:
              metadata:
                annotations:
                  core.openstack.org/ingress_name: my-cinder
```
* using the service overrides also allows to use a LoadBalancer for the pubic endpoint (cloud even change between route/LoadBalancer service), e.g.
```
      glanceAPIExternal:
        override:
          service:
            metadata:
              annotations:
                metallb.universe.tf/address-pool: internalapi
                metallb.universe.tf/allow-shared-ip: internalapi
                metallb.universe.tf/loadBalancerIPs: 172.17.0.90
            spec:
              type: LoadBalancer
```
* Also allows to customize the route via override, so that timeouts or tls settings can be applied/customized.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/332
Depends-On: https://github.com/openstack-k8s-operators/keystone-operator/pull/307
Depends-On: https://github.com/openstack-k8s-operators/glance-operator/pull/309
Depends-On: https://github.com/openstack-k8s-operators/placement-operator/pull/61
Depends-On: https://github.com/openstack-k8s-operators/cinder-operator/pull/262
Depends-On: https://github.com/openstack-k8s-operators/neutron-operator/pull/204
Depends-On: https://github.com/openstack-k8s-operators/nova-operator/pull/522
Depends-On: https://github.com/openstack-k8s-operators/heat-operator/pull/238
Depends-On: https://github.com/openstack-k8s-operators/horizon-operator/pull/212
Depends-On: https://github.com/openstack-k8s-operators/manila-operator/pull/130
Depends-On: https://github.com/openstack-k8s-operators/swift-operator/pull/49
Depends-On: https://github.com/openstack-k8s-operators/infra-operator/pull/119

Jira: [OSP-26690](https://issues.redhat.com//browse/OSP-26690)


pkg/openstack/placement.go
```diff
@@ -32,9 +38,63 @@ func ReconcilePlacementAPI(ctx context.Context, instance *corev1beta1.OpenStackC
...
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		var ctrlResult reconcile.Result
+		instance.Spec.Placement.Template.Override.Service, ctrlResult, err = EnsureRoute(
+			ctx,
+			instance,
+			helper,
+			placementAPI,
```
gibizer, line 81: I'm wondering if we need that the service CR owns the Route. I don't really see a technical issue with this now just feels like we are reaching over the head of the service operator and forcing it to "adopt" a route. What are the pros of having the owner reference pointing to the service CR instead of simply letting the OpenstackControlPlane CR to own the Route it creates.

pkg/openstack/common.go
```diff
@@ -27,3 +41,255 @@ func EnsureDeleted(ctx context.Context, helper *helper.Helper, obj client.Object
...
+						UID:                instance.GetUID(),
+						BlockOwnerDeletion: pointer.Bool(true),
+						Controller:         pointer.Bool(true),
+					}
+
+					owner := metav1.GetControllerOf(&r.ObjectMeta)
+
+					if r.Name != rd.RouteName && owner != nil && referSameObject(*owner, instanceRef) {
+						// Delete any other owner refs from ref list to not block deletion until owners are gone
+						r.SetOwnerReferences([]metav1.OwnerReference{instanceRef})
```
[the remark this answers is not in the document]
stuggi: we add the service operator CR to the owner ref list. When the osctlplane CR gets deleted, this prevents that the route gets deleted immediately before the osp service resources are deleted. This is required that keystoneendpoints can be cleaned up properly without adding cleanup logic in the openstack-operator. In case of someone overrides the ingress_name after initial create, a new route gets created. In that case we override the refs to just the ctlplane CR ref.

pkg/openstack/common.go
```diff
@@ -27,3 +41,255 @@ func EnsureDeleted(ctx context.Context, helper *helper.Helper, obj client.Object
...
+			if rd.endpointURL != "" {
+				// Any trailing path will be added on the service-operator level.
+				svcOverride.EndpointURL = &rd.endpointURL
+				instance.Status.Conditions.MarkTrue(condType, condition.ExposeServiceReadyMessage)
+			}
+		}
+
+		// make sure the route is deleted and endpointURL cleaned
+		if svc.ObjectMeta.Annotations[service.AnnotationIngressCreateKey] == "false" {
+			err := rd.DeleteRoute(ctx, helper)
```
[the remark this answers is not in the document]
stuggi: could also add a check to delete it only if the ctlplane CR are is the controller.

pkg/openstack/common.go
```diff
@@ -27,3 +41,255 @@ func EnsureDeleted(ctx context.Context, helper *helper.Helper, obj client.Object
...
+			if err != nil {
+				return svcOverrides, ctrl.Result{}, err
+			}
+
+			instance.Status.Conditions.Remove(condType)
+
+			//svcOverride.EndpointURL = nil
+			//helper.GetLogger().Info(fmt.Sprintf("Service %s override endpointURL removed", svc.Name))
```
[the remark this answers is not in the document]
stuggi: yeah, thats something I still need to test/add

pkg/openstack/common.go
```diff
@@ -27,3 +41,255 @@ func EnsureDeleted(ctx context.Context, helper *helper.Helper, obj client.Object
...
+
+				return svcOverrides, ctrlResult, err
+			} else if (ctrlResult != ctrl.Result{}) {
+				return svcOverrides, ctrlResult, nil
+			}
+
+			// update override for the service with the route endpoint url
+			if rd.endpointURL != "" {
+				// Any trailing path will be added on the service-operator level.
+				svcOverride.EndpointURL = &rd.endpointURL
```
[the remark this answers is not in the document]
stuggi: yes

pkg/openstack/common.go
```diff
@@ -27,3 +41,255 @@ func EnsureDeleted(ctx context.Context, helper *helper.Helper, obj client.Object
...
+						Kind:               instance.Kind,
+						Name:               instance.GetName(),
+						UID:                instance.GetUID(),
+						BlockOwnerDeletion: pointer.Bool(true),
+						Controller:         pointer.Bool(true),
+					}
+
+					owner := metav1.GetControllerOf(&r.ObjectMeta)
+
+					if r.Name != rd.RouteName && owner != nil && referSameObject(*owner, instanceRef) {
```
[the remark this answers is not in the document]
stuggi: thanks, will check

pkg/openstack/placement.go
stuggi: the ctlplane CR still owns it as the controller, as mentioned above we just add the service CR to the ownerref list so that we block early deletion and make sure it stays until the service cr is gone

pkg/openstack/common.go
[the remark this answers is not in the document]
dprince: Correct me if I'm wrong but it appears the nova function relies on having ownership setup correctly. The referSameObject function instead just checks the group,type, name and is perhaps more robust that way. My take unless I missed something here. I guess referSameObject was just copied from controllerutil or something? Too bad we couldn't reuse that (private) function from there.

pkg/openstack/nova.go
```diff
@@ -48,6 +54,106 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
...
+		// cell NoVNCProxy service override
+		for cellName := range instance.Spec.Nova.Template.CellTemplates {
+			cellTemplate := instance.Spec.Nova.Template.CellTemplates[cellName]
+			if cellTemplate.NoVNCProxyServiceTemplate.Override.Service == nil {
+				cellTemplate.NoVNCProxyServiceTemplate.Override.Service = map[string]service.OverrideSpec{}
+			}
+			cellTemplate.NoVNCProxyServiceTemplate.Override.Service[string(endpointType)] =
+				AddServiceComponentLabel(
+					ptr.To(cellTemplate.NoVNCProxyServiceTemplate.Override.Service[string(endpointType)]),
+					nova.Name+"-"+cellName+"-novnc")
```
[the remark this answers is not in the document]
dprince: functions are fine but documenting the convention and using it (like we do here) is also fine I think. OpenStack operator has access to the nova data as is so it should be able to align here

dprince:
I really like how this consolidates Route creation logic. One concern I have is if we were to split out certain services from the OpenStackControlplane (making it more of a *core* or *enabledByDefault* set of controlplane services). Like if we were to make the AdvancedNetworking services like Octavia or Designate top level or something like that to make the OpenStackControlplane smaller. If we did that how would that impact the pattern we use here?

pkg/openstack/common.go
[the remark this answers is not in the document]
stuggi: yes, I remembered having seen that there, but unfortunately its not a public one, thats why I copied it.

stuggi:
> I really like how this consolidates Route creation logic. One hypothetical concern I can think of is if we were to split out certain services from the OpenStackControlplane (making it more of a _core_ or _enabledByDefault_ set of controlplane services). Like if we were to make the AdvancedNetworking services like Octavia or Designate top level or something like that to make the OpenStackControlplane smaller. If we did that then perhaps we'd make an AdvancedNetworking umbrella CRD in the openstack-operator and could create routes there similarly. Again all this is hypothetical, sort of a what if.
> 
> I'd vote we start landing the parts and pieces here in the dependent PRs

yes, controllers handling additional CRDs with split out/additional services would then just do the same.

pkg/openstack/common.go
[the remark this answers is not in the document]
stuggi: delete is now handled

pkg/openstack/common.go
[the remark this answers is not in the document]
stuggi: its there

pkg/openstack/common.go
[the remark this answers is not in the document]
gibizer: OK I think I see the difference.As there is only maximum one controller per route and we only want to check if the OpenStackControlPlane CR (`instance`) is the controller for this route we don't need to iterate all the owner references here. So you can ignore my suggestion from nova-operator. 

But I have two comments:
* `owner` can potentially be nil if somebody created a route manually with the same selector so we need to a nil check
* I think the conditions in referSameObject is overkill here. We know the UID of the controller of the route object and we know the UID of the OpenstackControPlane CR. As UID is unique for the whole cluster (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids) it is enough to compare owner.GetUID() with instance.GetUID().

pkg/openstack/common.go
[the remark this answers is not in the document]
gibizer: yepp, that is my point. We the ctrlplane CR is not the owner of this Route the we should not delete it.

pkg/openstack/nova.go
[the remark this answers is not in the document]
gibizer: > it is just a label for each cell. it is not important what the content is, it just needs to be uniq per cell that we can query the services and label them. Also use the same label then on the route

Good point. This does not need to match with the name of the NovaNoVNCProxy CR. It just needs to match between the places setting the service override and then querying the service to create a route. Btw, could you pull this string generation out to a function and call it from these two places to keep them in sync?

pkg/openstack/nova.go
```diff
@@ -45,9 +51,110 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
...
+			return ctrlResult, err
+		} else if (ctrlResult != ctrl.Result{}) {
+			return ctrlResult, nil
+		}
+	}
+
+	if nova.Status.Conditions.IsTrue(novav1.NovaAllCellsDBReadyCondition) {
+		// cell NoVNCProxy
+		for cellName := range instance.Spec.Nova.Template.CellTemplates {
+			labelValue := nova.Name + "-" + cellName + "-novnc"
```
gibizer, line 123: This needs to match with L78 but does not need to match anything in the nova-operator

pkg/openstack/placement.go
```diff
@@ -32,9 +38,63 @@ func ReconcilePlacementAPI(ctx context.Context, instance *corev1beta1.OpenStackC
...
 
+	// add selector to service overrides
+	for _, endpointType := range []service.Endpoint{service.EndpointPublic, service.EndpointInternal} {
+		if instance.Spec.Placement.Template.Override.Service == nil {
+			instance.Spec.Placement.Template.Override.Service = map[string]service.OverrideSpec{}
+		}
+		instance.Spec.Placement.Template.Override.Service[string(endpointType)] = AddServiceComponentLabel(
+			ptr.To(instance.Spec.Placement.Template.Override.Service[string(endpointType)]),
+			placementAPI.Name)
+	}
```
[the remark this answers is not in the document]
gibizer: I was wrong. We don't actually want to use the names as label, we only want arbitrary (but unique) labels the openstack-operator can define and query. It is just coincidence that the name of the PlacemnetAPI CR is the same as the label here.

pkg/openstack/placement.go
gibizer: I see. Is there a specific need for the route to outlive the OpenStackControlPlane CR? I guess the PlacementAPI only owner is also the  OpenStackControlPlane CR? So during OpenStackControlPlane deletion both will be deleted anyhow.

gibizer (commented):
As we discussed we don't want to spend effort creating an abstraction over the Route object due to the cost of maintaining such abstraction. I feel that we will pay that cost in a different form, maintaining nova (and potentially other service operator) logic in openstack-operator.

apis/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -388,6 +388,23 @@ type NovaSection struct {
...
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// APIOverride, provides the ability to override the generated manifest of several child resources.
+	APIOverride Override `json:"apiOverride,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// CellOverride, provides the ability to override the generated manifest of several child resources
+	// for a nova cell.
+	CellOverride map[string]NovaCellOverrideSpec `json:"cellOverride,omitempty"`
```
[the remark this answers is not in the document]
stuggi: > ```yaml
>     nova:
>       enabled: true
>       cellOverride:
>         cell1:
>           noVNCProxy:
>             route:
>               spec:
>                 tls
>       template:
>         cellTemplates:
>           cell0:
>           cell1:
>             noVNCProxyServiceTemplate:
>               replicas: 1
> ```
> 
> Form API UX perspective this feels odd as we now have two parallel structures to define cells. The end user need to mentally merge the two parallel structures by key names to get the full picture of the configuration of a cell.
> 
> From implementation perspective openstack-operator needs to learn special rules about nova cells. E.g. that cell0 never have compute nodes and therefore never have running VMs and therefore it does not make sense to have noVNCProxy deployed. Today nova-operator prevents cell0 to have noVNCProxy service deployed. https://github.com/openstack-k8s-operators/nova-operator/blob/8cf10addb78028c2387e0208aeaa9b72d7920226/api/v1beta1/novanovncproxy_webhook.go#L103-L115 So far this information and the related validation is encapsulated in the nova-operator but the parallel cell route configuration here now requires that the openstack-operator also ensure no route is created for noVNCProxy in cell0.

not sure I get that, there is no knowledge required in the openstack-operator on cell specifics. The nova-operator won't create a novnxproxy for cell0 and therefore there is no service for a novncproxy. The openstack-operator therefore won't create a route, or am I missing something here?

pkg/openstack/placement.go
stuggi: the case where this is required is for keystone. when the osctlplane CR gets deleted, it might immediately start cleaning the resources owned by the osctlplane. when the route for keystone is gone, the cleanup of  services/endpoints might fail. Therefore the routes should be cleaned after the service is gone.

gibizer: Ohh. Where do we rely on the public keystone endpoint to clean up resources? That seems like a bug. I would assume that the operators use the internal keystone endpoint to do the cleanup.

apis/core/v1beta1/openstackcontrolplane_types.go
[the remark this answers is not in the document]
gibizer: Ahh you are right. No k8s service means no route creation. So there won't be an unused Route object for cell0. So the following config is always is ignored as cell0 has no NoVNCProxy.
``` yaml
    nova:
      enabled: true
      cellOverride:
        cell0:
          noVNCProxy:
            route:
              spec:
                tls
```
To avoid confusion from the user we at least needs to document it that adding route override for cell0 is ignored. Or we can add a validation webhook check to reject such config. In both case we duplicating pieces of nova-operator. 
[1] https://github.com/openstack-k8s-operators/nova-operator/blob/8cf10addb78028c2387e0208aeaa9b72d7920226/api/v1beta1/novacell_types.go#L75-L80
[2] https://github.com/openstack-k8s-operators/nova-operator/blob/8cf10addb78028c2387e0208aeaa9b72d7920226/api/v1beta1/novanovncproxy_webhook.go#L105-L114

pkg/openstack/common.go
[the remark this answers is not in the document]
stuggi: > OK I think I see the difference.As there is only maximum one controller per route and we only want to check if the OpenStackControlPlane CR (`instance`) is the controller for this route we don't need to iterate all the owner references here. So you can ignore my suggestion from nova-operator.
> 
> But I have two comments:
> 
> * `owner` can potentially be nil if somebody created a route manually with the same selector so we need to a nil check

we have the nil of `owner` as part of the condition

```
// Delete the route if the service was changed not to expose a route
				if svc.ObjectMeta.Annotations[service.AnnotationIngressCreateKey] == "false" &&
					r.Spec.To.Name == svc.Name &&
					owner != nil && referSameObject(*owner, instanceRef) {
```

> * I think the conditions in referSameObject is overkill here. We know the UID of the controller of the route object and we know the UID of the OpenstackControPlane CR. As UID is unique for the whole cluster (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids) it is enough to compare owner.GetUID() with instance.GetUID().

right, we can drop referSameObject and just check the uids.

pkg/openstack/common.go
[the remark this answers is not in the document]
stuggi: its gone now

pkg/openstack/common.go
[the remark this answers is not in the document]
stuggi: done

pkg/openstack/nova.go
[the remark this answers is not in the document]
stuggi: done

apis/core/v1beta1/openstackcontrolplane_types.go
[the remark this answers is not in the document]
stuggi: I have updated the comment in the CRD

pkg/openstack/placement.go
stuggi: the admin service client from [1] is using the public endpoint. its probably ok to switch that to internal

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/main/api/v1beta1/keystoneapi.go#L78

stuggi:
@gibizer I think I have addresses all your comments. switching the endpoint for the admin client used could be done outside. Maybe you could check if I missed something?

pkg/openstack/nova.go
stuggi: this has now a func to better keep it in sync

pkg/openstack/nova.go
```diff
@@ -45,9 +51,110 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
...
+			instance.Spec.Nova.Template.APIServiceTemplate.Override.Service = map[string]service.RoutedOverrideSpec{}
+		}
+		instance.Spec.Nova.Template.APIServiceTemplate.Override.Service[string(endpointType)] =
+			AddServiceComponentLabel(
+				ptr.To(instance.Spec.Nova.Template.APIServiceTemplate.Override.Service[string(endpointType)]),
+				nova.Name+"-api")
+
+		// cell NoVNCProxy service override
+		for cellName := range instance.Spec.Nova.Template.CellTemplates {
+			cellTemplate := instance.Spec.Nova.Template.CellTemplates[cellName]
```
gibizer, line 71: you can simplify this to 
```golang
for cellName, cellTemplate := range ... {}
```

pkg/openstack/nova.go
```diff
@@ -45,9 +51,110 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
...
+			if cellTemplate.NoVNCProxyServiceTemplate.Override.Service == nil {
```
gibizer, line 72: to reduce clutter in the generate OpenstackControlPlane CR you can skip adding override for all the cells where novncproxy is disabled.
```golang
			if cellTemplate.NoVNCProxyServiceTemplate.Enabled == False {
                continue
            }
```

pkg/openstack/nova.go
```diff
@@ -45,9 +51,110 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
...
+			corev1beta1.OpenStackControlPlaneExposeNovaReadyCondition,
+		)
+		if err != nil {
+			return ctrlResult, err
+		} else if (ctrlResult != ctrl.Result{}) {
+			return ctrlResult, nil
+		}
+	}
+
+	if nova.Status.Conditions.IsTrue(novav1.NovaAllCellsDBReadyCondition) {
```
gibizer, line 120: NovaAllCellsDBReadyCondition will be true before NovaNoVNCProxy service is created. 

The NovaAllCellReady would be the ideal target. But if the NovaCell wants to avoid updating the compute config after the NoVNCProyx EndpointURL is set[1], then NovaAllCellReady will not be true until we have the route.

So we have multiple options:
* you can look up the NovaNoVNCProxy CR and look at its ExposeServiceReadyCondition (or whathever we rename that to) status.
* the Nova CR grows a new status condition that shows when all NoVNCProxy service in all cells is ready

[1] https://github.com/openstack-k8s-operators/nova-operator/pull/522#discussion_r1324441019

pkg/openstack/nova.go
```diff
@@ -45,9 +51,110 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
...
+		// cell NoVNCProxy
+		for cellName := range instance.Spec.Nova.Template.CellTemplates {
+			labelValue := getNoVNCProxyServiceLabel(nova.Name, cellName)
+			cellTemplate := instance.Spec.Nova.Template.CellTemplates[cellName]
```
gibizer, line 124: you can simplify:
```golang
for cellName, cellTemplate := range ...{}
```

pkg/openstack/nova.go
```diff
@@ -45,9 +51,110 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
...
+			var ctrlResult reconcile.Result
+			cellTemplate.NoVNCProxyServiceTemplate.Override.Service, ctrlResult, err = EnsureRoute(
+				ctx,
+				instance,
+				helper,
+				nova,
+				svcs,
+				cellTemplate.NoVNCProxyServiceTemplate.Override.Service,
+				instance.Spec.Nova.CellOverride[cellName].NoVNCProxy.Route,
+				corev1beta1.OpenStackControlPlaneExposeNovaReadyCondition,
```
gibizer, line 145: This is problematic. There are multiple EnsureRoute calls for a single reconcile and each of that call sets the same condition. This means that if there are multiple errors then the earlier errors are lost. Also if there is an error then a success then the success will override the error. E.g. if the route creation of the NovaAPI fails but the the route creation of the NoVNCProxy succeeds the later will set the condition to True and we loose the error.

pkg/openstack/nova.go
```diff
@@ -100,3 +207,7 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
 
 	return ctrl.Result{}, nil
 }
+
+func getNoVNCProxyServiceLabel(name string, cellName string) string {
```
gibizer, line 211: Thanks!

pkg/openstack/placement.go
gibizer: I would switch [1] to internal just to be on the safe side. But obviously that is a separate change. 
I have no hard opinion about whether to keep or remove the owner ref.

gibizer (commented):
Thanks you for the updates. I have some further comments. I see two condition handling issue we need to solve.

pkg/openstack/nova.go
gibizer: Sorry I mixed up. You return on each error. So the only overlap is when a previous success is overridden by a later failure. But that I guess is OK. The above described problem only arises when we switch to parallel route creation. But we don't need that now. So you can ignore the above comment

pkg/openstack/common.go
```diff
@@ -27,3 +39,213 @@ func EnsureDeleted(ctx context.Context, helper *helper.Helper, obj client.Object
...
+				// Any trailing path will be added on the service-operator level.
+				svcOverride.EndpointURL = &rd.endpointURL
+				instance.Status.Conditions.MarkTrue(condType, corev1.OpenStackControlPlaneExposeServiceReadyMessage, owner.GetName())
+			}
+		}
+
+		svcOverrides[rd.Endpoint] = svcOverride
+	}
+
+	if _, ok := cleanCondition[true]; !ok {
```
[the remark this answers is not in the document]
stuggi: yes, this now can simplified after I did some reorg of this function. thanks for the pointer.

pkg/openstack/nova.go
stuggi: right, I think its ok to return on the first failure.

pkg/openstack/nova.go
stuggi: done

pkg/openstack/nova.go
stuggi: for now I switched to use NovaAllCellReady. I'd like to wait for do a change from one of the options until we are further into the TLS topic where we also update the service operator CR with certs to be used for TLS-E.

pkg/openstack/nova.go
stuggi: done

pkg/openstack/common.go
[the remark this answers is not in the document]
stuggi: done

pkg/openstack/nova.go
gibizer: OK. The current nova-operator impl makes the Cell ready without waiting for the EndpointURL set so the current code works. And we can refine it later.

pkg/openstack/common.go
```diff
@@ -27,3 +39,214 @@ func EnsureDeleted(ctx context.Context, helper *helper.Helper, obj client.Object
...
+					owner.GetName(),
+					rd.RouteName,
+					err.Error()))
+
+				return svcOverrides, ctrlResult, err
+			} else if (ctrlResult != ctrl.Result{}) {
+				return svcOverrides, ctrlResult, nil
+			}
+
+			cleanCondition = true
```
gibizer, line 197: At this point we have a Route created successfully. So I'm not sure why we set the clearCondition to true here

pkg/openstack/common.go
```diff
@@ -27,3 +39,214 @@ func EnsureDeleted(ctx context.Context, helper *helper.Helper, obj client.Object
...
+
+			// update override for the service with the route endpoint url
+			if rd.endpointURL != "" {
+				// Any trailing path will be added on the service-operator level.
+				svcOverride.EndpointURL = &rd.endpointURL
+				instance.Status.Conditions.MarkTrue(condType, corev1.OpenStackControlPlaneExposeServiceReadyMessage, owner.GetName())
```
gibizer, line 203: Will this condition gets cleared at L211 as we set cleanCondition at L197 after the Route was created?

pkg/openstack/nova.go
```diff
@@ -45,9 +51,110 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
...
+			if err != nil {
+				return ctrlResult, err
+			} else if (ctrlResult != ctrl.Result{}) {
+				return ctrlResult, nil
+			}
+			instance.Spec.Nova.Template.CellTemplates[cellName] = cellTemplate
+
+		}
+	}
+	//instance.Status.Conditions.MarkTrue(corev1beta1.OpenStackControlPlaneServiceOverrideReadyCondition, corev1beta1.OpenStackControlPlaneServiceOverrideReadyMessage)
```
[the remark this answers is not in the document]
stuggi: yes, it seems that is from an old push. I already removed these commented conditions

pkg/openstack/common.go
stuggi: yes, this is wrong. I have to flip the `cleanCondition` on initialization and set it to false when the route got created.

pkg/openstack/common.go
stuggi: see above

stuggi:
recheck

abays (approved):
/lgtm

stuggi:
rebased

rabi:
recheck

Jobs are fixed

rabi:
recheck

stuggi:
restoring lgtm from @abays , just rebased

abays:
> @stuggi: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> Test name 	Commit 	Details 	Required 	Rerun command
> ci/prow/openstack-operator-build-deploy-kuttl 	[a1e296f](https://github.com/openstack-k8s-operators/openstack-operator/commit/a1e296f292ad21ca4bfd9782fe9a1d8dad758498) 	[link](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_openstack-operator/457/pull-ci-openstack-k8s-operators-openstack-operator-main-openstack-operator-build-deploy-kuttl/1705946833336930304) 	true 	`/test openstack-operator-build-deploy-kuttl`
> 
> [Full PR test history](https://prow.ci.openshift.org/pr-history?org=openstack-k8s-operators&repo=openstack-operator&pr=457). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:stuggi).

Hiccup

```
Removing debug pod ...
error: unable to upgrade connection: container container-00 not found in pod oko-15-x49j4-master-0-debug_openstack-kuttl-tests
```

abays:
> @stuggi: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> Test name 	Commit 	Details 	Required 	Rerun command
> ci/prow/openstack-operator-build-deploy-kuttl 	[a1e296f](https://github.com/openstack-k8s-operators/openstack-operator/commit/a1e296f292ad21ca4bfd9782fe9a1d8dad758498) 	[link](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_openstack-operator/457/pull-ci-openstack-k8s-operators-openstack-operator-main-openstack-operator-build-deploy-kuttl/1705964068893888512) 	true 	`/test openstack-operator-build-deploy-kuttl`
> 
> [Full PR test history](https://prow.ci.openshift.org/pr-history?org=openstack-k8s-operators&repo=openstack-operator&pr=457). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:stuggi).

Same transient error as last time.

PR #459, opened 2023-09-11
Enable webhooks for dataplane-operator

rabi:
Now webhooks are enabled in every operator with separate CSVs other than
dataplane-operator. dataplane-operator is merged in openstack-operator
and it's webhook has to be included. Atm, it's disabled with ENABLE_WEBHOOK=false.


abays (approved):
/lgtm

PR #461, opened 2023-09-12
Switch the operator to use the multinode layout

pablintino:
Since we are getting rid off the nested EDPM jobs we are now changing the openstack-operator to use the multinode based EDPM deployment job.


stuggi (approved):
/lgtm

slagle (approved):

PR #462, opened 2023-09-12
Replace EDPM job redefinition with a proper template

pablintino:
To avoid redefining jobs that are exact copies of the ci-framework ones we are introducing zuul templates that already define those jobs. This will allow easier maintainability as a change that needs to be applied in the base job and in every child one (ex. var removal, type change, etc.) won't require a change in every repo, just one change in the base one in ci-framework.


abays:
```
case.go:366: resource OpenStackControlPlane:openstack-kuttl-tests/openstack-basic: .status.conditions.message: value mismatch, expected: Setup complete != actual: OpenStackControlPlane Nova in progress
```

abays (approved):
/lgtm

PR #468, opened 2023-09-15
Support disconnected network environments

dprince:
This PR adds support for installing the operator in disconnected network environments. To build with image-digests set USE_IMAGE_DIGESTS=true before running make bundle.

For Prow jobs we are enabling this via .prow-ci.env

This drops the old logic from create_bundle.sh which has been broken with operator-sdk's make bundle for some time.

(NOTE: this currently requires a secure container registry to test)

Jira: [OSP-26486](https://issues.redhat.com//browse/OSP-26486)

### DIFF
--- a/.github/create_bundle.sh
+++ b/.github/create_bundle.sh
@@ -1,69 +1,13 @@
 #!/bin/bash
 set -e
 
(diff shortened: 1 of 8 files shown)


abays (approved):
/lgtm

PR #475, opened 2023-09-19
Fix RabbitMQ image default/override logic

abays:
We were always setting the default for the RabbitMQ image, even if the user had provided an explicit image


olliewalsh (approved):
/lgtm

abays:
recheck

abays:
recheck

abays:
recheck

abays:
recheck

abays:
recheck

abays:
recheck

abays:
recheck

This keeps failing with hiccups on the `cifmw-crc-podified-edpm-baremetal` job.  One last try and then I'm overriding because we need this fix.

PR #478, opened 2023-09-20
Fix network isolation samples as a pre step for route mgmt

stuggi:


gibizer (approved):
Thanks I validated this manually locally. After the override is removed nova-compute can access the rabbitmq again from edpm.

PR #481, opened 2023-09-21
Test: add 30 ENVs to the manager pod

dprince:


dprince:
@bshephar this is the test to see if we have found another limitation

PR #488, opened 2023-09-25
Create Ironic and Octavia services routes based on annotations on the svc endpoint

stuggi:
This is a follow up to https://github.com/openstack-k8s-operators/openstack-operator/pull/457 to also migrate ironic and octavia to have their routes created via the openstack-operator.

Depends-On: https://github.com/openstack-k8s-operators/ironic-operator/pull/334
Depends-On: https://github.com/openstack-k8s-operators/octavia-operator/pull/171


pkg/openstack/ironic.go
```diff
@@ -29,9 +34,96 @@ func ReconcileIronic(ctx context.Context, instance *corev1beta1.OpenStackControl
...
+		)
+		if err != nil {
+			return ctrlResult, err
+		} else if (ctrlResult != ctrl.Result{}) {
+			return ctrlResult, nil
+		}
+	}
+
+	// Ironic Inspector
+	if ironic.Status.Conditions.IsTrue(ironicv1.IronicInspectorReadyCondition) {
```
hjensas, line 98: This condition is currently set to True by ironic-operator also when Inspector service is not requested (by setting replicas = 0). Would the remaining code here fail in that case? Or will `svcs` simply be empty an `EnsureRoute` would simply not create any resources?

Setting the condition to `True` when service is not requested in ironic-operator might not be the right thing to do ...

stuggi: when there is no service created by the ironic operator which has the IngressCreate annotation [1] (done by the service operator), there won't be a route created.

[1] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/service/types.go#L44

abays (approved):
/lgtm

PR #489, opened 2023-09-25
Fix nova-novncproxy route endpoint type

stuggi:
since there should be a route created for the novncproxy, the endpoint into EnsureRoute() needs to be of type public, not internal.


gibizer (approved):
I tested this locally and it solves the VNC routing problem for me.

PR #490, opened 2023-09-25
Get manila CR before checking the condition state of the api resource

stuggi:
The condition never met as the manila CR needs to be fetched.

### DIFF
--- a/pkg/openstack/manila.go
+++ b/pkg/openstack/manila.go
@@ -14,7 +14,9 @@ import (
 
 	manilav1 "github.com/openstack-k8s-operators/manila-operator/api/v1beta1"
 	corev1beta1 "github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1"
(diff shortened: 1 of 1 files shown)


pkg/openstack/manila.go
```diff
@@ -47,6 +49,13 @@ func ReconcileManila(ctx context.Context, instance *corev1beta1.OpenStackControl
...
 
+	// When component services got created check if there is the need to create a route
+	if err := helper.GetClient().Get(ctx, types.NamespacedName{Name: "manila", Namespace: instance.Namespace}, manila); err != nil {
+		if !k8s_errors.IsNotFound(err) {
+			return ctrl.Result{}, err
+		}
+	}
+
 	// When component services got created check if there is the need to create a route
 	if manila.Status.Conditions.IsTrue(manilav1.ManilaAPIReadyCondition) {
```
abays, line 60: If the error is `NotFound`, won't `manila` be `nil` here?

PR #493, opened 2023-09-28
Improve SHA query for different registries

viroel:
It now compare with the proper registry before sending api query. For registry != from quay.io (e.g. quay.rdoproject.org) don't support 'filter_tag_name' but might support 'limit', so we can retrieve more older tags.

### DIFF
--- a/hack/pin-bundle-images.sh
+++ b/hack/pin-bundle-images.sh
@@ -31,28 +31,31 @@ for MOD_PATH in $(go list -mod=readonly -m -json all | jq -r '. | select(.Path |
         REF=$(git ls-remote https://${GIT_REPO} | grep ${REF} | awk 'NR==1{print $1}')
     fi
     GITHUB_USER=$(echo $MOD_PATH | sed -e 's|github.com/\(.*\)/.*-operator/.*$|\1|')
-    REPO_CURL_URL="https://quay.io/api/v1/repository/openstack-k8s-operators"
-    REPO_URL="quay.io/openstack-k8s-operators"
+    CURL_REGISTRY="quay.io"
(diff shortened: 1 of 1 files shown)


hack/pin-bundle-images.sh
```diff
@@ -33,9 +33,11 @@ for MOD_PATH in $(go list -mod=readonly -m -json all | jq -r '. | select(.Path |
     GITHUB_USER=$(echo $MOD_PATH | sed -e 's|github.com/\(.*\)/.*-operator/.*$|\1|')
     REPO_CURL_URL="https://quay.io/api/v1/repository/openstack-k8s-operators"
     REPO_URL="quay.io/openstack-k8s-operators"
+    CURL_REGISTRY="quay.io"
```
rebtoor, line 36: what about using this var also in REPO_CURL_URL and REPO_URL?

viroel:
from results[[1]](https://logserver.rdoproject.org/93/493/af5d226dbee0e9c735d52c918d2def2e41de8fe1/github-check/openstack-k8s-operators-content-provider/6e0e88e/ci-framework-data/logs/ci_make_004_openstack-operator_-_call_bundle.log), using the proper api call for quay.io now, in content-provider job:
`++ curl -s 'https://quay.io/api/v1/repository/openstack-k8s-operators/cinder-operator-bundle/tag/?onlyActiveTags=true&filter_tag_name=like:1b9a7de3d088'`

hack/pin-bundle-images.sh
viroel: done

viroel:
recheck

stuggi:
recheck

cjeanner:
/lgtm

content-provider job passed. Let's unblock CI :)

pablintino:
recheck

rebtoor (approved):

PR #503, opened 2023-10-02
Handle `openstack-ansibleee-operator` repo in `pin-bundle-images.sh` script

rebtoor:
We noticed a lot of failures in the `openstack-k8s-operators-content-provider` job for the ansibleee operator repo so we discovered that the generation of the `REPO_CURL_URL` variable is not correct because the repo doesn't have the same naming convention as the other repositories (`openstack-ansibleee-operator` instead of `ansibleee-operator`) so the conditions in the if cycle that fill that variable ([here](https://github.com/openstack-k8s-operators/openstack-operator/blob/39bb237d6562856928c6dd0eb745bfebae54662e/hack/pin-bundle-images.sh#L37-L52)) aren't met because: 

1. GITHUB_USER is `openstack-k8s-operators` (since the PR is autogenerated by resonate)
2. BASE (`openstack-ansibleee`) is different from IMAGEBASE (`ansibleee`)


rebtoor:
recheck

fao89 (approved):

PR #508, opened 2023-10-10
Update network-isolation-ceph to use Glance image conversion

fultonj:
When Glance is configured with Ceph RBD it's recommended to enable image conversion. Update the example with Ceph and Glance accordingly.


fultonj:
We have a number of examples of glance and image conversion.

1. https://github.com/openstack-k8s-operators/architecture/pull/11
2. https://github.com/openstack-k8s-operators/docs/pull/68
3. https://github.com/openstack-k8s-operators/glance-operator/tree/main/config/samples/import_plugins/image_conversion

This change will break CI which uses this file is updated to accommodate it (e.g. creating the conversion PVC) or if the CI
jobs is updated.

I think the three above examples are sufficient so I'm abandoning this patch.

PR #515, opened 2023-10-13
Designate/routes

dkehn:


config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml
```diff
@@ -310,3 +310,33 @@ spec:
       secret: osp-secret
   redis:
     enabled: false
+  cinder:
```
stuggi, line 313: should be designate?

stuggi (changes requested):

config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml
dkehn: yes, it should

PR #518, opened 2023-10-14
Increase kuttl test default timeout

viroel:
While running tests in ci-framework, the collapsed-cell is failing with one missing exposed service. The logs show routes with status ok. This patch increase the default timeout to 23min (was 16).


viroel:
> basic

Yes @fmount - I was looking at the logs and it was everything ready, so I was testing increasing the time.

PR #521, opened 2023-10-19
Designate-operator integration

dkehn:


dprince:
We need to add designate to the top level dependencies.yaml. Otherwise that operator won't get installed via the meta operator installation via OLM.

PR #525, opened 2023-10-20
Rework custom-bundle.Dockerfile to use staged data

dprince:
This reworks custom-bundle.Dockerfile so that it involves less stages and can support 2 distinct steps: sync time and build time. This will more cleanly align to our downstream build needs

At sync time a bundle_extra_data directory is populated with ENV variables from all operators (required for webhooks). Additionally dataplane-operator CRDs are cached in this repo.

At build time the custom-bundle.Dockerfile pulls from bundle_extra_data directly.

The csv-merger is used at 'sync time' to extract ENV variables and again at 'build time' to merge those ENV variables along with the dataplane-operators into the combined openstack operator CSV.

Jira: [OSP-29916](https://issues.redhat.com//browse/OSP-29916)

### DIFF
--- a/hack/bundle-cache-data.sh
+++ b/hack/bundle-cache-data.sh
@@ -1,0 +1,71 @@
+#!/bin/bash
+
+# extract select data from bundles:
+#  -env vars from all service operators
+#  -dataplane-operator bundle is cached (in order to merge at build time)
+set -ex
+
+function extract_bundle {
+    local IN_DIR=$1
+    local OUT_DIR=$2
+    for X in $(file ${IN_DIR}/* | grep gzip | cut -f 1 -d ':'); do
+        tar xvf $X -C ${OUT_DIR}/;
+    done
+}
+
+function extract_csv {
+    local IN_DIR=$1
+    local OUT_DIR=$2
+    for X in $(file ${IN_DIR}/* | grep gzip | cut -f 1 -d ':'); do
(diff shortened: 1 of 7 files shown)


hack/bundle-cache-data.sh
```diff
@@ -0,0 +1,67 @@
...
+
+function extract_csv {
+    local IN_DIR=$1
+    local OUT_DIR=$2
+    for X in $(file ${IN_DIR}/* | grep gzip | cut -f 1 -d ':'); do
+        tar xvf $X -C ${OUT_DIR} --wildcards --no-anchor '**/*clusterserviceversion.yaml'
+    done
+}
+
+OUT_BUNDLE=bundle_extra_data
```
gibizer, line 22: do we want to put this dir into .gitignore

gibizer (commented):
Overall look OK to me. I was able to build the bundles locally with this PR.

hack/bundle-cache-data.sh
dprince: we can for now. In a future PR I was considering making this something we generated and commit on each PR. That would require that we get it working with Renovate and other CI actions as well.

dprince: added the .gitignore in the latest commit

gibizer: Thanks. I agree.

gibizer (commented):
+1 looks good to me

dprince:
recheck

abays (approved):
/lgtm

PR #531, opened 2023-10-23
Initial commit to add barbican to openstack-operator

vakwetu:
This is WIP right now.  We need to land a change to update barbican to use the override route first. Then we won't have to do some gymnastics to get the right image of barbican in place and we can test properly.


stuggi:
the integration looks good to me. best to land the barbican PR and then check the failing jobs here

vakwetu:
new pull request created with different branch

PR #532, opened 2023-10-24
Added a flag to toggle podman build cache while building images

rebtoor:


dprince:
@rebtoor should we make this configurable? It seems caching could be useful in some contexts right?

rebtoor:
> @rebtoor should we make this configurable? It seems caching could be useful in some contexts right?

Yeah, you're right.

BTW this patch didn't help with the problem that we faced so it could be an (unrequired) improvement. Let me push the fix.

dprince:
what about just a more generic PODMAN_BUILD_OPTS, similar to the already existing BUNDLE_METADATA_OPTS

PR #544, opened 2023-11-07
[openstackclient] demo - watch input resources

stuggi:
This PR demonstrates how to watch named input resources in two different ways, split into two commits:
* [[openstackclient] Use ownerref and watches on ca bundle secret](https://github.com/openstack-k8s-operators/openstack-operator/pull/544/commits/a8bcb44139f16840ae8988d164ee81aa7ea97510)
Also adds itself to the secret specified via the `CaSecretName` as an additional owner. Using `Watches` allows to get notified on a change of the resource. The only downside with this is that it won't get notified if the resource gets recreated from scratch.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/385

* [[openstackclient] watch named input resources](https://github.com/openstack-k8s-operators/openstack-operator/pull/544/commits/ce9708f0372cd79bbcce3aa66ceeb7c81aba151c)
Adds watches for name secret and configmap resources from the `OpenStackClient` CRD. This allows to watch the specific resources when they change. The advantage over the notification when added as additional owner is that it also reconciles when the resource gets re-created. ( https://kubebuilder.io/reference/watching-resources/externally-managed )

### DIFF
--- a/controllers/client/openstackclient_controller.go
+++ b/controllers/client/openstackclient_controller.go
@@ -33,6 +33,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	keystonev1 "github.com/openstack-k8s-operators/keystone-operator/api/v1beta1"
 	"github.com/openstack-k8s-operators/lib-common/modules/common"
@@ -40,6 +42,7 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/configmap"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/env"
 	helper "github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	object "github.com/openstack-k8s-operators/lib-common/modules/common/object"
 	common_rbac "github.com/openstack-k8s-operators/lib-common/modules/common/rbac"
 
 	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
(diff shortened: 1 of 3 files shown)


stuggi:
> When parameters like image or hash on input resources change, restart the openstackclient pod as its not possible to get those updated.
> 
> This PR also demonstrates how to watch named input resources in two different ways, split into two commits:
> 
> * [[openstackclient] Use ownerref and watches on ca bundle secret](https://github.com/openstack-k8s-operators/openstack-operator/commit/69fe2ecbbb907de69cfa1c0710890fa13951a893)
>   Also adds itself to the secret specified via the `CaSecretName` as an additional owner. Using `Watches` allows to get notified on a change of the resource. The only downside with this is that it won't get notified if the resource gets recreated from scratch.
> * [[openstackclient] watch named input resources](https://github.com/openstack-k8s-operators/openstack-operator/pull/544/commits/940bc84aa9f63fd5060ebd7a2a5f8a2078f0b4f0)
>   Adds watches for name secret and configmap resources from the `OpenStackClient` CRD. This allows to watch the specific resources when they change. The advantage over the notification when added as additional owner is that it also reconciles when the resource gets re-created.

The 2nd approach is similar to use a label watcher with the advantage that it can be used on resources a user creates and won't have the expected label, like password secret.

```
		Watches(&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(
			func(o client.Object) []reconcile.Request {
				labels := o.GetLabels()

				reconcileCR, hasLabel := labels[reconcileLabel]
				if !hasLabel {
					return []reconcile.Request{}
				}

				return []reconcile.Request{
					{NamespacedName: types.NamespacedName{
						Name:      reconcileCR,
						Namespace: o.GetNamespace(),
					}},
				}
			},
		)).
```

controllers/client/openstackclient_controller.go
```diff
@@ -360,14 +373,108 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		openStackConfigMapField,
+		openStackConfigSecretField,
+	}
+)
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *OpenStackClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
+	// index caSecretName
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &clientv1.OpenStackClient{}, caSecretNameField, func(rawObj client.Object) []string {
```
gibizer, line 395: OK so the field indexer based solution looks better than what we have today in nova-operator where we list all the secret and filter in golang https://github.com/openstack-k8s-operators/nova-operator/blob/0578e9c7d47dd718b46eacd1be4bfbf49bb54285/controllers/common.go#L442

PR #550, opened 2023-11-10
Bump operators api version to `v0.2.0`

rebtoor:


dprince:
As discussed, rather than doing this we should run a 'make force-bump' on dev-preview2 and then push that as a PR

PR #562, opened 2023-11-17
Add Barbican

vakwetu:
This commit adds the barbican operator to the openstack-operator.


vakwetu:
recheck

config/samples/core_v1beta1_openstackcontrolplane.yaml
abays, line 1: One question here and for all the other samples: do we intend to have Barbican enabled by default?  I'm just asking because I don't know one way or the other.  Right now it's enabled.

abays:
One thing I've noticed while trying this out is that the `rabbitMqClusterName` field is empty in the `Barbican` section of the `OpenStackControlPlane` CR.  This results in the control plane never finishing deployment, because the `TransportURL` created by the Barbican controller then uses that empty string as the RabbitMQ cluster name (which the Infra operator considers an error and thus it goes no further).

I see that other operators have the `rabbitMqClusterName` in their base spec struct as opposed to a template struct that is in-lined [1].  Perhaps using that same approach would fix the problem here.  I think the webhooks and the kubebuilder defaulting are interfering here and are setting the `rabbitMqClusterName` to the empty string, which is somehow tricking the validation into allowing such a zero-value as if the user had explicitly provided it.  We've seen similar problems in other operators.

I think this is the cause of the CI failure too:

```
TASK [edpm_prepare : Wait for OpenStack controlplane to be deployed _raw_params=oc wait OpenStackControlPlane {{ cr_name }} --namespace={{ cifmw_install_yamls_defaults['NAMESPACE'] }} --for=condition=ready --timeout={{ cifmw_edpm_prepare_timeout }}m] ***
Friday 17 November 2023  16:49:55 -0500 (0:00:00.724)       0:10:14.725 ******* 
fatal: [localhost]: FAILED! => ...
```

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/3c5c40e6cc3aba6e287b94d26770519f76ae9528/api/v1beta1/cinder_types.go#L55-L59

abays:
> One thing I've noticed while trying this out is that the `rabbitMqClusterName` field is empty in the `Barbican` section of the `OpenStackControlPlane` CR. This results in the control plane never finishing deployment, because the `TransportURL` created by the Barbican controller then uses that empty string as the RabbitMQ cluster name (which the Infra operator considers an error and thus it goes no further).
> 
> I see that other operators have the `rabbitMqClusterName` in their base spec struct as opposed to a template struct that is in-lined [1]. Perhaps using that same approach would fix the problem here. I think the webhooks and the kubebuilder defaulting are interfering here and are setting the `rabbitMqClusterName` to the empty string, which is somehow tricking the validation into allowing such a zero-value as if the user had explicitly provided it. We've seen similar problems in other operators.
> 
> I think this is the cause of the CI failure too:
> 
> ```
> TASK [edpm_prepare : Wait for OpenStack controlplane to be deployed _raw_params=oc wait OpenStackControlPlane {{ cr_name }} --namespace={{ cifmw_install_yamls_defaults['NAMESPACE'] }} --for=condition=ready --timeout={{ cifmw_edpm_prepare_timeout }}m] ***
> Friday 17 November 2023  16:49:55 -0500 (0:00:00.724)       0:10:14.725 ******* 
> fatal: [localhost]: FAILED! => ...
> ```
> 
> [1] https://github.com/openstack-k8s-operators/cinder-operator/blob/3c5c40e6cc3aba6e287b94d26770519f76ae9528/api/v1beta1/cinder_types.go#L55-L59

I did some more testing.  It seems like the struct definition as-is is actually fine.  Something else must be causing CI to fail.

vakwetu:
recheck

tests/kuttl/common/assert-sample-deployment.yaml
```diff
@@ -159,6 +159,17 @@ spec:
         replicas: 0 # backend needs to be configured
       designateBackendbind9:
         replicas: 0 # backend needs to be configured
+  barbican:
+    enabled: false
```
abays, line 163: The current samples have Barbican enabled by default, since they don't have `enabled: false` in the YAML.  So we need to reflect that here.  Also, since Barbican is enabled by default, its condition entry needs to be added to the `status.conditions` list.

abays: It also needs an entry for the Barbican "expose" condition:

```
    - message: OpenStackControlPlane barbican service exposed
      reason: Ready
      status: "True"
      type: OpenStackControlPlaneExposeBarbicanReady
```

That should go after the `OpenStackControlPlaneClientReady` condition.

dmendiza: Yes, per JP Jung, the current plan is to always enable Barbican.  This also means other projects that integrate with Barbican will be set up to use it by default. e.g. encrypted volumes will store keys there, etc.

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -155,6 +155,10 @@ func (r *OpenStackControlPlane) checkDepsEnabled(name string) string {
 		if !((r.Spec.Mariadb.Enabled || r.Spec.Galera.Enabled) && r.Spec.Memcached.Enabled && r.Spec.Keystone.Enabled) {
 			reqs = "MariaDB or Galera, Memcached, Keystone"
 		}
+	case "Barbican":
+		if !((r.Spec.Mariadb.Enabled || r.Spec.Galera.Enabled) && r.Spec.Memcached.Enabled && r.Spec.Keystone.Enabled) {
```
dmendiza, line 159: IIRC, Barbican does not depend or use Memcached in any way.

vakwetu: fixed

vakwetu:
recheck

stuggi:
recheck

gibizer:
To disable the volume encryption tempest tests you can try to add
```
compute-feature-enabled.attach_encrypted_volume: false
```
to  https://github.com/openstack-k8s-operators/openstack-operator/blob/1425f890060a3fcb2f39a3d35c9f422028050b66/zuul.d/projects.yaml#L2-L9 
With a comment explaining that it can only be enabled when we have cinder backed configured in these tests.

vakwetu:
recheck

vakwetu:
recheck

abays:
recheck

```
error: Failed to attach interface\nerror: Cannot get interface MTU on 'default': No such device
```

abays:
Between the last two Zuul attempts, we have full success for the 4 jobs there.  Let's try one more recheck and then we can override if it flakes-out again.

abays (approved):
/lgtm

abays:
The Zuul jobs have all passed independently over several attempts, so I think we're good there despite the hiccups.  I will override by the end of the day unless another reviewer requests changes to the PR.

gibizer:
I try to deploy this from a local build and I see that for me barbican is disabled by default even though the code suggests it should be enabled by default.

gibizer:
> I try to deploy this from a local build and I see that for me barbican is disabled by default even though the code suggests it should be enabled by default.

I used the wrong sample and updating it from the sample from this PR did not work. But recreating the OpenStackControlPlane from the sample from this PR works. Sorry for the noise

PR #568, opened 2023-11-24
Deduplicate related images

fao89:
Ensuring controlplane and dataplane operator share the same hash when the related image has the same name

context: https://issues.redhat.com/browse/OSPRH-1504
https://github.com/openstack-k8s-operators/dataplane-operator/pull/534

### DIFF
--- a/cmd/csv-merger/csv-merger.go
+++ b/cmd/csv-merger/csv-merger.go
@@ -27,6 +27,7 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/imdario/mergo"
 	"github.com/openstack-k8s-operators/openstack-operator/pkg/util"
+	"golang.org/x/exp/maps"
 
 	semver "github.com/blang/semver/v4"
 	"github.com/operator-framework/api/pkg/lib/version"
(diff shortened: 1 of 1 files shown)


fao89:
recheck

dprince:
@fao89 so I'm mixed on doing validations with csv-merger. At some point I might actually like to eliminate this tool however it is still required for now.

One idea for validations is to perhaps use 'scorecard' tests. Scorecard allows use to have custom validations for which I have started putting together a few things here: https://github.com/dprince/scorecard-openstack/blob/main/main.go I presented this a few weeks back and it didn't get much interest. But I still think it is perhaps a good validation layer. Wdyt?

dprince:
Also, to follow up on what I said about 'bundle validate' I think it is actually 'USE_IMAGE_DIGESTS=true make bundle' which checks for duplicate images. So not exactly what you'd need here.

dprince:
> I liked the scorecard tests, it would same me time if we already had them (I had to do some hacking to manually test the CSV).
> 
> Actually, about the duplication, when I started this [openstack-k8s-operators/dataplane-operator#534](https://github.com/openstack-k8s-operators/dataplane-operator/pull/534) it was breaking the kuttl tests because the CSV was invalid, that's what made me start this PR. Having the same var name in dataplane-operator and ovn-operator was duplicating the vars as we just use append, so this was what motivated me to start this PR

Do you have a link showing the failure you hit? That PR is passing now it seems

fao89:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_dataplane-operator/534/pull-ci-openstack-k8s-operators-dataplane-operator-main-dataplane-operator-build-deploy-kuttl/1728075030333493248

fao89:
@dprince did you have a chance to review this PR?
Some parts are "dataplane smart" but others are needed, for example: the related images are currently being dropped

fao89:
closing in favor of https://github.com/openstack-k8s-operators/openstack-operator/pull/712

PR #574, opened 2023-11-29
Flatten telemetry resources

paramite:
This patch replaces Ceilometer resource with Telemetry resouse to avoid adding Autoscaling resource. The Telemetry resource is deploying both Ceilometer and Autoscaling resources and will deploy also Logging resource in future.


paramite:
@dprince @abays: Hey guys, can I please be added to org, so that I don't need /ok-to-test flag in my future PRs?

abays:
> @dprince @abays: Hey guys, can I please be added to org, so that I don't need /ok-to-test flag in my future PRs?

Invite sent.

abays:
I think this will require KUTTL changes as well, since we reference Ceilometer there.

paramite:
True, will fix it.

dprince:
in general PR looks good to me. Fix the pre-commit and kuttl and we should be good to go here

jlarriba:
I dont know what happens, the precommit check should be working fine, as I pushed the manifests after "make manifests". If I download the patch, the pre-commit check passes succesfully in my env.

apis/bases/client.openstack.org_openstackclients.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.11.1
+    controller-gen.kubebuilder.io/version: v0.10.0
```
abays, line 6: @jlarriba I think these are the problem for precommit.  We need to run `make manifests` with the newer version of `controller-gen` and push that to the PR.

paramite:
Sorry for the delay. Now it should be mergable.

PR #584, opened 2023-12-04
[bundle-cache-data] handle multiple gzip per bundle

gibizer:
It seems that in the CI build (not in the local builds) we might have multiple gzip (layers?) per bundle and not all of them having a csv file. This fix is ignores the error when the csv is missing.


fmount:
tested via https://github.com/openstack-k8s-operators/manila-operator/pull/188 and Prow jobs are green.

stuggi (approved):
/lgtm

PR #588, opened 2023-12-05
Fx load balancer config for barbican

vakwetu:
This fixes the load balancer config.  With these settings, we see the loadbalancer config set correctly.


gibizer (approved):
Looks good. With this change I was able to locally integrate nova and barbican.

SeanMooney (approved):
makes sense to me

stuggi (approved):
/lgtm

PR #600, opened 2023-12-13
Re-enable encrypted volume test

vakwetu:
Now that barbican is integrated with cinder and nova, re-enable this test to make sure it still works

### DIFF
--- a/zuul.d/projects.yaml
+++ b/zuul.d/projects.yaml
@@ -10,10 +10,6 @@
             compute-feature-enabled.vnc_console: true
             compute-feature-enabled.stable_rescue: true
             compute_feature_enabled.hostname_fqdn_sanitization: true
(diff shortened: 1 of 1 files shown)


ASBishop:
I don't have experience defining these jobs, but from what I can tell from the logs the controlplane isn't defining any cinder-volume service. You end up with cinder-api and cinder-scheduler but no cinder-volume backend(s). The cinder gang is still (slowly) marching toward support for the LVM backend, which will be handy for CI purposes, but until then we'll need to limit testing volume encryption to a job that includes ceph in order to get a cinder RBD backend.

dprince (changes requested):
marking changes needed based on feedback

stuggi:
@vakwetu is this PR still needed?

vakwetu:
No longer needed.

PR #603, opened 2023-12-18
Use latest telemetry-operator version

jlarriba:


jlarriba:
This is not going to pass with the update of dependencies that is already included in telemetry-operator, all operators should be upgraded at the same time.

PR #618, opened 2024-01-08
Allow to customize service operator tag in pin-bundle-images

viroel:
This patch allow us to provide a custom tag to service operator, defined by IMAGEBASE, while pinning all bundles using 'pin-bundle-images' script. This will enable us to build bundles with custom tags in Prow, avoiding conflicts with concurrent jobs.

### DIFF
--- a/hack/pin-bundle-images.sh
+++ b/hack/pin-bundle-images.sh
@@ -8,6 +8,7 @@ DOCKERFILE=${DOCKERFILE:-""}
 IMAGENAMESPACE=${IMAGENAMESPACE:-"openstack-k8s-operators"}
 IMAGEREGISTRY=${IMAGEREGISTRY:-"quay.io"}
 IMAGEBASE=${IMAGEBASE:-}
+IMAGECUSTOMTAG=${IMAGECUSTOMTAG:-}
 LOCAL_REGISTRY=${LOCAL_REGISTRY:-0}
 
 if [ -n "$DOCKERFILE" ]; then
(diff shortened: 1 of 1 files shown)


viroel:
/hold

under tests

viroel:
recheck

viroel:
Without changing any job's step, it seems that the build continues to work:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_openstack-operator/618/pull-ci-openstack-k8s-operators-openstack-operator-main-openstack-operator-build-deploy-kuttl/1744738953837154304/artifacts/openstack-operator-build-deploy-kuttl/openstack-k8s-operators-build/build-log.txt

viroel:
/hold

until we merge https://github.com/openstack-k8s-operators/openstack-operator/pull/618

hack/pin-bundle-images.sh
```diff
@@ -26,10 +27,17 @@ for MOD_PATH in $(go list -mod=readonly -m -json all | jq -r '. | select(.Path |
 
     GIT_REPO=${MOD_PATH%"/apis"}
     GIT_REPO=${GIT_REPO%"/api"}
-    REF=$(echo $MOD_VERSION | sed -e 's|v[0-9]*.[0-9]*.[0-9]*-.*[0-9]*-\(.*\)$|\1|')
-    if [[ "$REF" == v* ]]; then
-        REF=$(git ls-remote https://${GIT_REPO} | grep ${REF} | awk 'NR==1{print $1}')
+
+    # Check if there is a custom tag for IMAGEBASE operator
+    if [[ -n "$IMAGECUSTOMTAG" && "$BASE" == "$IMAGEBASE" ]]; then
```
stuggi, line 32: how about where "$BASE" != "$IMAGEBASE" ? this is mostly for CI, but e.g. for me my github user is != quay, so "$BASE" != "$IMAGEBASE". do we have to consider that, or do we expect not to use a custom tag for that case?

viroel: No sure if I understood the question. BASE here is the operator in the loop of all required operators in the meta bundle.
When we want to test an service operator image, e.g. neutron-operator, we provide IMAGEBASE=neutron, and IMAGEREGISTRY + IMAGENAMESPACE from where to pull the recently builded neutro-operator-bundle. So in that case, we want to replace things only when BASE == IMAGEBASE
This PR would allow us to also change the tag.

viroel: So for  "$BASE" != "$IMAGEBASE" it will continue to get REF from:
` REF=$(git ls-remote https://${GIT_REPO} | grep ${REF} | awk 'NR==1{print $1}')`

viroel: I see your point now, but your scenario is very similar to our CI environment, since in CI we also use different registry and different namespace. But the script only replace those values when "$BASE" == "$IMAGEBASE", i.e. for a single operator. We need to improve this script to support more scenarios.

stuggi: sure, lets get this in and do improvements in follow ups

stuggi (approved):
/lgtm

PR #620, opened 2024-01-11
[tls] Add functionality to create service certs if internal tls enabled

stuggi:
- creates internal CA when internal TLS enabled
- adds functionality to create the certificates via CertManager for the k8s services used for the API endpoints. In follow ups when the service components allow to pass in the service certificate secrets those will be set on the component cr.

Adds information on created CAs for the endpoints and their expire information to the osctlplane CR:
```
    tls:
      caBundleSecretName: combined-ca-bundle
      endpoint:
        internal:
          expires: 2024-04-09 15:28:04 +0000 UTC
          name: rootca-internal
        public:
          expires: 2024-04-09 15:28:04 +0000 UTC
          name: rootca-public
```

Also refactors the nova pkg in preparation for better/easier tls cert create for metadata/vncproxy as they are not typical api endpoints.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/428

Jira: [OSPRH-3268](https://issues.redhat.com//browse/OSPRH-3268)


config/crd/bases/core.openstack.org_openstackcontrolplanes.yaml
```diff
@@ -15953,6 +15953,23 @@ spec:
...
                   type: object
                 type: array
+              tls:
+                properties:
+                  caBundleSecretName:
+                    type: string
+                  endpoint:
+                    additionalProperties:
+                      properties:
+                        expires:
```
olliewalsh, line 15963: should this have `format: date-time`?

pkg/openstack/ca.go
```diff
@@ -130,6 +131,19 @@ func ReconcileCAs(ctx context.Context, instance *corev1.OpenStackControlPlane, h
...
+			err = caOnlyBundle.getCertsFromPEM(caCert)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+
+			status := corev1.TLSCAStatus{
+				Name: caName,
+			}
+			if len(caOnlyBundle.certs) > 0 {
+				status.Expires = caOnlyBundle.certs[0].expire.String()
```
olliewalsh, line 143: Is this used anywhere else? Is it just so the use can see the expire time from the status?

config/crd/bases/core.openstack.org_openstackcontrolplanes.yaml
stuggi: you mean to format it different? right now its just the NotAfter string from the CA

stuggi: wondering if this should be just a list of CAs, instead of endpoint map. we know that there will be also others

olliewalsh: IIUC format is also used to be more specific about the type https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#dataTypes

stuggi: done

olliewalsh (approved):
/lgtm

stuggi:
recheck

PR #625, opened 2024-01-11
[tlse] internal TLS support for placement

stuggi:
Creates certs for k8s service of the service operator when spec.tls.endpoint.internal.enabled: true

For a service like nova which talks to multiple service internal endpoints, this has to be set for each of them for, like:

~~~
  customServiceConfig: |
    [keystone_authtoken]
    insecure = true
    [placement]
    insecure = true
    [neutron]
    insecure = true
    [glance]
    insecure = true
    [cinder]
    insecure = true
~~~

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/428
Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/620
Depends-On: https://github.com/openstack-k8s-operators/placement-operator/pull/92

Jira: [OSPRH-2368](https://issues.redhat.com//browse/OSPRH-2368)

### DIFF
--- a/pkg/openstack/placement.go
+++ b/pkg/openstack/placement.go
@@ -55,6 +55,12 @@ func ReconcilePlacementAPI(ctx context.Context, instance *corev1beta1.OpenStackC
 		}
 	}
 
+	// set CA cert and preserve any previously set TLS certs
+	if instance.Spec.TLS.Enabled(service.EndpointInternal) {
+		instance.Spec.Placement.Template.TLS = placementAPI.Spec.TLS
+	}
+	instance.Spec.Placement.Template.TLS.CaBundleSecretName = instance.Status.TLS.CaBundleSecretName
+
 	if placementAPI.Status.Conditions.IsTrue(condition.ExposeServiceReadyCondition) {
 		svcs, err := service.GetServicesListWithLabel(
 			ctx,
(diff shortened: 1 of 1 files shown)


stuggi:
needs to be rebased when https://github.com/openstack-k8s-operators/openstack-operator/pull/620 landed. only last commit is relevant for this PR

pkg/openstack/placement.go
```diff
@@ -55,6 +55,12 @@ func ReconcilePlacementAPI(ctx context.Context, instance *corev1beta1.OpenStackC
 		}
 	}
 
+	// set CA cert and preserve any previously set TLS certs
+	if instance.Spec.TLS.Enabled(service.EndpointInternal) {
+		instance.Spec.Placement.Template.TLS = placementAPI.Spec.TLS
+	}
+	instance.Spec.Placement.Template.TLS.CaBundleSecretName = instance.Status.TLS.CaBundleSecretName
```
gibizer, line 62: It is a bit strange that we are updating our own input fields (i.e. Spec) but if it is not causing any trouble with the controller-runtime / k8s then fine.

gibizer (commented):
look OK to me

pkg/openstack/placement.go
stuggi: we are not patching the CR and writing the spec in our own. the defer function is only updating metadata and status[1] . Setting it here on the instance because we copy the template to the placement spec `instance.Spec.Placement.Template.DeepCopyInto(&placementAPI.Spec)` 

[1] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/helper/helper.go#L193

gibizer: Oh cool then. Thanks for the explanation.

stuggi:
recheck

olliewalsh (approved):
/lgtm

PR #631, opened 2024-01-16
Version update

dprince:


controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,310 @@
...
+		instance.Status.ServicesNeedingUpdates = appendServiceIfMissing(instance.Status.ServicesNeedingUpdates, corev1beta1.OpenStackService{ServiceName: "glance"})
+	}
+
+	cinder := CheckCinderImages(ctx, Log, &controlPlane.Spec.Cinder.Template, instance.Spec.ServiceExcludes, instance.Spec.CinderVolumeExcludes, envVarsFromMain)
+	if !cinder {
+		instance.Status.ServicesNeedingUpdates = appendServiceIfMissing(instance.Status.ServicesNeedingUpdates, corev1beta1.OpenStackService{ServiceName: "cinder"})
+	}
+
+	if instance.Spec.TargetVersion != envOpenStackVersion {
+		return ctrl.Result{}, fmt.Errorf(fmt.Sprintf("Target version must equal: %s", envOpenStackVersion))
```
gibizer, line 183: I would move this to before the service specific checks or even before the r.lookupOpenStackControlPlane() as if the requested target is wrong then we don't need to do all that work. 
//later
Or move it to a validation webhook to provide direct feedback to the user that the specified value is wrong

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,310 @@
...
+		if !glance {
+			UpdateGlanceImages(ctx, Log, &controlPlane.Spec.Glance.Template, instance.Spec.ServiceExcludes, envVarsFromMain)
+			glanceService := corev1beta1.OpenStackService{ServiceName: "glance"}
+			instance.Status.DeployedVersions = appendServiceIfMissing(instance.Status.DeployedVersions, glanceService)
+			instance.Status.ServicesNeedingUpdates = removeService(instance.Status.ServicesNeedingUpdates, glanceService)
+		}
+		if !cinder {
+			UpdateCinderImages(ctx, Log, &controlPlane.Spec.Cinder.Template, instance.Spec.ServiceExcludes, instance.Spec.CinderVolumeExcludes, envVarsFromMain)
+			cinderService := corev1beta1.OpenStackService{ServiceName: "cinder"}
+			instance.Status.DeployedVersions = appendServiceIfMissing(instance.Status.DeployedVersions, cinderService)
```
gibizer, line 197: If I understand correctly this tracks the fact that cinder needs update / updated but the serviceExclude might excluded just CindeScheduler but not excluded CinderAPI. So there is a granularity mismatch

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,310 @@
...
+			instance.Status.ServicesNeedingUpdates = removeService(instance.Status.ServicesNeedingUpdates, cinderService)
+		}
+
+		err = helper.PatchInstance(ctx, controlPlane)
```
gibizer, line 201: I'm not sure it is correct. The helper was originally created with an OpenStackVersion not an OpenStackControlPlane instance. So the Patch will be created between objects of different types.

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,310 @@
...
+	"github.com/go-logr/logr"
+	cinderv1 "github.com/openstack-k8s-operators/cinder-operator/api/v1beta1"
+	glancev1 "github.com/openstack-k8s-operators/glance-operator/api/v1beta1"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	corev1beta1 "github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1"
+	"golang.org/x/exp/slices"
+)
+
+var envVarsFromMain (map[string]string)
+var envOpenStackVersion string
```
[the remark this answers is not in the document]
dprince: sure, no problem

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,310 @@
...
+
+	"github.com/go-logr/logr"
+	cinderv1 "github.com/openstack-k8s-operators/cinder-operator/api/v1beta1"
+	glancev1 "github.com/openstack-k8s-operators/glance-operator/api/v1beta1"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	corev1beta1 "github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1"
+	"golang.org/x/exp/slices"
+)
+
+var envVarsFromMain (map[string]string)
```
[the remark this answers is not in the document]
dprince: sure, done

controllers/core/openstackversion_controller.go
dprince: Agree, just haven't gotten to the webhooks work yet. I will add a comment to this effect

controllers/core/openstackversion_controller.go
dprince: I can double check, but I can say what is here seems to be working (functionally) just fine

dprince: Now creating 2 unique helpers. All should be well

gibizer:
Thanks for fixing my comments. What is in the PR looks good now. I will check back after it is out from Draft.

PR #640, opened 2024-01-25
Drop OVNController default external-ids from sample CRs

karelyatin:
These are defaults so not required to be duplicated.

Depends-On: https://github.com/openstack-k8s-operators/ovn-operator/pull/215


karelyatin:
Noticed some wierdness with default, so checking those

PR #645, opened 2024-01-30
ovn: fix ovn-controller cleanup on ovn disabled

booxter:
This PR fixes a bug in ovn cleanup code where ovn-controller was left hanging when ovn is disabled; it also adds test scenarios to validate basic ovn behavior when enabling and disabling the service.


abays (approved):
/lgtm

PR #647, opened 2024-01-31
Explicitly add enabled: true in the galera sample

fmount:
In the DP3 content galera is not enabled by default. This patch adds the missing boolean to make sure we don't be blocked by kubebuilder during the CR creation.


abays (approved):
/lgtm

abays:
recheck

fmount:
recheck

agarcia64 (approved):
I verified this change in my environment

PR #648, opened 2024-01-31
[TLS] Fix enable/disable of tls configuration

stuggi:
The following is the current expected behavior:
* When TLS was enable and gets disabled previous created certificates won't get deleted.

* adding additional custom CA certs can only be passed to the services if tls is enabled, otherwise CA bundle creation will be skipped and no bundle will be passed to the service CAs. Right now the default is to have TLS enabled for at least routes.

Jira: [OSPRH-3268](https://issues.redhat.com//browse/OSPRH-3268)


stuggi:
recheck

olliewalsh (approved):
/lgtm

PR #655, opened 2024-02-05
Attach storage network to Swift pods

cschwede:
With NetworkAttachments now being supported by Swift as well, these pods should be attached to the storage networks too.


abays:
recheck

```
error: You must be logged in to the server (Unauthorized)
```

stuggi (approved):
/lgtm

PR #659, opened 2024-02-08
Do not block Ready when OVN is disabled

gibizer:
The fix f7fe693e253652a50174285c498d6b325059a5e4 introduced a regression. If OVN is disabled in OpenStackControlPlane CR then the CR will never become Ready.

PR #661, opened 2024-02-08
Validate ready status is removed for ovn when it's disabled

booxter:
This is built on top of the other PR from @gibizer that fixed the regression. This PR adds necessary test coverage (plus removes some redundant Remove calls).

### DIFF
--- a/tests/functional/openstackoperator_controller_test.go
+++ b/tests/functional/openstackoperator_controller_test.go
@@ -22,7 +22,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	corev1 "k8s.io/api/core/v1"
+	k8s_corev1 "k8s.io/api/core/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
@@ -33,6 +33,7 @@ import (
 	. "github.com/openstack-k8s-operators/lib-common/modules/common/test/helpers"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
 	clientv1 "github.com/openstack-k8s-operators/openstack-operator/apis/client/v1beta1"
+	corev1 "github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1"
 	ovnv1 "github.com/openstack-k8s-operators/ovn-operator/api/v1beta1"
 )
 
@@ -162,10 +163,10 @@ var _ = Describe("OpenStackOperator controller", func() {
 					names.OpenStackClientName,
 					ConditionGetterFunc(OpenStackClientConditionGetter),
 					clientv1.OpenStackClientReadyCondition,
-					corev1.ConditionTrue,
+					k8s_corev1.ConditionTrue,
 				)
 
-				pod := &corev1.Pod{}
+				pod := &k8s_corev1.Pod{}
 				err := th.K8sClient.Get(ctx, names.OpenStackClientName, pod)
 				g.Expect(pod).Should(Not(BeNil()))
 				g.Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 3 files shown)


booxter:
The original fix (first patch in the series) is from https://github.com/openstack-k8s-operators/openstack-operator/pull/659

tests/functional/openstackoperator_controller_test.go
```diff
@@ -197,6 +198,8 @@ var _ = Describe("OpenStackOperator controller", func() {
 					},
 				},
 			}
+			// TODO: had to disable tls to allow control plane status to update
+			spec["tls"] = map[string]interface{}{}
```
booxter, line 202: @stuggi any idea why I have to disable tls to allow the CP to get to Ready in functional tests? I see errors in the log about expires field being empty and not conforming to date-time format when the reconcile loop attempts to update Status. (and then the test case here fails because Ready states never get updated on CP).

booxter:
recheck

tests/functional/openstackoperator_controller_test.go
gibizer: I think we have a bug here https://github.com/openstack-k8s-operators/openstack-operator/blob/00c81bcd4ba46e1fcb12df2a62a22eb166ea7eb7/pkg/openstack/ca.go#L138-L143

The TLSCAStatus.Expires fields expects a data-time but the above code does not initialize it if there is no certs, so it ends up being initialized to the go empty value `""`. 

I put an extra log there and I see:
```
  2024-02-09T09:55:30.298+0100	INFO	!!! Setting endpoint status	{"controller": "openstackcontrolplane", "controllerGroup": "core.openstack.org", "controllerKind": "OpenStackControlPlane", "OpenStackControlPlane": {"name":"3d79d521-10c2-4707-8462-a","namespace":"ee00d7df-3130-4715-b928-58d93c968689"}, "namespace": "ee00d7df-3130-4715-b928-58d93c968689", "name": "3d79d521-10c2-4707-8462-a", "reconcileID": "d36a4742-3150-42f4-acfd-da70f7be3dcd", "endpoint": "public", "status": {"name":"rootca-public","expires":""}, "len(caOnlyBundle.certs)": 0}
```

gibizer: In the test the PEM data is "test" (in base64) and that result in an empty bundle due to breaking here
https://github.com/openstack-k8s-operators/openstack-operator/blob/00c81bcd4ba46e1fcb12df2a62a22eb166ea7eb7/pkg/openstack/ca.go#L397-L402

gibizer: so we either need a bit better test data with certs in it or we need to make the expires field to allow being empty (maybe a pointer type allow nil without forcing the date-time validation, I don't know)

gibizer: For this test case and this PR I'm OK to keep the TLS disabled.

tests/functional/openstackoperator_controller_test.go
```diff
@@ -261,6 +280,12 @@ var _ = Describe("OpenStackOperator controller", func() {
 				instance := &ovnv1.OVNController{}
 				g.Expect(th.K8sClient.Get(th.Ctx, names.OVNControllerName, instance)).Should(Not(Succeed()))
 			}, timeout, interval).Should(Succeed())
+
+			// expect the ready status to become unknown since ovn is disabled
```
gibizer, line 284: The comment is wrong. We expect the OVN condition to be removed and the Ready condition to be True

tests/functional/openstackoperator_controller_test.go
stuggi: we discussed in our tls meeting a change we want to make to the openstack-operator for the CA handling. I will rework on this in the next week or the following. I'll then also add more tests. for now we could disable it as in this pr

gibizer (approved):
thanks!

stuggi (approved):
/lgtm

abays:
```
Removing debug pod ...
error: unable to upgrade connection: container container-00 not found in pod oko-14-lmt7f-master-1-debug-mnmpm_openstack-kuttl-tests
make[3]: *** [Makefile:526: crc_storage_cleanup] Error 1
```

booxter:
recheck

fao89 (approved):

PR #668, opened 2024-02-16
Use qualified container image names in Dockerfile

gibizer:
After we bumped to use the golang:1.20 image content provider job fails with.

```
Error: creating build container: short-name resolution enforced but cannot prompt without a TTY
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-ARG GOLANG_BUILDER=golang:1.20
+ARG GOLANG_BUILDER=docker.io/library/golang:1.20
 ARG OPERATOR_BASE_IMAGE=gcr.io/distroless/static:nonroot
 
 # Build the manager binary


bshephar (approved):

bshephar (approved):
We need to do this for FIPs anyway. Should we just implement all of the FIPs changes here while we're going?

gibizer:
with ubi9 I get 
```
[1/2] STEP 10/12: RUN mkdir -p ${DEST_ROOT}/usr/local/bin/
mkdir: cannot create directory '/dest-root': Permission denied
Error: building at STEP "RUN mkdir -p ${DEST_ROOT}/usr/local/bin/": while running runtime: exit status 1
make: *** [Makefile:155: docker-build] Error 1
``
so I fall back to golagn:1.20 with a qualified name

gibizer:
content provider passing, so this is a step in a good direction

abays (approved):
/lgtm

SeanMooney:
> with ubi9 I get
> 
> ```
> [1/2] STEP 10/12: RUN mkdir -p ${DEST_ROOT}/usr/local/bin/
> mkdir: cannot create directory '/dest-root': Permission denied
> Error: building at STEP "RUN mkdir -p ${DEST_ROOT}/usr/local/bin/": while running runtime: exit status 1
> make: *** [Makefile:155: docker-build] Error 1
> ``
> so I fall back to golagn:1.20 with a qualified name
> ```

that is why we had to add root here

https://github.com/openstack-k8s-operators/nova-operator/blob/main/Dockerfile#L18

the ubi image runs as uid 1001 and is not allowed to do the mkdir

we could use different directors but just using root in the build like the docker images was the approach we ended up taking

gibizer:
> > with ubi9 I get
> > ```
> > [1/2] STEP 10/12: RUN mkdir -p ${DEST_ROOT}/usr/local/bin/
> > mkdir: cannot create directory '/dest-root': Permission denied
> > Error: building at STEP "RUN mkdir -p ${DEST_ROOT}/usr/local/bin/": while running runtime: exit status 1
> > make: *** [Makefile:155: docker-build] Error 1
> > ``
> > so I fall back to golagn:1.20 with a qualified name
> > ```
> 
> that is why we had to add root here
> 
> https://github.com/openstack-k8s-operators/nova-operator/blob/main/Dockerfile#L18
> 
> the ubi image runs as uid 1001 and is not allowed to do the mkdir
> 
> we could use different directors but just using root in the build like the docker images was the approach we ended up taking

good point. I guess this will be done separately when the FIPS support is implemented for this operator

PR #678, opened 2024-02-19
[tlse] enable galera tls for internal TLS

stuggi:
Create cert for galera instance and configure the instance to use it. Galera will always be configured to support TLS as its on the DB user level if TLS will be enforced or not.

Jira: [OSPRH-2440](https://issues.redhat.com//browse/OSPRH-2440)

### DIFF
--- a/pkg/openstack/galera.go
+++ b/pkg/openstack/galera.go
@@ -5,14 +5,19 @@ import (
 	"fmt"
 	"strings"
 
+	certmgrv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
+	"github.com/openstack-k8s-operators/lib-common/modules/certmanager"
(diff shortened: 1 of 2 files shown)


Deydra71 (approved):

dciabrin:
I could not test it locally, but from my previous work on the mariadb-operator, the certificate created for galera should populate the 'organization' field, otherwise this trip the SST transfer setup done by wsrep_sst_rsync. This will not show at the initial deployment because there is no SST involved when the DB is bootstrapped as there are no data to be transferred yet.
In the [kuttl test](https://github.com/openstack-k8s-operators/mariadb-operator/blob/main/tests/kuttl/tests/galera_deploy_tls/01-tls-certificate.yaml#L15) from the mariadb-operator, I set the organization to 'cluster.local', and it is sufficient to make the SST scripts work with the certificate.

PR #680, opened 2024-02-20
[OSPRH-5019] Consider OpenStackClient pod status when calculating CR's ready state

abays:
We should consider the state of the `OpenStackClient`'s associated pod's state when setting the `OpenStackClientReadyCondition` on the `OpenStackClient` resource.  This fixes a problem where the parent `OpenStackControlPlane` resource mistakenly assumed that everything was fine when in fact the `OpenStackClient` pod was failing.


stuggi:
recheck

stuggi (approved):
/lgtm

PR #682, opened 2024-02-26
Use the target port name annotation for routes

vyzigold:
This uses the new annotation to specify the target port name used by routes created by the openstack-operator. In telemetry we don't manage the prometheus service, so for prometheus the assumption that the target port name is the same as the name of the service is incorrect.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/465

### DIFF
--- a/pkg/openstack/common.go
+++ b/pkg/openstack/common.go
@@ -410,13 +410,18 @@ func (ed *EndpointDetail) CreateRoute(
 	owner metav1.Object,
 ) (ctrl.Result, error) {
 	// initialize the route with any custom provided route override
+	// per default use the service name as targetPortName if we don't have the annotation.
+	targetPortName := ed.Service.Spec.Name
+	if name, ok := ed.Service.Spec.ObjectMeta.Annotations[service.AnnotationIngressTargetPortNameKey]; ok && name != "" {
+		targetPortName = name
+	}
 	enptRoute, err := route.NewRoute(
 		route.GenericRoute(&route.GenericRouteDetails{
 			Name:           ed.Name,
 			Namespace:      ed.Namespace,
 			Labels:         ed.Labels,
 			ServiceName:    ed.Service.Spec.Name,
-			TargetPortName: ed.Service.Spec.Name,
+			TargetPortName: targetPortName,
 		}),
 		time.Duration(5)*time.Second,
 		[]route.OverrideSpec{ed.Route.OverrideSpec},


pkg/openstack/common.go
```diff
@@ -410,13 +410,18 @@ func (ed *EndpointDetail) CreateRoute(
 	owner metav1.Object,
 ) (ctrl.Result, error) {
 	// initialize the route with any custom provided route override
+	targetPortName := ed.Service.Spec.ObjectMeta.Annotations[service.AnnotationIngressTargetPortNameKey]
+	if targetPortName == "" {
```
stuggi, line 414: should it be like this to check if the annotation exists and has a value?
```
// per default use the service name as targetPortName if we don't have the annotation.
targetPortName = ed.Service.Spec.Name
if name, ok := ed.Service.Spec.ObjectMeta.Annotations[service.AnnotationIngressTargetPortNameKey]; ok && name != ""{
        targetPortName = name
}
```

vyzigold: When I was testing this, I tested with a service which had the annotation and with a service, which didn't have it. Both cases behaved correctly. If I understand the suggestion correctly, it should also deal with the case, when the annotation isn't defined in lib-common, which was the issue of this PR before updating it. I feel like it's also a little bit more visible what's the default behavior. I'll try it out and use it if it works :+1:

stuggi: it would also deal with the situation where the annotation is defined, but the value is empty. so the annotation would only be used if it is set and has a non empty value

vyzigold: So it didn't help with the issue, when the annotation isn't defined in lib-common. Guess that's resolved during compile time (makes sense). As far as I can tell the functionality is the same, but I'll leave it in. Thanks for the suggestion.

vyzigold:
I rebased the PR on top of main, applied stuggi's suggestion and updated lib-common/modules/common, which now has the dependend PR merged.

vyzigold:
recheck
```
Requesting bearer token: invalid status code from registry 502 (Bad Gateway)
```

vyzigold:
recheck
Adoption job failed during step: "stop control plane services"

vyzigold:
I rebased on top of main. Main now has the required lib-common included.

vyzigold:
recheck
```
status.conditions.message: value mismatch, expected: Setup complete != actual: OpenStackControlPlane Nova in progress
```

vyzigold:
recheck
I used the wrong command before

stuggi (approved):
/lgtm

PR #683, opened 2024-02-27
Reconcile Barbican before Swift

cschwede:
If Barbican is enabled it will be used by Swift to create and retrieve secrets, thus reconciling Barbican before Swift.


abays (approved):
/lgtm

PR #692, opened 2024-03-05
Update pin-bundle-images.sh

lmiccini:
Pin to latest rabbit 2.6 + fips support


dprince:
looks good to me. Just needs to pass CI

dprince (approved):

PR #694, opened 2024-03-06
Regenerate go.mod indirect dependencies

gibizer:
During the golang version bump we unintentionally introduced a too new
indirect dependency on kube-openapi:

k8s.io/kube-openapi@v0.0.0-20240209001042-7a0d5b415232

This is not required by any of our direct deps and this is also a
version that required golang 1.21. If your system has golang 1.21
installed the this dependency might force go mod tidy to change the
golang version in the go.mod file to 1.21.

This patch removed all the indirect dependencies and let go mod tidy to
regenerate them to get rid of the too new kube-openapi version.

Also some service operators pulled in k8s.io version 0.29 which also golang
1.21. Since then this was fixed the service operators and now downgraded
here as well.


abays (approved):
/lgtm

gibizer:
recheck https://github.com/openstack-k8s-operators/edpm-ansible/pull/590 merged and should unblock the adoption job

PR #705, opened 2024-03-11
Concatenate global and service specific extraMounts

ASBishop:
The list of global and service specific extraMounts are now concatenated, instead of the previous behavior where service specific extraMounts would supersede/override the global list.

The services that support their own list of extraMounts is unchanged. Those services are cinder, glance, manila and neutron.


fmount (approved):
/lgtm

fmount (approved):

stuggi (approved):
/lgtm

PR #709, opened 2024-03-14
Revert "Bump kustomize"

dprince:
This reverts commit a0eb7c30553f9a25b11f33fc5fc509a813179afc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,7 +215,7 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
 GINKGO ?= $(LOCALBIN)/ginkgo
 
 ## Tool Versions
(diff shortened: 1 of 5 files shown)


abays (approved):
/lgtm

dprince:
The reverted commit breaks the openstack-operators CSV in that ENV variables are no longer merged correctly.

fao89 (approved):

fao89:
should I revert it in dataplane-operator also?

dprince:
@fao89 I haven't checked dataplane-operator yet. So long as the bundle still has images set we should be okay. Need to look into why this caused the issue here as well.

abays:
recheck

PR #712, opened 2024-03-15
OpenStackVersion: minor updates support

dprince:
This PR adds a new CRD for OpenStackVersion to help drive minor updates. When openstack-operator is deployed obtains all the 'containerImage' references from RELATED_IMAGE... env variables. A new OPENSTACK_RELEASE_VERSION env has been added that should be bumped for each release of the openstack-operator. 

Now during deployment of the OpenStackControlplane a new OpenStackVersion resource is created. (this can also be pre-created).
Then during reconciliation of the Controlplane it waits on the version to initialize and the 'targetVersion' is used to provide images to all the controlplane services.

Any time the openstack-controller manager is restarted (or updated) an new 'AvailableVersion' can be discovered. The targetVersion on the OpenStackVersion CR can then be set to any AvailableVersion to trigger the start of the minor update.

Because images are now managed on the OpenStackVersion CR the OpenStackControlplane image references have been removed by using new *core* API structs (core meaning without images... for the 'core' openstackcontrolplane CR)

The initial support here is for Controlplane only. Support for Dataplane services will follow.


controllers/core/openstackcontrolplane_controller.go
```diff
@@ -170,10 +172,81 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
...
+	Log.Info("Looking up the current OpenStackVersion")
+	ctrlResult, err, version := openstack.ReconcileVersion(ctx, instance, helper)
+	if err != nil {
+		return ctrl.Result{}, err
+	} else if (ctrlResult != ctrl.Result{}) {
+		return ctrlResult, nil
+	}
+
+	// wait until the version is initialized so we have images on the version.Status
+	if version.Status.Conditions.IsFalse(corev1beta1.OpenStackVersionInitialized) {
```
gibizer, line 184: I think it would be safer to say `!version.Status.Conditions.IsTrue(...)` so that if it is Unknown or nil then the controller also waits. See https://github.com/openstack-k8s-operators/lib-common/blob/fec16b14c33be8622459fa2a652e5553d4912e68/modules/common/condition/funcs.go#L146-L151

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -170,10 +172,81 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
...
+			return ctrl.Result{}, err
+		} else if (ctrlResult != ctrl.Result{}) {
+			Log.Info("Reconciling normal")
+			return ctrlResult, nil
+		}
+		// this will allow reconcileNormal to proceed in subsequent reconciles
+		instance.Status.DeployedVersion = &version.Spec.TargetVersion
+		return ctrl.Result{}, nil
+	} else {
+		if version.Status.Conditions.IsFalse(corev1beta1.OpenStackVersionMinorUpdateOVNControlplane) {
```
gibizer, line 202: I think it would be safer to say `!version.Status.Conditions.IsTrue` to properly handle when the condition is nil or Unknown

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -170,10 +172,81 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
...
+			Log.Info("Minor update OVN on the ControlPlane")
+			ctrlResult, err := r.reconcileOVNControllers(ctx, instance, version, helper)
+			if err != nil {
+				return ctrl.Result{}, err
+			} else if (ctrlResult != ctrl.Result{}) {
+				return ctrlResult, nil
+			}
+			instance.Status.DeployedOVNVersion = &version.Spec.TargetVersion
+			return ctrl.Result{}, nil
+		} else if version.Status.Conditions.IsFalse(corev1beta1.OpenStackVersionMinorUpdateControlplane) {
```
gibizer, line 212: ditto

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -170,10 +172,81 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
...
+			if err != nil {
+				return ctrl.Result{}, err
+			} else if (ctrlResult != ctrl.Result{}) {
+				return ctrlResult, nil
+			}
+			// this will allow reconcileNormal to proceed in subsequent reconciles
+			instance.Status.DeployedVersion = &version.Spec.TargetVersion
+			return ctrl.Result{}, nil
+		} else {
+			Log.Info("Skipping reconcile. Waiting on minor update to proceed.")
```
gibizer, line 224: I think the message here is misleading. At this point both the ovn and the control plane update is ready. I'm not even sure when we hit this branch. If both are ready then I expect that when the control plane finished it already set the DeployedVersion in the status so we should not end up here but in the normal reconcile part in the if branch.

Maybe this is the result of the bug that the else if branches are not handling the nil or Unknown condition states?

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -170,10 +172,81 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
...
+
+	OVNControllerReady, err := openstack.ReconcileOVNController(ctx, instance, version, helper)
+	if err != nil {
+		return ctrl.Result{}, err
+	} else if !OVNControllerReady {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			corev1beta1.OpenStackControlPlaneOVNReadyCondition,
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			corev1beta1.OpenStackControlPlaneOVNReadyRunningMessage))
```
gibizer, line 241: I would mirror the message of the Ready conditon form OVN to propagate the reason. https://github.com/openstack-k8s-operators/lib-common/blob/fec16b14c33be8622459fa2a652e5553d4912e68/modules/common/condition/funcs.go#L321

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,288 @@
...
+			envAvailableVersion = envArr[1]
+		}
+		if strings.HasPrefix(envArr[0], "RELATED_IMAGE_") {
+			localVars[envArr[0]] = &envArr[1]
+		}
+	}
+	envContainerImages = localVars
+}
+
+func compareStringPointers(a, b *string) bool {
```
gibizer, line 60: It does not read nicely in the conditions below. As it returns a boolean I would call it `equalStringPointers` or `stringPointersEqual`instead

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,288 @@
...
+	if err != nil {
+		// we ignore not found
+		if !k8s_errors.IsNotFound(err) {
+			return ctrl.Result{}, err
+		}
+	}
+
+	instance.Status.Conditions = condition.Conditions{}
+	// greenfield deployment
+	//cl := condition.Conditions{}
```
gibizer, line 125: the commented out code can be dropped

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,288 @@
...
+	if controlPlane == nil || controlPlane.Status.DeployedVersion == nil {
+		Log.Info("No controlplane or controlplane is not deployed")
+		return ctrl.Result{}, nil
+	}
+
+	// TODO minor update for OVN Dataplane in progress
+
+	// minor update for OVN Controlplane in progress
+	if instance.Status.DeployedVersion != nil && instance.Spec.TargetVersion != *instance.Status.DeployedVersion {
+		if controlPlane.Status.Conditions.IsFalse(condition.ReadyCondition) {
```
gibizer, line 209: I would use !IsTrue to handle nil and Unknown condition states.

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,288 @@
...
+			Log.Info("Minor update Controlplane is not in ready condition, waiting")
+			return ctrl.Result{}, nil
+		}
+		if !compareStringPointers(controlPlane.Status.ContainerImages.OvnControllerImage, instance.Status.ContainerImages.OvnControllerImage) ||
+			controlPlane.Status.Conditions.IsFalse(corev1beta1.OpenStackVersionMinorUpdateOVNControlplane) {
```
gibizer, line 214: ditto I would use !IsTrue

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,288 @@
...
+				corev1beta1.OpenStackVersionMinorUpdateReadyRunningMessage))
+			Log.Info("Minor update for OVN Controlplane in progress")
+			return ctrl.Result{}, nil
+		}
+		instance.Status.Conditions.MarkTrue(
+			corev1beta1.OpenStackVersionMinorUpdateOVNControlplane,
+			corev1beta1.OpenStackVersionMinorUpdateReadyMessage)
+
+		// minor update for Controlplane in progress
+		// we only check keystone here as it will only get updated during this phase
```
gibizer, line 228: Why don't we check all the service container images and the Ready condition of the OpenStackControlPlane instead to wait for all the services to be redeployed with the new image?

gibizer: Or do we need only keystone to be updated before we do the dataplane update?

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,288 @@
...
+			return ctrl.Result{}, nil
+		}
+		// TODO minor update for Dataplane in progress goes here
+
+		instance.Status.Conditions.MarkTrue(
+			corev1beta1.OpenStackVersionMinorUpdateControlplane,
+			corev1beta1.OpenStackVersionMinorUpdateReadyMessage)
+	}
+
+	if controlPlane.Status.Conditions.AllSubConditionIsTrue() {
```
gibizer, line 245: Can we just check the Ready condition of the controlPlane?

controllers/core/openstackversion_controller.go
dprince: the isReady conditions need to be carefully reviewed before we can consolidate this. I would like to check them all, and that would certainly be more complete. For now this was just sort of a spot check.

apis/core/v1beta1/openstackversion_types.go
```diff
@@ -0,0 +1,180 @@
...
+
+// struct that contains container image default URLs for each service (internal use only)
+type ContainerDefaults struct {
+	ContainerTemplate `json:",inline"`
+	CinderVolumeImage *string `json:"cinderVolumeImage,omitempty"`
+	ManilaShareImage  *string `json:"manilaShareImage,omitempty"`
+}
+
+// struct acts as the source of truth for container image URLs to be deployed
+type ContainerImages struct {
```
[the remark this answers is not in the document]
dprince: One is use for status: and the other spec: They do have different comments, and duplication here is minimal as most of the stuct is inherited from the ,inline ContainerTemplate.

apis/core/v1beta1/openstackversion_types.go
```diff
@@ -0,0 +1,180 @@
...
+package v1beta1
+
+import (
+	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	MinorUpdateOvnControlPlane string = "Minor Update OVN Controlplane In Progress"
+	MinorUpdateControlPlane    string = "Minor Update OVN Controlplane In Progress"
```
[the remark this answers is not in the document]
dprince: fixed in latest push

controllers/core/openstackversion_controller.go
dprince: this one line yes. dropping it.

controllers/core/openstackversion_controller.go
dprince: done in the latest push

apis/core/v1beta1/openstackversion_types.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// OpenStackVersionStatus defines the observed state of OpenStackVersion
+type OpenStackVersionStatus struct {
+	//+operator-sdk:csv:customresourcedefinitions:type=status,xDescriptors={"urn:alm:descriptor:io.kubernetes.conditions"}
+	// Conditions
+	Conditions condition.Conditions `json:"conditions,omitempty" optional:"true"`
+
+	DeployedVersion  *string `json:"deployedVersion,omitempty"`
+	AvailableVersion *string `json:"availableVersion,omitempty"`
```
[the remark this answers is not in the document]
dprince: I think that is correct

apis/core/v1beta1/openstackversion_types.go
```diff
@@ -0,0 +1,180 @@
...
+
+	// This is the source of truth for the container images to be deployed.
```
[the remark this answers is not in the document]
dprince: correct

controllers/core/openstackversion_controller.go
dprince: done

apis/core/v1beta1/openstackversion_types.go
```diff
@@ -0,0 +1,180 @@
...
+	ContainerImages ContainerImages `json:"containerImages,omitempty"`
+
+	// where we keep track of the container images for previous versions
```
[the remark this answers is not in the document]
gibizer: OK. If this is the intended behavior then I only ask for documenting that the volume and share override images for a new minor version can only be provided in the same k8s API request in which the targetVersion is updated to point to the new minor version.

apis/core/v1beta1/openstackversion_types.go
```diff
@@ -0,0 +1,180 @@
...
+	ContainerImageVersionDefaults map[string]*ContainerDefaults `json:"containerImageVersionDefaults,omitempty"`
```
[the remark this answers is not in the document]
gibizer: No I don't think it is a hard requirement right now either to disallow skipping a version, or keeping a history of deployed versions. Both can be done later.

apis/core/v1beta1/openstackversion_types.go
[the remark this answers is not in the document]
dprince: this will all be documented for sure. thanks

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,281 @@
...
+		Log.Info("No controlplane or controlplane is not deployed")
+		return ctrl.Result{}, nil
+	}
+
+	// TODO minor update for OVN Dataplane in progress
+
+	// minor update for OVN Controlplane in progress
+	if instance.Status.DeployedVersion != nil && instance.Spec.TargetVersion != *instance.Status.DeployedVersion {
+		if !compareStringPointers(controlPlane.Status.ContainerImages.OvnControllerImage, instance.Status.ContainerImages.OvnControllerImage) ||
+			!controlPlane.Status.Conditions.IsTrue(corev1beta1.OpenStackVersionMinorUpdateOVNControlplane) {
```
gibizer, line 207: Where do we set OpenStackVersionMinorUpdateOVNControlplane on the controlPlane CR?

gibizer (commented):
I have a question about how the sequencing of events between OVN and the rest of the services works. I understand that we need to update OVN first then update the rest of the control plane. (I ignore the dataplane for now.) I believe the proposed logic was the folllowing (even though I failed to find exactly this logic directly in the code).

1. openstack-operator restarts and picks up a new availableVersion from the ENV
2. human changes the targetVersion to the new availableVersion
3. openstack-operator rolls out the new contanerImage for OVN by updating The OVN CR
4. openstack-operator waits for the OVN CR to report that the new image is rolled out successfully
5. openstack-operator updates the containerImages of the rest of the control plane.

I understood that you discovered a bug in OVN that it reported Ready true earlier than the new image is fully rolled out. And it is being fixed in https://github.com/openstack-k8s-operators/ovn-operator/pull/252 by re-initializing conditions for each reconcile.

While I agree that we need to fix our condition handling logic, I believe such fix only reduce the likelihood of the occurrence of the bug but does not eliminate it.

When the controlplane controller updates the OVN spec with the new image the API request returns to controller after the defaulting and validation webhook succeeded. Also asynchronously a reconcile request is queued on the ovn controller as the CR changed.
At this point both the controllplane and the OVN controllers run independently and in parallel. If the controlplane controller checks the status of the OVN CR before the OVN controller handled the reconcile request then it will see that it is Ready=true, even though the new image is set, and the new image is not deployed yet. This issue cannot be fully fixed by just changing the condition logic in OVN controller as it can be that the controller had no chance to run on the new Spec content at all.

ciecierski:
> I have a question about how the sequencing of events between OVN and the rest of the services works. I understand that we need to update OVN first then update the rest of the control plane. (I ignore the dataplane for now.) I believe the proposed logic was the folllowing (even though I failed to find exactly this logic directly in the code).
> 
> 1. openstack-operator restarts and picks up a new availableVersion from the ENV
> 2. human changes the targetVersion to the new availableVersion
> 3. openstack-operator rolls out the new contanerImage for OVN by updating The OVN CR
> 4. openstack-operator waits for the OVN CR to report that the new image is rolled out successfully
> 5. openstack-operator updates the containerImages of the rest of the control plane.
> 
> I understood that you discovered a bug in OVN that it reported Ready true earlier than the new image is fully rolled out. And it is being fixed in [openstack-k8s-operators/ovn-operator#252](https://github.com/openstack-k8s-operators/ovn-operator/pull/252) by re-initializing conditions for each reconcile.
> 
> While I agree that we need to fix our condition handling logic, I believe such fix only reduce the likelihood of the occurrence of the bug but does not eliminate it.
> 
> When the controlplane controller updates the OVN spec with the new image the API request returns to controller after the defaulting and validation webhook succeeded. Also asynchronously a reconcile request is queued on the ovn controller as the CR changed. At this point both the controllplane and the OVN controllers run independently and in parallel. If the controlplane controller checks the status of the OVN CR before the OVN controller handled the reconcile request then it will see that it is Ready=true, even though the new image is set, and the new image is not deployed yet. This issue cannot be fully fixed by just changing the condition logic in OVN controller as it can be that the controller had no chance to run on the new Spec content at all.

One note on ignoring dataplane for now. There are ovn-controllers on dataplane too and those must be updated before the rest of control plane. Not to add more complexity, but 'Ready' state should take into account state of ovn-controllers state on dataplane too.

gibizer:
> 
> One note on ignoring dataplane for now. There are ovn-controllers on dataplane too and those must be updated before the rest of control plane. Not to add more complexity, but 'Ready' state should take into account state of ovn-controllers state on dataplane too.

Yeah, I only meant that ignoring dataplane for this comment only, not ignoring dataplane in this PR.

gibizer:
> I have a question about how the sequencing of events between OVN and the rest of the services works. I understand that we need to update OVN first then update the rest of the control plane. (I ignore the dataplane for now.) I believe the proposed logic was the folllowing (even though I failed to find exactly this logic directly in the code).
> 
> 1. openstack-operator restarts and picks up a new availableVersion from the ENV
> 2. human changes the targetVersion to the new availableVersion
> 3. openstack-operator rolls out the new contanerImage for OVN by updating The OVN CR
> 4. openstack-operator waits for the OVN CR to report that the new image is rolled out successfully
> 5. openstack-operator updates the containerImages of the rest of the control plane.
> 
> I understood that you discovered a bug in OVN that it reported Ready true earlier than the new image is fully rolled out. And it is being fixed in [openstack-k8s-operators/ovn-operator#252](https://github.com/openstack-k8s-operators/ovn-operator/pull/252) by re-initializing conditions for each reconcile.
> 
> While I agree that we need to fix our condition handling logic, I believe such fix only reduce the likelihood of the occurrence of the bug but does not eliminate it.
> 
> When the controlplane controller updates the OVN spec with the new image the API request returns to controller after the defaulting and validation webhook succeeded. Also asynchronously a reconcile request is queued on the ovn controller as the CR changed. At this point both the controllplane and the OVN controllers run independently and in parallel. If the controlplane controller checks the status of the OVN CR before the OVN controller handled the reconcile request then it will see that it is Ready=true, even though the new image is set, and the new image is not deployed yet. This issue cannot be fully fixed by just changing the condition logic in OVN controller as it can be that the controller had no chance to run on the new Spec content at all.

We discussed this with @mrkisaolamb and we see some possible ways forward. The main idea of each solution is that besides Ready=true also use an extra piece of information that tells the control plane controller that the the OVN controller saw the new spec content. This can be done by:
* looking at the image field of the OVN Pod spec and comparing it with the expected new value
* storing and comparing the input hash form OVN Status in the controlplane controller and wait until it changes after the image is changed in the OVN spec
* introducing a new status field in OVN that show what is the last reconciled resource version of the Spec. Then the controllplane operator can wait until the resource version in the status becomes the same as in the spec.

apis/core/v1beta1/openstackversion_types.go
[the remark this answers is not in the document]
gibizer: We figured out one more side of this with @fmount . The are two independent use cases for the share and volume image override fields:
* minor upgrade where the override image needs to change together with the rest of the service images. (This is what is discussed in this thread above)
* an override image patch / hotfix deployment without a minor upgrade. If only the 3rd party override image needs to change (due to e.g. a hotfix) then there is no way to trigger a minor update as there is no new availableVersion exists (that would require a new openstack-operator bundle build), but only the 3rd party image url is updated in the OpenStackVersion object, and that image update is rolled out immediately without any cross service coordination.

My point here is to let's document both use cases / solutions for the uses case.

dprince:
> > I have a question about how the sequencing of events between OVN and the rest of the services works. I understand that we need to update OVN first then update the rest of the control plane. (I ignore the dataplane for now.) I believe the proposed logic was the folllowing (even though I failed to find exactly this logic directly in the code).
> > 
> > 1. openstack-operator restarts and picks up a new availableVersion from the ENV
> > 2. human changes the targetVersion to the new availableVersion
> > 3. openstack-operator rolls out the new contanerImage for OVN by updating The OVN CR
> > 4. openstack-operator waits for the OVN CR to report that the new image is rolled out successfully
> > 5. openstack-operator updates the containerImages of the rest of the control plane.
> > 
> > I understood that you discovered a bug in OVN that it reported Ready true earlier than the new image is fully rolled out. And it is being fixed in [openstack-k8s-operators/ovn-operator#252](https://github.com/openstack-k8s-operators/ovn-operator/pull/252) by re-initializing conditions for each reconcile.
> > While I agree that we need to fix our condition handling logic, I believe such fix only reduce the likelihood of the occurrence of the bug but does not eliminate it.
> > When the controlplane controller updates the OVN spec with the new image the API request returns to controller after the defaulting and validation webhook succeeded. Also asynchronously a reconcile request is queued on the ovn controller as the CR changed. At this point both the controllplane and the OVN controllers run independently and in parallel. If the controlplane controller checks the status of the OVN CR before the OVN controller handled the reconcile request then it will see that it is Ready=true, even though the new image is set, and the new image is not deployed yet. This issue cannot be fully fixed by just changing the condition logic in OVN controller as it can be that the controller had no chance to run on the new Spec content at all.
> 
> One note on ignoring dataplane for now. There are ovn-controllers on dataplane too and those must be updated before the rest of control plane. Not to add more complexity, but 'Ready' state should take into account state of ovn-controllers state on dataplane too.

This is understood. See the commented out Dataplane conditions already in this PR. I have pushed this initial PR as without it we can't actually develop the Dataplane parts for which we'll be able to satisfy those conditions. And yes, there are already plans for a specific ready condition for Dataplane OVN updates.

apis/core/v1beta1/openstackversion_types.go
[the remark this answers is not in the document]
fmount: Right, as @dprince mentioned I think we're going to document both use cases for storage, including how to provide a hotfix, which is covered by `serviceCustomContainerImages` (`CustomContainerImages` in the `OpenStackVersion` API) (see `Custom images for other OpenStack services` in [1])

[1] https://github.com/openstack-k8s-operators/docs/pull/89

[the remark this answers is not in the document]
dprince: exactly. This is all in the works. The intent is to use OpenStackVersion for both of these cases.

controllers/core/openstackversion_controller.go
dprince: I fixed this and just haven't pushed it yet as I was testing a new pattern. The actual condition here is OpenStackControlPlaneOVNReadyCondition

dprince:
I added a new pattern to my ovn-operator PR for "observedGeneration". This should help make sure the spec is actually applied.

dprince:
recheck

dprince:
The RDO jobs are failing due to 'jq: error (at <stdin>:1): Cannot iterate over null (null)' during catalog-build. I think this is related to the go.mod updates.

raukadah:
recheck content provider health

raukadah:
recheck nodehold

apis/core/v1beta1/openstackversion_types.go
```diff
@@ -0,0 +1,197 @@
...
+	DesignateUnboundImage         *string `json:"designateUnboundImage,omitempty"`
+	DesignateWorkerImage          *string `json:"designateWorkerImage,omitempty"`
+	EdpmFrrImage                  *string `json:"edpmFrrImage,omitempty"`
+	EdpmIscsidImage               *string `json:"edpmIscsidImage,omitempty"`
+	EdpmLogrotateCrondImage       *string `json:"edpmLogrotateCrondImage,omitempty"`
+	EpdmNeutronMetadataAgentImage *string `json:"epdmNeutronMetadataAgentImage,omitempty"`
+	EdpmSriovAgentImage           *string `json:"edpmSriovAgentImage,omitempty"`
+	EdpmOvnBpgAgentImage          *string `json:"edpmOvnBpgAgentImage,omitempty"`
+	EpdmCeilometerComputeImage    *string `json:"epdmCeilometerComputeImage,omitempty"`
+	EpdmCeilometerIpmiImage       *string `json:"epdmCeilometerIpmiImage,omitempty"`
```
slagle, line 103: These should be EdpmTelemetryCeilometer... to match the env vars.

On further though, are these 2 ceilometer images needed to be specific to EDPM? Do we expect them to be different from CeilometerComputeImage and CeilometerIpmiImage?

apis/core/v1beta1/openstackversion_types.go
```diff
@@ -0,0 +1,197 @@
...
+	EdpmNodeExporterImage         *string `json:"edpmNodeExporterImage,omitempty"`
```
slagle, line 104: Should be EdpmTelemetryNodeExporterImage

dprince:
Looks like the RDO jobs are failing due to 
```deploymentStatuses:
    edpm-deployment-ipam:
    - lastTransitionTime: "2024-04-02T20:04:29Z"
      message: Deployment error occurred Secret "ceilometer-compute-config-data" not
        found```

File: apis/core/v1beta1/openstackversion_types.go (hunk 19 above)
Comment (jlarriba) [reply apis/core/v1beta1/openstackversion_types.go:103]: No, they are the same, but looking at all other image names, the correct images should be the Edpm ones. I will fix this.

**Comment by dprince** [comment]
> Looks like the RDO jobs are failing due to
> 
> ```
>     edpm-deployment-ipam:
>     - lastTransitionTime: "2024-04-02T20:04:29Z"
>       message: Deployment error occurred Secret "ceilometer-compute-config-data" not
>         found```
> ```

Looks like ceilometer isn't getting set to enabled. Code in pkg/openstack/telemetry.go likely isn't setting this up correctly

**Comment by dprince** [comment]
> > Looks like the RDO jobs are failing due to
> > ```
> >     edpm-deployment-ipam:
> >     - lastTransitionTime: "2024-04-02T20:04:29Z"
> >       message: Deployment error occurred Secret "ceilometer-compute-config-data" not
> >         found```
> > ```
> 
> Looks like ceilometer isn't getting set to enabled. Code in pkg/openstack/telemetry.go likely isn't setting this up correctly

Okay. Got it fixed up for now. Long term may need to convert to bool pointers in telemetry to make the defaulting work better there.

File: apis/core/v1beta1/openstackversion_types.go (hunk 20 above)
Comment (dprince) [reply apis/core/v1beta1/openstackversion_types.go:104]: Hmmm. In my CSV I see this defined like this: RELATED_IMAGE_OPENSTACK_EDPM_NODE_EXPORTER_IMAGE. We'll need to first rename this ENV var in order to make this name change

Comment (slagle) [reply apis/core/v1beta1/openstackversion_types.go:102]: Ok, I didn't realize that. This can stay the same then.

File: apis/core/v1beta1/openstackversion_types.go [hunk 21]
```diff
@@ -0,0 +1,195 @@
...
+	DesignateMdnsImage            *string `json:"designateMdnsImage,omitempty"`
+	DesignateProducerImage        *string `json:"designateProducerImage,omitempty"`
+	DesignateUnboundImage         *string `json:"designateUnboundImage,omitempty"`
+	DesignateWorkerImage          *string `json:"designateWorkerImage,omitempty"`
+	EdpmFrrImage                  *string `json:"edpmFrrImage,omitempty"`
+	EdpmIscsidImage               *string `json:"edpmIscsidImage,omitempty"`
+	EdpmLogrotateCrondImage       *string `json:"edpmLogrotateCrondImage,omitempty"`
+	EpdmNeutronMetadataAgentImage *string `json:"epdmNeutronMetadataAgentImage,omitempty"`
+	EdpmSriovAgentImage           *string `json:"edpmSriovAgentImage,omitempty"`
+	EdpmOvnBpgAgentImage          *string `json:"edpmOvnBpgAgentImage,omitempty"`
```
slagle, line 101: Bpg

apis/core/v1beta1/openstackversion_types.go
```diff
@@ -0,0 +1,195 @@
...
+	DesignateBackendbind9Image    *string `json:"designateBackendbind9Image,omitempty"`
+	DesignateCentralImage         *string `json:"designateCentralImage,omitempty"`
+	DesignateMdnsImage            *string `json:"designateMdnsImage,omitempty"`
+	DesignateProducerImage        *string `json:"designateProducerImage,omitempty"`
+	DesignateUnboundImage         *string `json:"designateUnboundImage,omitempty"`
+	DesignateWorkerImage          *string `json:"designateWorkerImage,omitempty"`
+	EdpmFrrImage                  *string `json:"edpmFrrImage,omitempty"`
+	EdpmIscsidImage               *string `json:"edpmIscsidImage,omitempty"`
+	EdpmLogrotateCrondImage       *string `json:"edpmLogrotateCrondImage,omitempty"`
+	EpdmNeutronMetadataAgentImage *string `json:"epdmNeutronMetadataAgentImage,omitempty"`
```
slagle, line 99: Edpm

apis/core/v1beta1/openstackversion_types.go
```diff
@@ -0,0 +1,195 @@
...
+	EdpmSriovAgentImage           *string `json:"edpmSriovAgentImage,omitempty"`
```
slagle, line 100: This one is RELATED_IMAGE_OPENSTACK_EDPM_NEUTRON_SRIOV_AGENT_DEFAULT_IMG, so it should be EdpmNeutrionSriovAgentImage

dprince:
The adoption RDO test failures are valid due to cinderVolumes not getting created correctly. A bug perhaps in the "deep copy" code for volumes in pkg/openstack/cinder

dprince:
recheck

dprince:
recheck

slagle (commented):
+2, I've tested this with https://github.com/openstack-k8s-operators/dataplane-operator/pull/757 and it's working enough to where I'm good for this to merge.

dprince:
recheck

cescgina:
recheck

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	cl := condition.CreateList(
+		condition.UnknownCondition(corev1beta1.OpenStackVersionInitialized, condition.InitReason, string(corev1beta1.OpenStackVersionInitializedInitMessage)),
+	)
+	// no minor update conditions unless we have a deployed version
+	if instance.Status.DeployedVersion != nil && instance.Spec.TargetVersion != *instance.Status.DeployedVersion {
+		cl = append(cl, *condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateOVNControlplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)),
+			*condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateControlplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)))
+		// fixme add dataplane conditions here
+		//condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateOVNDataplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)),
+		//condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateDataplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)))
```
gibizer, line 157: Is this expected to happen in a follow up PR? Or we hold the current PR until the dataplane side is implemented?

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	// store the defaults for the currently available version
+	if instance.Status.ContainerImageVersionDefaults == nil {
+		instance.Status.ContainerImageVersionDefaults = make(map[string]*corev1beta1.ContainerDefaults)
+	}
+	instance.Status.ContainerImageVersionDefaults[envAvailableVersion] = defaults
+
+	// calculate the container images for the target version
+	val, ok := instance.Status.ContainerImageVersionDefaults[instance.Spec.TargetVersion]
+	if !ok {
+		Log.Info("Target version not found in defaults", "targetVersion", instance.Spec.TargetVersion)
```
gibizer, line 182: Should we set some condition here to make this state visible? Is this an error state?

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+	}
+
+	// greenfield deployment //FIXME check dataplane here too
+	if controlPlane.Status.DeployedVersion == nil {
+		Log.Info("No controlplane or controlplane is not deployed")
+		return ctrl.Result{}, nil
+	}
+
+	// TODO minor update for OVN Dataplane in progress
```
gibizer, line 214: Will this come in a follow up PR or we hold this PR until the dataplane implementation is ready?

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+		if len(result) > 0 {
+			Log.Info("Reconcile request for:", "result", result)
+			return result
+		}
+		return nil
+	})
+
+	return ctrl.NewControllerManagedBy(mgr).
+		Watches(&corev1beta1.OpenStackControlPlane{}, versionFunc).
```
gibizer, line 292: Could you explain the logic behind this? How we are watching the the OpenStackControlPlane objects by querying the OpenStackVersion objects in the versionFunc?

pkg/openstack/barbican.go
```diff
@@ -142,6 +149,9 @@ func ReconcileBarbican(ctx context.Context, instance *corev1beta1.OpenStackContr
 			condition.SeverityInfo,
 			corev1beta1.OpenStackControlPlaneBarbicanReadyRunningMessage))
 	}
+	instance.Status.ContainerImages.BarbicanAPIImage = version.Status.ContainerImages.BarbicanAPIImage
```
gibizer, line 152: I guess we will check the observedGeneration on barbican along with IsReady before setting the images in our status as soon as observedGeneration field is available

pkg/openstack/cinder.go
```diff
@@ -139,7 +166,11 @@ func ReconcileCinder(ctx context.Context, instance *corev1beta1.OpenStackControl
 		Log.Info(fmt.Sprintf("Cinder %s - %s", cinder.Name, op))
 	}
 
-	if cinder.IsReady() {
+	if cinder.Status.ObservedGeneration == cinder.Generation && cinder.IsReady() {
```
gibizer, line 169: looks good!

pkg/openstack/nova.go
```diff
@@ -302,7 +309,12 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
 		Log.Info(fmt.Sprintf("Nova %s - %s", nova.Name, op))
 	}
 
-	if nova.IsReady() {
+	if nova.IsReady() { //FIXME ObservedGeneration
```
gibizer, line 312: observedGeneration is available since https://github.com/openstack-k8s-operators/nova-operator/commit/1559023133be0a6f248b7992414fa5a2dc74d75d

pkg/openstack/ovn.go
```diff
@@ -146,17 +158,26 @@ func ReconcileOVNDbClusters(ctx context.Context, instance *corev1beta1.OpenStack
...
 		if err != nil {
+			Log.Error(err, "Failed to reconcile OVNDBCluster")
 			return false, err
 		}
 		if op != controllerutil.OperationResultNone {
 			Log.Info(fmt.Sprintf("OVNDBCluster %s - %s", OVNDBCluster.Name, op))
 		}
-		OVNDBClustersReady = OVNDBClustersReady && OVNDBCluster.IsReady()
+
+		OVNDBClustersReady = OVNDBClustersReady && (OVNDBCluster.Status.ObservedGeneration == OVNDBCluster.Generation) && OVNDBCluster.IsReady()
```
gibizer, line 168: looks good. It is only ready if each OVN is up to date and ready.

pkg/openstack/placement.go
```diff
@@ -130,6 +131,7 @@ func ReconcilePlacementAPI(ctx context.Context, instance *corev1beta1.OpenStackC
 	}
 
 	if placementAPI.IsReady() {
```
gibizer, line 133: observedGeneration is available since https://github.com/openstack-k8s-operators/placement-operator/commit/c710647a7745767a1004728d99d9eb62759ba43f

tests/functional/openstackoperator_controller_test.go
```diff
@@ -181,6 +183,19 @@ var _ = Describe("OpenStackOperator controller", func() {
 			// make keystoneAPI ready and create secrets usually created by keystone-controller
 			keystone.SimulateKeystoneAPIReady(names.KeystoneAPIName)
 
+			// openstackversion exists
+			Eventually(func(g Gomega) {
+				osclient := GetOpenStackVersion(names.OpenStackControlplaneName)
```
gibizer, line 188: nit: osversion

tests/functional/openstackversion_controller_test.go
```diff
@@ -0,0 +1,141 @@
...
+					k8s_corev1.ConditionTrue,
+				)
+
+				version := GetOpenStackVersion(names.OpenStackVersionName)
+				g.Expect(version).Should(Not(BeNil()))
+
+				g.Expect(*version.Status.AvailableVersion).Should(Equal("0.0.1"))
+				g.Expect(version.Spec.TargetVersion).Should(Equal("0.0.1"))
+
+				g.Expect(*version.Status.ContainerImages.AgentImage).ShouldNot(BeNil())
```
gibizer, line 63: here and below: why do you take the address of the image field? It is already a *string.

controllers/core/openstackversion_controller.go
dprince: This is expected to happen in a follow up.

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	)
+	// no minor update conditions unless we have a deployed version
+	if instance.Status.DeployedVersion != nil && instance.Spec.TargetVersion != *instance.Status.DeployedVersion {
+		cl = append(cl, *condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateOVNControlplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)),
+			*condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateControlplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)))
+		// fixme add dataplane conditions here
+		//condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateOVNDataplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)),
+		//condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateDataplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)))
+	}
+	instance.Status.Conditions.Init(&cl)
```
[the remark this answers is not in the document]
dprince: I'm holding on this for now as it has been pointed out the saveReconcile logic also has some concerns. Lets treat as a follow up.

controllers/core/openstackversion_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	if isNewInstance {
+		// Register overall status immediately to have an early feedback e.g. in the cli
+		return ctrl.Result{}, nil
+	}
+
+	instance.Status.Conditions.Set(condition.FalseCondition(
```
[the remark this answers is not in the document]
dprince: agree, lets do in a follow up.

controllers/core/openstackversion_controller.go
gibizer: OK.

controllers/core/openstackversion_controller.go
dprince: lets address this in a follow up.

controllers/core/openstackversion_controller.go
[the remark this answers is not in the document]
gibizer: I agree.

controllers/core/openstackversion_controller.go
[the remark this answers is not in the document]
gibizer: OK

controllers/core/openstackversion_controller.go
dprince: same as above. There is no way to implement this in this patch (and keep minor updates passing) until we land the appropriate Dataplane code. If I were to wire this in now, minor updates would fail to work...

pkg/openstack/placement.go
dprince: I don't think we should chase these now. We can do them in a follow up

tests/functional/openstackoperator_controller_test.go
dprince: I will fix this since I have to rebase anyway.

controllers/core/openstackversion_controller.go
gibizer: OK

controllers/core/openstackversion_controller.go
gibizer: OK

pkg/openstack/nova.go
gibizer: Sure, lets do this in a follow up

pkg/openstack/placement.go
gibizer: OK

dprince:
@gibizer I answered you last round of comments. NOTE: I think some of these have come up multiple times now. To be clear my intention is only to check OVN, and Keystone conditions on this PR. I will expend the checks in a follow up PR to include all of the Controlplane Services.
Additionally, the Dataplane conditions can't be enabled now as the minor update would fail (the Dataplane PRs haven't landed as they depend on this). The plan is to land this PR first and then follow up to enabled those Dataplane conditions later.

I will fix the test nits since I have to rebase anyway. Anyone else reviewing this please consider the above ^^

tests/functional/openstackversion_controller_test.go
dprince: initially I was checking the empty string values. Since I changed to just checking for nil I can remove the dereference bits I think

controllers/core/openstackversion_controller.go
dprince: fixed in the lastest push

dprince:
The RDO failures here are due to:

''ansible.utils.unsafe_proxy.AnsibleUnsafeText object' has no attribute 'metadata'

A recent breakage outlined in: https://issues.redhat.com/browse/OSPRH-6210

It seems there is no point in rechecking until the fixes there land

dprince:
recheck

slagle (approved):

PR #722, opened 2024-03-22
Add Neutron's default route annotations

elvgarrui:
When the webhook calls for Neutron defaults, it should also write the route override annotation on Neutron's CR section so that the route gets modified.

Currently it is failing when I try to modify the openstack controlplane CR, and I don't get to see anything under neutron apiOverride on the yaml. But I do see the Annotation applied on the OverrideSpec object. I assume I'm doing something wrong here, any help appreciated!

(Goes together with https://github.com/openstack-k8s-operators/neutron-operator/pull/324)
Depends-On: https://github.com/openstack-k8s-operators/neutron-operator/pull/324

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_webhook.go
+++ b/apis/core/v1beta1/openstackcontrolplane_webhook.go
@@ -19,6 +19,7 @@ package v1beta1
 import (
 	"fmt"
 
+	"github.com/openstack-k8s-operators/lib-common/modules/common/route"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -318,6 +319,14 @@ func (r *OpenStackControlPlane) Default() {
 	r.DefaultServices()
 }
 
+// Helper function to initialize overrideSpec object. Could be moved to lib-common.
+func initializeOverrideSpec(override **route.OverrideSpec, anno map[string]string) {
+	if *override == nil {
+		*override = &route.OverrideSpec{}
+	}
+	(*override).AddAnnotation(anno)
+}
+
 // DefaultServices - common function for calling individual services' defaulting functions
 func (r *OpenStackControlPlane) DefaultServices() {
 	// RabbitMQ
@@ -381,6 +390,7 @@ func (r *OpenStackControlPlane) DefaultServices() {
 
 	// Neutron
 	r.Spec.Neutron.Template.Default()
+	initializeOverrideSpec(&r.Spec.Neutron.APIOverride.Route, r.Spec.Neutron.Template.GetDefaultRouteAnnotations())
 
 	// Nova
 	r.Spec.Nova.Template.Default()
--- a/tests/functional/openstackoperator_controller_test.go
+++ b/tests/functional/openstackoperator_controller_test.go
@@ -96,6 +96,11 @@ var _ = Describe("OpenStackOperator controller", func() {
 				g.Expect(keystoneAPI).Should(Not(BeNil()))
 			}, timeout, interval).Should(Succeed())
 		})
+		// Default route timeouts are set
+		It("should have default timeout for the routes set", func() {
+			OSCtlplane := GetOpenStackControlPlane(names.OpenStackControlplaneName)
+			Expect(OSCtlplane.Spec.Neutron.APIOverride.Route.Annotations).Should(HaveKeyWithValue("haproxy.router.openshift.io/timeout", "120s"))
+		})
 
 		It("should create selfsigned issuer and public+internal CA and issuer", func() {
 			OSCtlplane := GetOpenStackControlPlane(names.OpenStackControlplaneName)


dprince:
@elvgarrui hi, just left you a comment on the neutron-operator CR. I think you just need to move the validation into the *core* NeutronAPISpecCore.Default() validation function which currently has nothing in it. There is actually a comment in the Neutron code that says 'only container image validations go here' for the NeutronAPISpec.Default() function. I'm not sure what else we can do other than have a comment to go along with these things.

dprince:
One more comment: We actually switched to the *core* structs in OpenStackOperator yet. But soon will once the OpenStackVersion stuff lands.

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -381,6 +382,14 @@ func (r *OpenStackControlPlane) DefaultServices() {
 
 	// Neutron
 	r.Spec.Neutron.Template.Default()
+	// WIP: Error when trying to trigger webhook by modifying the CR.
+	if r.Spec.Neutron.APIOverride.Route == nil {
+		r.Spec.Neutron.APIOverride.Route = &route.OverrideSpec{}
+		openstackcontrolplanelog.Info("Neutron Route Override initialized")
+
+	}
+	r.Spec.Neutron.APIOverride.Route.AddAnnotation(r.Spec.Neutron.Template.GetDefaultRouteAnnotations())
```
booxter, line 391: I think this is the important part and you don't need a new field in neutron struct at all.

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -381,6 +382,14 @@ func (r *OpenStackControlPlane) DefaultServices() {
 
 	// Neutron
 	r.Spec.Neutron.Template.Default()
+	// WIP: Error when trying to trigger webhook by modifying the CR.
+	if r.Spec.Neutron.APIOverride.Route == nil {
```
booxter, line 386: This logic may be helpful for other services, consider introducing a helper function that would receive an override object and default annotations and that would initialize the override route accordingly.

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -381,6 +382,14 @@ func (r *OpenStackControlPlane) DefaultServices() {
...
+		r.Spec.Neutron.APIOverride.Route = &route.OverrideSpec{}
+		openstackcontrolplanelog.Info("Neutron Route Override initialized")
```
booxter, line 388: remove the logs

booxter:
It would be nice to include a functional test here confirming that the default is set for neutron as expected.

booxter (changes requested):

apis/core/v1beta1/openstackcontrolplane_webhook.go
elvgarrui: That is what I was told at first. However, I added it because before that I got "runtime error: invalid memory address or nil pointer dereference " since the r.Spec.Neutron.APIOverride.Route value there is actually null (therefore I cannot use the AddAnnotation function)

apis/core/v1beta1/openstackcontrolplane_webhook.go
elvgarrui: This should be defined in lib-common, right? Should I propose it there and then change it here on a follow-up patch? Or is there a better way?

booxter: You could have it here for now since the code that will use it will be here too, no? If it implies lib-common, definitely don't go there not to slow down merging it. 

(Other projects should probably take it from there and iterate over all service CRs to initialize defaults.)

elvgarrui: you are right, at first I thought of creating the function with a receiver, and that can only be done from the lib-common repository, but I can pass the * OverrideSpec as an argument.

elvgarrui:
pushed some of the requested changes and also rebased, but it is still missing the functional test

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -318,6 +319,13 @@ func (r *OpenStackControlPlane) Default() {
 	r.DefaultServices()
 }
 
+// Helper function to initialize overrideSpec object. Could be moved to lib-common.
+func initializeOverride(override **route.OverrideSpec) {
```
booxter, line 323: I thought it would also receive route annotations as arg, then it would

override.Route.AddAnnotations(annotations)...

otherwise this helper is very bare and probably is not very helpful

booxter:
looks fine but need a test, thanks

apis/core/v1beta1/openstackcontrolplane_webhook.go
elvgarrui: Oooh I see... I have changed it now to include that. Let me know if it looks better. Functional test was also added.

karelyatin:
recheck zuul one with Depends-On

abays:
recheck

abays:
I think this needs a `go get -u github.com/openstack-k8s-operators/neutron-operator/api`

elvgarrui:
> I think this needs a `go get -u github.com/openstack-k8s-operators/neutron-operator/api`
@abays 
Hi! I've tried running that command, and it creates no changes on the go.mod or any other file.
I've tried rebasing and running the command too but there was also no output from it. 
I'm uploading the rebased verison anyway, in case it helps.

abays:
> > I think this needs a `go get -u github.com/openstack-k8s-operators/neutron-operator/api`
> 
> @abays
> 
> Hi! I've tried running that command, and it creates no changes on the go.mod or any other file. I've tried rebasing and running the command too but there was also no output from it. I'm uploading the rebased version anyway, in case it helps.

@elvgarrui My bad, the command needs `@latest` at the end:

```
go get -u github.com/openstack-k8s-operators/neutron-operator/api@latest
```

elvgarrui:
@abays  Oooh that makes sense, thanks :)
It's updated now in both the root directory and the apis/ one. 
Hope it passes the CI this time.

abays:
@elvgarrui Bah, I'm so sorry.  I was wrong again.  I was thinking of containers, not Golang.  The proper command is:

```
go get -u github.com/openstack-k8s-operators/neutron-operator/api@main
```

elvgarrui:
@abays Nevermind! I'm really grateful for the quick help :) Re-pushed again with tags from main

PR #730, opened 2024-04-02
Create ONVDB client certificate for Octavia

gthiemonge:
The certificate is required by the Octavia OVN provider when TLS is enabled at the Pod level.

Jira: [OSPRH-6065](https://issues.redhat.com//browse/OSPRH-6065)


pkg/openstack/octavia.go
```diff
@@ -67,6 +69,36 @@ func ReconcileOctavia(ctx context.Context, instance *corev1beta1.OpenStackContro
...
 		instance.Spec.Octavia.Template.OctaviaAPI.TLS = octavia.Spec.OctaviaAPI.TLS
+
+		serviceName := "octavia"
+		// create ovndb client certificate for octavia
+		certRequest := certmanager.CertificateRequest{
+			IssuerName: instance.GetOvnIssuer(),
+			CertName:   fmt.Sprintf("%s-ovndbs", serviceName),
+			Duration:   nil,
+			Hostnames: []string{
+				fmt.Sprintf("%s.%s.svc", serviceName, instance.Namespace),
```
olliewalsh, line 80: Should this be a wildcard cert since the pod dns names are not predictable.

pkg/openstack/octavia.go
```diff
@@ -67,6 +69,36 @@ func ReconcileOctavia(ctx context.Context, instance *corev1beta1.OpenStackContro
...
+				fmt.Sprintf("%s.%s.svc.%s", serviceName, instance.Namespace, "cluster.local"),
```
olliewalsh, line 81: May be best to move `cluster.local` to a const

stuggi: I have added a const in https://github.com/openstack-k8s-operators/openstack-operator/pull/740

pkg/openstack/octavia.go
gthiemonge: I see some services like neutron not using the wildcard: https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/openstack/neutron.go#L73
is it really useful here?

pkg/openstack/octavia.go
gthiemonge: ack

weinimo:
recheck
RETRY_LIMIT

stuggi:
needs a rebase

stuggi (approved):
/lgtm

PR #742, opened 2024-04-10
Switch to tempest role for running tempest

raukadah:
openstack-k8s-operators/ci-framework#1391 moves to test_operator role for running tempest tests.
 It broke the existing vars based on tempest role. It also needs to be migrated to test_operator. 
Let use tempest role here to bring back running tempest tests.


olliewalsh (approved):
/lgtm

olliewalsh (approved):
/lgtm

PR #750, opened 2024-04-12
Save/Restore condition transition time for control plane

gibizer:
This is needed as we are resetting the condition state of each reconcile
but the changing lastTransitionTime field causes reconcile loops.


fmount (commented):
/lgtm

fmount:
Even with this patch I still see looping in my local test. I'll try a clean environment but I'm not sure it will change the current behavior

dprince:
@gibizer as discussed yesterday. Here is my PR with all the observedGeneration and savedConditions work: https://github.com/openstack-k8s-operators/openstack-operator/pull/751

gibizer:
abandoned in favor of #751

PR #760, opened 2024-04-17
Add type label to osctlplane object

stuggi:
Adds map[string]string{"core.openstack.org/openstackcontrolplane": ""} to the instance, if not already provided in the CR. With this ctlplane object can be queried using the default label. Which is useful e.g. in tests to query the ctlplane in a generic way without the object name.

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_webhook.go
+++ b/apis/core/v1beta1/openstackcontrolplane_webhook.go
@@ -18,6 +18,7 @@ package v1beta1
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/openstack-k8s-operators/lib-common/modules/common/route"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -420,3 +422,17 @@ func (r *OpenStackControlPlane) DefaultServices() {
...
+func (r *OpenStackControlPlane) DefaultLabel() {
+	// adds map[string]string{"core.openstack.org/openstackcontrolplane": r.name>} to the
+	// instance, if not already provided in the CR. With this ctlplane object can be
+	// queried using the default label.
+	typeLabel := strings.ToLower(r.GroupVersionKind().Group + "/" + r.Kind)
+	if _, ok := r.Labels[typeLabel]; !ok {
+		if r.Labels == nil {
+			r.Labels = map[string]string{}
+		}
+		r.Labels[typeLabel] = r.Name
```
stuggi, line 436: maybe don't have to set `r.Name`. when you know the name you could query it direct

stuggi:
v1.3.x PR https://github.com/openstack-k8s-operators/osp-director-operator/pull/1029

stuggi:
recheck

dprince (approved):

PR #763, opened 2024-04-18
Add finalizer on version

dprince:
Jira: [OSPRH-6344](https://issues.redhat.com//browse/OSPRH-6344)

### DIFF
--- a/controllers/core/openstackcontrolplane_controller.go
+++ b/controllers/core/openstackcontrolplane_controller.go
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
@@ -13,7 +13,6 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-
 package core
 
 import (
@@ -31,7 +30,7 @@ import (
 	ironicv1 "github.com/openstack-k8s-operators/ironic-operator/api/v1beta1"
 	keystonev1 "github.com/openstack-k8s-operators/keystone-operator/api/v1beta1"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
-	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	common_helper "github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	corev1 "k8s.io/api/core/v1"
 
 	designatev1 "github.com/openstack-k8s-operators/designate-operator/api/v1beta1"
@@ -60,6 +59,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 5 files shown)


controllers/core/openstackcontrolplane_controller.go
```diff
@@ -418,6 +441,23 @@ func (r *OpenStackControlPlaneReconciler) reconcileNormal(ctx context.Context, i
 	return ctrl.Result{}, nil
 }
 
+func (r *OpenStackControlPlaneReconciler) reconcileDelete(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, version *corev1beta1.OpenStackVersion, helper *common_helper.Helper, versionHelper *common_helper.Helper) (ctrl.Result, error) {
+	helper.GetLogger().Info("reconcile delete")
+	controllerutil.RemoveFinalizer(version, versionHelper.GetFinalizer())
```
abays, line 446: This will return `true` if it actually removes the finalizer, so we could gate on that and only call `versionHelper.PatchInstance` if needed

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -418,6 +441,23 @@ func (r *OpenStackControlPlaneReconciler) reconcileNormal(ctx context.Context, i
...
+	err := versionHelper.PatchInstance(ctx, version)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	helper.GetLogger().Info(fmt.Sprintf("finalizer removed '%s' successfully", versionHelper.GetFinalizer()))
+
+	// remove instance finalizer
+	controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
```
abays, line 454: Same thing here with checking for return of `true`

tests/functional/openstackoperator_controller_test.go
```diff
@@ -1065,6 +1065,40 @@ var _ = Describe("OpenStackOperator controller", func() {
...
+					corev1.OpenStackVersionInitialized,
+					k8s_corev1.ConditionTrue,
+				)
+			}, timeout, interval).Should(Succeed())
+
+			th.DeleteInstance(GetOpenStackControlPlane(names.OpenStackControlplaneName))
+
+			instance := &corev1.OpenStackVersion{}
+			Eventually(func(g Gomega) {
+				g.Expect(k8sClient.Get(ctx, names.OpenStackControlplaneName, instance)).Should(BeNil()) //nolint:all
```
dprince, line 1096: Trying to avoid the nolint here. I tried, Should(Not(Succeed())) but that fails which now makes me think this test isn't quite right. @gibizer if you have time take a look

controllers/core/openstackcontrolplane_controller.go
dprince: ack, good call. I will add try to add this.

controllers/core/openstackcontrolplane_controller.go
dprince: will do

gibizer: no need.

tests/functional/openstackoperator_controller_test.go
gibizer: It is a shortcoming of the envtest setup. The automatic garbage collection of k8s is not running in this env. https://book.kubebuilder.io/reference/envtest.html#testing-considerations Checking the OwnerReference and checking that the Finalizer is removed from the Version CR is what we can assert but we cannot assert that the  CR is deleted.

controllers/core/openstackcontrolplane_controller.go
gibizer: ohh sorry I mixed up. In this case we are removing the finalizer from the version object during control plane delete. It make sense to make the version object update conditional. What confused me that we have helper for both the instance (controlplane) and the version object. So here I'm OK to add something like:
```golang
if controllerutil.RemoveFinalizer(version, versionHelper.GetFinalizer()){
   err := versionHelper.PatchInstance(ctx, version)
   if err != nil {
       return ctrl.Result{}, err
    }
}
```

controllers/core/openstackcontrolplane_controller.go
gibizer: here I would just let the deferred PatchInstance does its job automatically

tests/functional/openstackoperator_controller_test.go
dprince: bummer. Thanks @gibizer This was a head scratcher. I will adjust it accordingly.

gibizer (approved):

PR #768, opened 2024-04-22
Restore lastTransitionTime of ReadyCondition

gibizer:
We placed the RestoreLastTransitionTimes before we decide if the
ReadyCondition needs to be set to True or not. This causes that the
lastTransitionTime of the ReadyCondition changes even if nothing is
actually changed during the reconciliation. The OpenStackControlPlane
and OpenStackVersion resources watching each other creating a reconcile
loop.

This change moves the RestoreLastTransitionTimes to happen after the
ReadyCondition status is finally set.


fmount (approved):
/lgtm

abays (approved):
/lgtm

PR #769, opened 2024-04-22
Add sample which only has services which are enabled per default

stuggi:
Depends-On: https://github.com/openstack-k8s-operators/ironic-operator/pull/418


stuggi:
@abays @dprince as we discussed .

abays (approved):
/lgtm

abays:
The `Depends-On` was not honored.

PR #772, opened 2024-04-24
[tls] adds conditions when custom issuer and custom service tls certs + adding envtests for scenarios

stuggi:
Add custom issuer not found condition +  adds OpenStackControlPlaneCustomTLSReadyCondition to be used when custom TLS certificates are used.

* If the custom cert secret is not available the condition is set.
* If no custom cert secret is used, the condition won't be added to the condition list.

Also adds additional tlse tests to cover following scenarios:
* TLSe OpenStackControlplane
* TLSe OpenStackControlplane with custom public issuer
* public TLS OpenStackControlplane with custom keystone public certificate

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/498


olliewalsh (approved):
/lgtm

PR #773, opened 2024-04-24
Ensure golangci-lint runs on all files

gibizer:
* removed unused func params
* ignore revive dot-imports rule on ginkgo and gomega as dot import
there is the recommended practice
* added missing error handling
* various small style fixes found by the linter


stuggi (approved):
/lgtm

PR #790, opened 2024-05-06
[TLS] Fix issuer labels when switch to custom issuer

stuggi:
Right now when first deploy with default TLS settings and later switch to a custom issuer, we'll end up with two issuers to have the CA identifier label. As a result the dataplane operator is waiting as there are more then one issuer identified.

This changes to remove the identifier label on all issuers, except the one configured for the current issuer. This allows to switch
* from default issuer to customer
* from custom issuer back to default
* and from custom issuer to a different custom issuer

Jira: [OSPRH-6746](https://issues.redhat.com//browse/OSPRH-6746)


stuggi:
recheck

Deydra71 (approved):

xek (approved):
LGTM!

PR #792, opened 2024-05-07
Enforce update order for OVN for Ctlplane/EDPM

dprince:
Enforce update order for OVN for Ctlplane/EDPM
   Jira: [OSPRH-6732](https://issues.redhat.com//browse/OSPRH-6732)

### DIFF
--- a/controllers/core/openstackcontrolplane_controller.go
+++ b/controllers/core/openstackcontrolplane_controller.go
@@ -245,7 +245,9 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 			}
 			instance.Status.DeployedOVNVersion = &version.Spec.TargetVersion
 			return ctrl.Result{}, nil
-		} else if !version.Status.Conditions.IsTrue(corev1beta1.OpenStackVersionMinorUpdateControlplane) {
+		} else if version.Status.Conditions.IsTrue(corev1beta1.OpenStackVersionMinorUpdateOVNDataplane) &&
+			!version.Status.Conditions.IsTrue(corev1beta1.OpenStackVersionMinorUpdateControlplane) {
+
 			Log.Info("Minor update on the ControlPlane")
 			ctrlResult, err := r.reconcileNormal(ctx, instance, version, helper)
 			if err != nil {
--- a/controllers/core/openstackversion_controller.go
+++ b/controllers/core/openstackversion_controller.go
@@ -56,16 +56,6 @@ func SetupVersionDefaults() {
 	envContainerImages = localVars
 }
 
-func compareStringPointers(a, b *string) bool {
-	if a == nil && b == nil {
-		return true
-	}
-	if a == nil || b == nil {
-		return false
-	}
-	return *a == *b
-}
-
 // OpenStackVersionReconciler reconciles a OpenStackVersion object
 type OpenStackVersionReconciler struct {
 	client.Client
@@ -226,7 +216,7 @@ func (r *OpenStackVersionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	// minor update for OVN Controlplane in progress
 	if instance.Status.DeployedVersion != nil && instance.Spec.TargetVersion != *instance.Status.DeployedVersion {
-		if !compareStringPointers(controlPlane.Status.ContainerImages.OvnControllerImage, instance.Status.ContainerImages.OvnControllerImage) ||
+		if !openstack.OVNControllerImageCheck(controlPlane, instance) ||
 			!controlPlane.Status.Conditions.IsTrue(corev1beta1.OpenStackControlPlaneOVNReadyCondition) {
 			instance.Status.Conditions.Set(condition.FalseCondition(
 				corev1beta1.OpenStackVersionMinorUpdateOVNControlplane,
@@ -241,9 +231,7 @@ func (r *OpenStackVersionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			corev1beta1.OpenStackVersionMinorUpdateReadyMessage)
 
 		// minor update for Controlplane in progress
-		// we only check keystone here as it will only get updated during this phase
-		// FIXME: add checks to all images on the Controlplane here once conditions and observedGeneration work are finished
-		if !compareStringPointers(controlPlane.Status.ContainerImages.KeystoneAPIImage, instance.Status.ContainerImages.KeystoneAPIImage) ||
+		if !openstack.ControlplaneContainerImageCheck(controlPlane, instance) ||
 			!controlPlane.IsReady() {
 			instance.Status.Conditions.Set(condition.FalseCondition(
 				corev1beta1.OpenStackVersionMinorUpdateControlplane,
--- a/controllers/core/openstackversion_controller.go
+++ b/controllers/core/openstackversion_controller.go
@@ -32,6 +32,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/go-logr/logr"
+	dataplanev1 "github.com/openstack-k8s-operators/dataplane-operator/api/v1beta1"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	corev1beta1 "github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1"
@@ -69,9 +70,11 @@ func (r *OpenStackVersionReconciler) GetLogger(ctx context.Context) logr.Logger 
 	return log.FromContext(ctx).WithName("Controllers").WithName("OpenStackVersion")
 }
 
-//+kubebuilder:rbac:groups=core.openstack.org,resources=openstackversions,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=core.openstack.org,resources=openstackversions/status,verbs=get;update;patch
-//+kubebuilder:rbac:groups=core.openstack.org,resources=openstackversions/finalizers,verbs=update
+// +kubebuilder:rbac:groups=core.openstack.org,resources=openstackversions,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=core.openstack.org,resources=openstackversions/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=core.openstack.org,resources=openstackversions/finalizers,verbs=update
+// +kubebuilder:rbac:groups=core.openstack.org,resources=openstackcontrolplanes,verbs=get;list;watch
+// +kubebuilder:rbac:groups=dataplane.openstack.org,resources=openstackdataplanenodesets,verbs=get;list;watch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
@@ -149,10 +152,10 @@ func (r *OpenStackVersionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	// no minor update conditions unless we have a deployed version
 	if instance.Status.DeployedVersion != nil && instance.Spec.TargetVersion != *instance.Status.DeployedVersion {
 		cl = append(cl, *condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateOVNControlplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)),
-			*condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateControlplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)))
-		// fixme add dataplane conditions here
-		//condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateOVNDataplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)),
-		//condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateDataplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)))
+			*condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateOVNDataplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)),
+			*condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateControlplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)),
+			*condition.UnknownCondition(corev1beta1.OpenStackVersionMinorUpdateDataplane, condition.InitReason, string(corev1beta1.OpenStackVersionMinorUpdateInitMessage)),
+		)
 	}
 	instance.Status.Conditions.Init(&cl)
 	instance.Status.ObservedGeneration = instance.Generation
@@ -172,13 +175,14 @@ func (r *OpenStackVersionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	instance.Status.AvailableVersion = &envAvailableVersion
 	defaults := openstack.InitializeOpenStackVersionImageDefaults(ctx, envContainerImages)
-	// store the defaults for the currently available version
 	if instance.Status.ContainerImageVersionDefaults == nil {
 		instance.Status.ContainerImageVersionDefaults = make(map[string]*corev1beta1.ContainerDefaults)
 	}
+	// store the defaults for the currently available version
 	instance.Status.ContainerImageVersionDefaults[envAvailableVersion] = defaults
 
 	// calculate the container images for the target version
+	Log.Info("Target version: ", "targetVersion", instance.Spec.TargetVersion)
 	val, ok := instance.Status.ContainerImageVersionDefaults[instance.Spec.TargetVersion]
 	if !ok {
 		Log.Info("Target version not found in defaults", "targetVersion", instance.Spec.TargetVersion)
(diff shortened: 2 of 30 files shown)


pkg/openstack/dataplane.go
```diff
@@ -0,0 +1,75 @@
...
+// returns true if OVNControllers are deployed on all nodesets
+func DataplaneNodesetsOVNControllerDeployedCheck(version *corev1beta1.OpenStackVersion, dataplaneNodesets *dataplanev1.OpenStackDataPlaneNodeSetList) bool {
+	for _, nodeset := range dataplaneNodesets.Items {
+		if nodeset.Generation != nodeset.Status.ObservedGeneration {
+			return false
+		}
+		if !nodeset.IsReady() {
+			return false
+		}
+		if nodeset.Status.ContainerImages["OvnControllerImage"] != "" && nodeset.Status.ContainerImages["OvnControllerImage"] != *version.Status.ContainerImages.OvnControllerImage {
```
abays, line 53: If `nodeset.Status.ContainerImages["OvnControllerImage"]` is current `""`, wouldn't that mean we should `return false` for `"returns true if OVNControllers are deployed on all nodesets"`?  Because if there's no image, then by inference OVNController is not deployed?

pkg/openstack/dataplane.go
```diff
@@ -0,0 +1,75 @@
...
+		// FIXME: DeployedVersion on the DataplaneNodeset should be a string pointer to match how Controlplane implements this
+		if nodeset.Status.DeployedVersion == "" {
+			return false
+		}
+	}
+	return true
+}
+
+//if !openstack.OVNControllerImageCheck(controlPlane, instance) ||
+//!controlPlane.Status.Conditions.IsTrue(corev1beta1.OpenStackControlPlaneOVNReadyCondition)
```
dprince, line 42: FIXME: remove this.

pkg/openstack/dataplane.go
dprince: so the logic here is different than everywhere else in openstack-operator because dataplane-operator is using strings instead of *string for its containerImage parameters. But my intent here was to ensure it was set to something before testing it. Also, as far as I know it would be possible to have a DataplaneNodeSet that doesn't deploy a specific service and thus we may need a fall through for that case so the minor update could proceed.

pkg/openstack/version.go
```diff
@@ -202,3 +202,47 @@ func ReconcileVersion(ctx context.Context, instance *corev1beta1.OpenStackContro
 
 	return ctrl.Result{}, version, nil
 }
+
+func compareStringPointers(a, b *string) bool {
```
gibizer, line 206: I would still rename this accoridng to https://github.com/openstack-k8s-operators/openstack-operator/pull/712/files#r1528375598

controllers/core/openstackversion_controller.go
```diff
@@ -240,10 +238,22 @@ func (r *OpenStackVersionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+		}
+		instance.Status.Conditions.MarkTrue(
+			corev1beta1.OpenStackVersionMinorUpdateOVNDataplane,
+			corev1beta1.OpenStackVersionMinorUpdateReadyMessage)
+
 		// minor update for Controlplane in progress
-		// we only check keystone here as it will only get updated during this phase
-		// FIXME: add checks to all images on the Controlplane here once conditions and observedGeneration work are finished
-		if !compareStringPointers(controlPlane.Status.ContainerImages.KeystoneAPIImage, instance.Status.ContainerImages.KeystoneAPIImage) ||
+		if !openstack.ControlplaneContainerImageCheck(controlPlane, instance) ||
```
gibizer, line 256: This would read a lot better if the name of the *Check functions suggest when the function returns true. E.g.
```golang
		if !openstack.ControlplaneContainerImageMatch(controlPlane, instance) ||
```

controllers/core/openstackversion_controller.go
```diff
@@ -216,17 +208,23 @@ func (r *OpenStackVersionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 	}
 
-	// TODO minor update for OVN Dataplane in progress
+	// greenfield deployment
+	if controlPlane.Status.DeployedVersion == nil && !openstack.DataplaneNodesetsDeployedVersionIsSet(dataplaneNodesets) {
+		Log.Info("Waiting for controlplane and dataplane nodesets to be deployed.")
+		return ctrl.Result{}, nil
+	}
 
 	// minor update for OVN Controlplane in progress
```
gibizer, line 224: nit: I think this comment more belongs to the next if now at L227, or need to be reworded that "minore update in progress" as the current condition only checks that.

controllers/core/openstackversion_controller.go
```diff
@@ -240,10 +238,22 @@ func (r *OpenStackVersionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			corev1beta1.OpenStackVersionMinorUpdateOVNControlplane,
 			corev1beta1.OpenStackVersionMinorUpdateReadyMessage)
 
+		// minor update for Dataplane OVN
+		if !openstack.DataplaneNodesetsOVNControllerDeployedCheck(instance, dataplaneNodesets) {
```
gibizer, line 242: here I would simply drop the Check from the end of the function name, as it will automatically suggest that it is true when OVN is deployed.

gibizer, line 242: The current logic in DataplaneNodesetsOVNControllerDeployedCheck will return false if there are NodeSets that are not ready even if that nodeset does not need to deploy OVNController. So I think we need to change the logic there to ignore these nodeset

pkg/openstack/dataplane.go
gibizer: > Also, as far as I know it would be possible to have a DataplaneNodeSet that doesn't deploy a specific service and thus we may need a fall through for that case so the minor update could proceed.

If there are NodeSets without OVN controller, then I think we don't need to wait for the readyness of those NodeSets either.

controllers/core/openstackversion_controller.go
```diff
@@ -253,7 +263,17 @@ func (r *OpenStackVersionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			Log.Info("Minor update for Controlplane in progress")
 			return ctrl.Result{}, nil
 		}
+
 		// TODO minor update for Dataplane in progress goes here
+		if !openstack.DataplaneNodesetsDeployedCheck(instance, dataplaneNodesets) {
```
gibizer, line 268: ditto I would drop the Check from end of the function name

pkg/openstack/keystone.go
```diff
@@ -147,3 +147,15 @@ func ReconcileKeystoneAPI(ctx context.Context, instance *corev1beta1.OpenStackCo
 
 	return ctrl.Result{}, nil
 }
+
+// KeystoneImageCheck - return true if the keystone images match on the ControlPlane and Version, or if Keystone is not enabled
+func KeystoneImageCheck(controlPlane *corev1beta1.OpenStackControlPlane, version *corev1beta1.OpenStackVersion) bool {
+
+	if controlPlane.Spec.Keystone.Enabled {
```
gibizer, line 154: Did we missed handling of Keystione.Enabled == false in the ReconcileKeystoneAPI above?

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -245,7 +245,7 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 			}
 			instance.Status.DeployedOVNVersion = &version.Spec.TargetVersion
 			return ctrl.Result{}, nil
-		} else if !version.Status.Conditions.IsTrue(corev1beta1.OpenStackVersionMinorUpdateControlplane) {
+		} else if version.Status.Conditions.IsTrue(corev1beta1.OpenStackVersionMinorUpdateOVNDataplane) && !version.Status.Conditions.IsTrue(corev1beta1.OpenStackVersionMinorUpdateControlplane) {
```
gibizer, line 248: this is a very long line now, could you fold it please

controllers/core/openstackversion_controller.go
```diff
@@ -240,10 +238,22 @@ func (r *OpenStackVersionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 			corev1beta1.OpenStackVersionMinorUpdateReadyMessage)
 
+		// minor update for Dataplane OVN
+		if !openstack.DataplaneNodesetsOVNControllerDeployedCheck(instance, dataplaneNodesets) {
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				corev1beta1.OpenStackVersionMinorUpdateOVNDataplane,
+				condition.RequestedReason,
+				condition.SeverityInfo,
+				corev1beta1.OpenStackVersionMinorUpdateReadyRunningMessage))
+			Log.Info("Minor update for OVN Dataplane in progress")
```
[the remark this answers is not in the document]
slagle: This also might not be accurate in that the dataplane deployment may not have started.

controllers/core/openstackversion_controller.go
```diff
@@ -253,7 +263,17 @@ func (r *OpenStackVersionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 		}
+
 		// TODO minor update for Dataplane in progress goes here
+		if !openstack.DataplaneNodesetsDeployedCheck(instance, dataplaneNodesets) {
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				corev1beta1.OpenStackVersionMinorUpdateDataplane,
+				condition.RequestedReason,
+				condition.SeverityInfo,
+				corev1beta1.OpenStackVersionMinorUpdateReadyRunningMessage))
+			Log.Info("Minor update for Dataplane in progress")
```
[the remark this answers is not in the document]
slagle: again, it might not be in progress yet as it needs to be manually triggered.

pkg/openstack/dataplane.go
dprince: should we label NodeSets with OVN then so we can more easily code this. I guess I can sort through them all here just as well but a label might be slightly cleaner.

controllers/core/openstackversion_controller.go
```diff
@@ -253,7 +263,17 @@ func (r *OpenStackVersionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			Log.Info("Minor update for Controlplane in progress")
 			return ctrl.Result{}, nil
 		}
+
 		// TODO minor update for Dataplane in progress goes here
```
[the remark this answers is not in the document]
dprince: got it, thanks

pkg/openstack/dataplane.go
```diff
@@ -0,0 +1,75 @@
+package openstack
+
+import (
+	"context"
+
+	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	corev1beta1 "github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1"
+
+	//metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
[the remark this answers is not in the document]
dprince: got it, thanks

pkg/openstack/keystone.go
dprince: Yes, I will add the missing 'instance.Status.ContainerImages.KeystoneAPIImage = nil' to that block

controllers/core/openstackcontrolplane_controller.go
dprince: got it.

pkg/openstack/version.go
```diff
@@ -202,3 +202,47 @@ func ReconcileVersion(ctx context.Context, instance *corev1beta1.OpenStackContro
...
+	}
+	if a == nil || b == nil {
+		return false
+	}
+	return *a == *b
+}
+
+// ControlplaneContainerImageCheck - function to compare the ContainerImages on the controlPlane to the OpenStackVersion
+// only enabled services are checked
+func ControlplaneContainerImageCheck(controlPlane *corev1beta1.OpenStackControlPlane, version *corev1beta1.OpenStackVersion) bool {
```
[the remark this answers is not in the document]
dprince: done

pkg/openstack/dataplane.go
```diff
@@ -0,0 +1,75 @@
...
+	for _, nodeset := range dataplaneNodesets.Items {
+		if nodeset.Generation != nodeset.Status.ObservedGeneration {
+			return false
+		}
+		if !nodeset.IsReady() {
+			return false
+		}
+		if nodeset.Status.DeployedVersion != version.Spec.TargetVersion {
+			return false
+		}
```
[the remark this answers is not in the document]
dprince: actually, the minor update can include containers other than OVN for this final step. I think we need to wait for the DeployedVersion to get set on all NodeSets here regardless of whether they have OVN install. I did correct the logic in the function above on line 49-50 so I think we are good here.

controllers/core/openstackversion_controller.go
dprince: done

pkg/openstack/dataplane.go
```diff
@@ -0,0 +1,75 @@
...
+	//metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	//"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	//"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	dataplanev1 "github.com/openstack-k8s-operators/dataplane-operator/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetDataplaneNodesets - returns the dataplanenodesets in the namespace of the controlplane
+func GetDataplaneNodesets(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, helper *helper.Helper) (ctrl.Result, *dataplanev1.OpenStackDataPlaneNodeSetList, error) {
```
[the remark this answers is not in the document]
dprince: fixed.

pkg/openstack/dataplane.go
dprince: good catches here. I have adjusted the logic. See what you think now

controllers/core/openstackversion_controller.go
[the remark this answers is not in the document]
gibizer: I still see this message in the current version. Do you plan to resolve this in a follow up? (That would be also OK to me)

controllers/core/openstackversion_controller.go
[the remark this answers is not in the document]
gibizer: same as above.

pkg/openstack/dataplane.go
gibizer: looks good to me

dprince:
recheck

dprince:
recheck

tests/functional/base_test.go
```diff
@@ -274,6 +278,55 @@ func CreateOpenStackControlPlane(name types.NamespacedName, spec map[string]inte
 	return th.CreateUnstructured(raw)
 }
 
+// Build OpenStackDataPlaneNodeSetSpec struct with empty `Nodes` list
+func DefaultDataPlaneNoNodeSetSpec(tlsEnabled bool) map[string]interface{} {
```
dprince, line 282: @fao89 hey, just a heads up on the dataplane-operator merge I needed to forklift a few functions here to implement some functional tests for the minor updates workflow.

dprince:
recheck

tests/functional/base_test.go
```diff
@@ -453,3 +506,71 @@ func CreateClusterConfigCM() client.Object {
...
+	}
+
+	if instance.Spec.Memcached.Enabled {
+		if instance.Spec.TLS.PodLevel.Enabled {
+			infra.SimulateTLSMemcachedReady(names.MemcachedName)
+		} else {
+			infra.SimulateMemcachedReady(names.MemcachedName)
+
+		}
+		// FIXME memcached helper to set ObservedGeneration https://github.com/openstack-k8s-operators/infra-operator/pull/228
```
stuggi, line 541: that PR is merged

controllers/core/openstackversion_controller.go
dprince: got it fixed

pkg/openstack/version.go
dprince: I added a commit for this into the PR

tests/functional/base_test.go
dprince: got it fixed now

slawqo (commented):
looks good for me, the only thing I am not sure if I understand correctly is how all those changes in various {service}.go files are related to the PR description :)

dprince:
> looks good for me, the only thing I am not sure if I understand correctly is how all those changes in various {service}.go files are related to the PR description :)

in order to ensure update order we need to make sure services have been deployed (ready state, and observed generation checks).

gibizer (commented):
Looks OK to me.

dprince:
recheck

booxter:
I haven't tried to run it, but looking at the code, the order of updates looks correct.

I couldn't figure out from just reading the code how we guarantee that only `ovn-controller` (and not e.g. `nova-compute`) is updated on DP, before CP `reconcileNormal` is triggered. Will it be managed using a separate Deployment that would run a single `ovn` DP service? (I assume updating `nova-compute` before CP nova services is a problem.) Perhaps someone could ELI5 to me. (Thank you.)

bshephar:
> I haven't tried to run it, but looking at the code, the order of updates looks correct.
> 
> I couldn't figure out from just reading the code how we guarantee that only `ovn-controller` (and not e.g. `nova-compute`) is updated on DP, before CP `reconcileNormal` is triggered. Will it be managed using a separate Deployment that would run a single `ovn` DP service? (I assume updating `nova-compute` before CP nova services is a problem.) Perhaps someone could ELI5 to me. (Thank you.)

Also trying to follow this. So we set the condition here:
https://github.com/openstack-k8s-operators/openstack-operator/pull/792/files#diff-32500fc60d27debdcd1f64468b83c0a318d79fa55591b2e17a5cb935d9fde650R248

But that just prevents the controller from continuing the update until the condition has been satisfied. It seems that the actual update of OVN on the Dataplane would need to be a manual process as described here:

https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/docs/assemblies/proc_updating-the-data-plane-ovn.adoc


So this would happen first and then pause until the condition get satisfied. To satisfy the condition, the image deployed would need to match the image the update is expecting as determined by:
https://github.com/openstack-k8s-operators/openstack-operator/pull/792/files#diff-308714db370a47145837acae5ff60352d9f352513007441b8c694f2c45c1031dR38-R53

Then the update can continue.

At least that's my 10 minute read on what's happening here.

dprince:
Dataplane updates are manual for GA. There are some Jira's filed (https://issues.redhat.com/browse/OSPRH-6421) which might help us fully streamline the minor update workflow.

dprince:
I do think we could also validate based on conditions set on the OpenStackVersion resource that when Dataplane resources get executed we are in the correct state. So for example if we need to just execute an OVN playbook we could have a crude validation on that. The adminstrator could always override this with Ansible, but I think a simple check like this could help us further guard the workflow in the future.

booxter:
Thank you @bshephar @dprince this (rolling in Deployment just for `ovn` service) makes sense. It's a bit more leg work for a user but the main point is we have a plan.

stuggi (approved):
/lgtm

PR #793, opened 2024-05-09
[ovndbs] For cert setup do not wait for dbs to be ready

karelyatin:
Currently when TLS is enabled OVNDBClusters are created first without Certs, wait for those to be ready, create certs and update ovndbclusters with tls enabled.

This causes issues as with this even with TLS enabled OVN raft is setup with local-address as non-ssl and later scale up do not work due to this. To fix this changing the conditions to create certs when TLS is enabled like other services.

Related-Issue: [OSPRH-6899](https://issues.redhat.com//browse/OSPRH-6899)


olliewalsh (approved):
/lgtm

booxter:
Thank you, I am late but this LGTM.

PR #796, opened 2024-05-13
[pre-commit][kuttl]Check for multiple TestAsserts

gibizer:
Add a new pre-commit check to avoid using more than one TestAssert in a
single kuttl assert file as that can lead to false positives as only the
last TestAssert is run by kuttl.

Fortunately there was no such issue in this repo.


abays (approved):
/lgtm

PR #797, opened 2024-05-14
[tls] Improve the TLS kuttl scenarios

Deydra71:
Test scenarios:

* **Custom Issuer**
  * Deploy with the default internal issuer, then switch to a custom internal issuer (and vice versa),
  * Deploy with the default ingress issuer, then switch to a custom ingress issuer  (and vice versa),
* **Service Certificate Rotation**
  * After deployment, certificate secrets are deleted, triggering service restarts. The test checks whether new secrets are mounted in service pods and if route configurations use these new secrets. 
* **Custom CA cert added to a bundle**
* **Customize cert and CA cert duration parameters**

Test scenarios to be added:
~~CA cert rotation~~


Deydra71:
The kuttl test is failing because Swift remains in a NotReady condition, even though all pods have been successfully restarted with new secrets. In the local testing environment, simply deleting the Swift pods was enough to trigger a change to the ready condition. This issue has already been discussed with @stuggi, and it is on the  todo list after the freeze.

Deydra71:
The scripts are now using cert-manager label `controller.cert-manager.io/fao=true` to delete service and route secrets. The `core.openstack.org/openstackcontrolplane` is now used to query the ctlplane without specifying the object name.

Deydra71:
recheck

stuggi (commented):
still need to run more tests. also I have found an issue with dataplane tests where it creates/reads/delete `test123.data` if the user which runs the kuttl jobs won't have permissions to write to the directory, it fails. we should probably write that file to `/tmp` . an example is https://github.com/openstack-k8s-operators/openstack-operator/blob/main/tests/kuttl/tests/dataplane-deploy-tls-test/02-assert.yaml#L50-L66 , but its used in multiple other kuttl files

morucci:
Hi, could you please remove the depends-on on the openshift/release repo  as it is not part of the Zuul/SF github app and we are experiencing issues with the github rate limiting. Perhaps refer to that reference using something else than "depends-on".

tests/kuttl/tests/ctlplane-tls-custom-issuers/11-cleanup.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+delete:
+- apiVersion: core.openstack.org/v1beta1
+  kind: OpenStackControlPlane
+  name: openstack
+commands:
+- script: |
+    oc delete --ignore-not-found=true -n $NAMESPACE pvc \
+    srv-swift-storage-0
```
stuggi, line 10: have to cleanup the custom issuers:
`oc delete --ignore-not-found=true -n $NAMESPACE issuer rootca-internal-custom rootca-ingress-custom`

tests/kuttl/common/osp_check_cert_issuer.sh
```diff
@@ -0,0 +1,81 @@
...
+function extract_host_port {
+    local endpoint_url=$1
+    local host_port
+
+    if [[ "$ENDPOINT_TYPE" == "public" ]]; then
+        # Extract the hostname and port for public endpoints
+        host_port=$(echo "$endpoint_url" | sed -E 's|^[^:/]+://([^:/]+).*|\1|')
+    else
+        # Extract the hostname and port for internal endpoints, keeping the port if specified
+        host_port=$(echo "$endpoint_url" | sed -E 's|^[^:/]+://([^:/]+(:[0-9]+)?).*|\1|')
```
olliewalsh, line 19: nit: could just use this for both endpoint types

stuggi (approved):
/lgtm

PR #799, opened 2024-05-15
Update octavia to include availabilityZone changes

karelyatin:
Required-By: https://github.com/openstack-k8s-operators/architecture/pull/227
Related-Issue: [OSPRH-6901](https://issues.redhat.com//browse/OSPRH-6901)

### DIFF
--- a/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
+++ b/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
@@ -10381,6 +10381,10 @@ spec:
                           manageLbMgmtNetworks: true
                           subnetIpVersion: 4
                         properties:
+                          availabilityZones:
+                            items:
+                              type: string
+                            type: array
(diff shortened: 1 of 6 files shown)


karelyatin:
recheck baremetal

abays (approved):
Approved beta blocker

/lgtm

abays:
Since all the individual Zuul jobs passed between the past two batch attempts, I would be okay with overriding the check here

abays:
recheck

Will try once more

karelyatin:
> Since all the individual Zuul jobs passed between the past two batch attempts, I would be okay with overriding the check here

Yeap should be fine to override, will also unblock some testing

karelyatin:
> @karelyatin: karelyatin unauthorized: /override is restricted to Repo administrators, approvers in top level OWNERS file, and the following github teams:.

So some else with permissions need do it

PR #801, opened 2024-05-16
Generate RabbitMQ TLS config that works with FIPS

dciabrin:
When TLS is enabled in RabbitMQ with FIPS engaged, the RabbitMQ server has to be given specific TLS options in the RabbitMQ configuration otherwise the rabbitmq Erlang application will fail to create TLS sockets with:

  {insufficient_crypto_support
     {'tlsv1.3',{versions,['tlsv1.3','tlsv1.2']}}}}}}}}}

We now reuse the same config as we had in previous versions of Openstack as they were known to work with FIPS. Moreover the configuration has to be passed as an Erlang structure to ensure the right orderning of option ssl_options.versions.

Related: [OSPRH-6889](https://issues.redhat.com//browse/OSPRH-6889)

### DIFF
--- a/pkg/openstack/rabbitmq.go
+++ b/pkg/openstack/rabbitmq.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 
+	certmgrv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	networkv1 "github.com/openstack-k8s-operators/infra-operator/apis/network/v1beta1"
 	"github.com/openstack-k8s-operators/lib-common/modules/certmanager"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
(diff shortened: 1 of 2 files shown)


pkg/openstack/rabbitmq.go
```diff
@@ -195,15 +196,28 @@ func reconcileRabbitMQ(
...
 		certRequest := certmanager.CertificateRequest{
 			IssuerName: instance.GetInternalIssuer(),
 			CertName:   fmt.Sprintf("%s-svc", rabbitmq.Name),
+			CommonName: &commonName,
 			Hostnames: []string{
 				hostname,
 				fmt.Sprintf("%s.%s", hostname, ClusterInternalDomain),
 			},
+			Subject: &certmgrv1.X509Subject{
+				Organizations: []string{fmt.Sprintf("%s.cluster.local", rabbitmq.Namespace)},
```
stuggi, line 211: `cluster.local` can be replaced with the const `ClusterInternalDomain`, like used above in L199

dciabrin: ack, done

pkg/openstack/rabbitmq.go
```diff
@@ -268,13 +282,60 @@ func reconcileRabbitMQ(
 			// This is the same situation as RABBITMQ_UPGRADE_LOG above,
 			// except for the "main" rabbitmq log we can just force it to use the console.
 			rabbitmq.Spec.Rabbitmq.AdditionalConfig = "log.console = true"
+			if tlsCert != "" {
+				rabbitmq.Spec.Rabbitmq.AdditionalConfig += "\nssl_options.verify = verify_none"
+			}
```
stuggi, line 287: nit, could make it a bit nicer using
```
var settings []string
settings = append(settings, "log.console = true")
if tlsCert != "" {
    settings = "ssl_options.verify = verify_none"
}
rabbitmq.Spec.Rabbitmq.AdditionalConfig = strings.Join(settings, "\n")
```

stuggi (commented):
should we remove draft so that all jobs are running?

pkg/openstack/rabbitmq.go
dciabrin: ack, slice done in the next revision.

stuggi (approved):
/lgtm

PR #802, opened 2024-05-21
Enable TLS for glanceAPI Edge instances

fmount:
When a `glanceAPI` `Edge` instance is deployed, the `glance-operator` creates an `internal` `SVC`, but the `public` one is skipped (because these instances are not exposed outside).
However, we're currently not able to enable `TLS` because the `EnsureEndpoint` function is skipped due to the different size of the generated `Overrides` that is compared with the service list (`svcs.Items`). This patch fixes the current behavior and generates an internal override only for an `Edge API` instance type: by doing this we're able to properly execute the `EnsureEndpoint` function and generate both key and cert, referenced by the `Edge API`.

Depends-On: https://github.com/openstack-k8s-operators/glance-operator/pull/535
Jira: [OSPRH-6720](https://issues.redhat.com/browse/OSPRH-6720)

### DIFF
--- a/pkg/openstack/glance.go
+++ b/pkg/openstack/glance.go
@@ -55,7 +55,12 @@ func ReconcileGlance(ctx context.Context, instance *corev1beta1.OpenStackControl
 
 	// add selector to service overrides
 	for name, glanceAPI := range instance.Spec.Glance.Template.GlanceAPIs {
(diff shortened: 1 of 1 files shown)


stuggi (approved):
/lgtm

PR #803, opened 2024-05-23
Bump dataplane-operator api

raukadah:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/889 Get AEE runner from OpenstackVersion. In order to use this feature in ci-framework. We need to bump the dataplane-operator version.

This will be used in update_containers role to update the ansibleee image.


fao89:
https://github.com/openstack-k8s-operators/openstack-operator/pull/805 already got merged

PR #813, opened 2024-05-28
Fix apache image tag

fao89:
current-podified tag does not exist


abays:
What builds `quay.io/openstack-k8s-operators/openstack-ansibleee-runner`?  Should we have a GitHub action on some repo here that pushes to that and tags `current-podified` when it does so?

fao89:
> What builds `quay.io/openstack-k8s-operators/openstack-ansibleee-runner`? Should we have a GitHub action on some repo here that pushes to that and tags `current-podified` when it does so?

https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/.github/workflows/openstack-ansibleee-runner.yaml
I'll open a PR there

fao89:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/660

dprince (approved):

PR #815, opened 2024-05-28
Ensure default issuer already exists before running custom issuer tests

olliewalsh:


gibizer (commented):
Thanks @olliewalsh the fix looks good to me.

PR #817, opened 2024-05-29
Additional SubjectName in the CertificateRequest

fmount:
This patch adds a way to check the appropriate annotation within a `SVC` and look for additional `SNs` that should be added to the `CertificateRequest`. `Glance` needs this mechanism because for each `SVC` (`public`, `internal`), we have an associated `headless service`, and it is used to resolve each replica Pod `worker_self_reference_url`. 
This allows to proxy `Pod2Pod` requests via `HTTPS`.

Jira: https://issues.redhat.com/browse/OSPRH-7276

### DIFF
--- a/pkg/openstack/common.go
+++ b/pkg/openstack/common.go
@@ -298,6 +298,11 @@ func EnsureEndpointConfig(
 						Labels:      util.MergeMaps(ed.Labels, map[string]string{serviceCertSelector: ""}),
 						Usages:      nil,
 					}
+
+					addSubjNames := util.GetStringListFromMap(svc.Annotations, tls.AdditionalSubjectNamesKey)
+					if len(addSubjNames) > 0 {
+						certRequest.Hostnames = append(certRequest.Hostnames, addSubjNames...)
+					}
 					if instance.Spec.TLS.Ingress.Cert.Duration != nil {
 						certRequest.Duration = &instance.Spec.TLS.Ingress.Cert.Duration.Duration
 					}
@@ -342,6 +347,11 @@ func EnsureEndpointConfig(
 					Annotations: ed.Annotations,
 					Labels:      util.MergeMaps(ed.Labels, map[string]string{serviceCertSelector: ""}),
 					Usages:      nil,
+				}
+
+				addSubjNames := util.GetStringListFromMap(svc.Annotations, tls.AdditionalSubjectNamesKey)
+				if len(addSubjNames) > 0 {
+					certRequest.Hostnames = append(certRequest.Hostnames, addSubjNames...)
 				}
 				if instance.Spec.TLS.PodLevel.Internal.Cert.Duration != nil {
 					certRequest.Duration = &instance.Spec.TLS.PodLevel.Internal.Cert.Duration.Duration


pkg/openstack/common.go
```diff
@@ -298,6 +298,13 @@ func EnsureEndpointConfig(
 						Labels:      util.MergeMaps(ed.Labels, map[string]string{serviceCertSelector: ""}),
 						Usages:      nil,
 					}
+					if svcName, ok := svc.Annotations["headless"]; ok {
```
fmount, line 301: @stuggi this is an example of what I need for glance: if the annotation is not present everything will work as usual, but we can embed in a service the information about additional SNs that we want in the `CertificateRequest`.
I used `headless` to give you the idea of what I need to achieve, but I'm ok to have a different name or a different structure to get additional information from the `SVC` we're iterating here.

In addition, I'd like to minimize the amount of changes, and extending the function signature requires an update of the function call everywhere, so I think we can embed this information to the SVC, which is the source of truth to compute the certificate SubjectName(s).

stuggi: - should it be possible to use the annotation to be a list of hostnames? 
```
annoations:
  additionalSubjectNames:
    - first
    - second
    - third
```
- is it ok to expect those should always be added with:
```
string(fmt.Sprintf("*.%s.%s.svc", svcName, instance.Namespace)),
string(fmt.Sprintf("*.%s.%s.svc.%s", svcName, instance.Namespace,ClusterInternalDomain)),
```
or might there be the need to register any dns name. if so, should the `additionalSubjectNames` provide any names to register and we do not generate the fqdn here?

fmount: Yes, I created a patch for lib-common where I can process a comma separated list of `additionalSubjectNames` set in the svc as an annotation from the service operator [1].
The idea is to remove:

```
string(fmt.Sprintf("*.%s.%s.svc", svcName, instance.Namespace)),
string(fmt.Sprintf("*.%s.%s.svc.%s", svcName, instance.Namespace,ClusterInternalDomain)),
```
because this detail can live in the annotation set by the service operator. Here we should just get the `[]string{}` of `additionalSubjectNames` and append them to the `CertificateRequest` field.

[1] https://github.com/openstack-k8s-operators/lib-common/pull/516

fmount: I'll follow up in a bit here.

stuggi:
we should add a new section here https://github.com/openstack-k8s-operators/dev-docs/blob/main/tls.md to document which annotations the service operators can add to get those names added to the cert.

fmount:
> we should add a new section here https://github.com/openstack-k8s-operators/dev-docs/blob/main/tls.md to document which annotations the service operators can add to get those names added to the cert.

Yes, let me draft something and I'll propose a patch for `dev-docs` as well! Thank you, good catch!

fmount:
> > we should add a new section here https://github.com/openstack-k8s-operators/dev-docs/blob/main/tls.md to document which annotations the service operators can add to get those names added to the cert.
> 
> Yes, let me draft something and I'll propose a patch for `dev-docs` as well! Thank you, good catch!

@stuggi FYI https://github.com/openstack-k8s-operators/dev-docs/pull/115 is supposed to cover this topic

fmount:
@stuggi we good with this?

stuggi (approved):
/lgtm

PR #819, opened 2024-05-30
Enable HSTS on OCP route

bshephar:
This change adds the required annotation to the OCP route to enable HSTS headers as per:
https://access.redhat.com/documentation/en-us/openshift_container_platform/4.7/html/networking/configuring-routes\#nw-enabling-hsts_route-configuration

Resolves: https://issues.redhat.com/browse/OSPRH-7367

### DIFF
--- a/pkg/openstack/horizon.go
+++ b/pkg/openstack/horizon.go
@@ -22,6 +22,11 @@ import (
 
 // ReconcileHorizon -
 func ReconcileHorizon(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, version *corev1beta1.OpenStackVersion, helper *helper.Helper) (ctrl.Result, error) {
+	const (
+		HorizonHSTSHeaderAnnotation      string = "haproxy.router.openshift.io/hsts_header"
+		HorizonHSTSHeaderAnnotationValue string = "max-age=31536000;includeSubDomains;preload"
+	)
+
 	horizon := &horizonv1.Horizon{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "horizon",
@@ -73,6 +78,14 @@ func ReconcileHorizon(ctx context.Context, instance *corev1beta1.OpenStackContro
 		return ctrl.Result{}, err
 	}
 
+	// Set HSTS Headers for Horizon Route
+	//
+	apiOverrides := instance.Spec.Horizon.APIOverride
+
+	if _, ok := apiOverrides.Route.Annotations[HorizonHSTSHeaderAnnotation]; !ok {
+		apiOverrides.Route.Annotations[HorizonHSTSHeaderAnnotation] = HorizonHSTSHeaderAnnotationValue
+	}
+
 	// make sure to get to EndpointConfig when all service got created
 	if len(svcs.Items) == 1 {
 		endpointDetails, ctrlResult, err := EnsureEndpointConfig(
@@ -82,7 +95,7 @@ func ReconcileHorizon(ctx context.Context, instance *corev1beta1.OpenStackContro
 			horizon,
 			svcs,
 			serviceOverrides,
-			instance.Spec.Horizon.APIOverride,
+			apiOverrides,
 			corev1beta1.OpenStackControlPlaneExposeHorizonReadyCondition,
 			false, // TODO (mschuppert) could be removed when all integrated service support TLS
 			tls.API{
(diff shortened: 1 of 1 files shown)


bshephar:
I'm not a massive fan of updating Spec sections outside of webhooks. But given that we don't have anything else using webhooks to modify the OSCP, I'm inclined to avoid additional webhook logic just for this use case. Happy to take alternative suggestions on putting this elsewhere? Else, we can stick with it for now and move it later.

pkg/openstack/horizon.go
```diff
@@ -73,6 +82,9 @@ func ReconcileHorizon(ctx context.Context, instance *corev1beta1.OpenStackContro
 		return ctrl.Result{}, err
 	}
 
+	// Set HSTS Headers for Horizon Route
+	instance.Spec.Horizon.APIOverride.Route.Annotations = HSTSAnnotation
```
gibizer, line 86: * I suggest to make this additive to any user passed annotation in instance.Spec.Horizon.APIOverride.Route.Annotations. You can user utils.MergeMaps to add your annotation to the existing ones.

* I'm a bit reluctant to modify instance.Spec directly as this is misleading. The controller should never modify its own Spec and I think to avoid that we explicitly ignore spec changes when persisting the instance at the end of the reconcile.

  To make it less confusing you could take a deep copy instance.Spec.Horizon.APIOverride, modify the copy and then pass that copy to EnsureEndpointConfig.

  Alternatively you can modify the result of EnsureEndpointConfig to add your annotations in `endpointDetails.EndpointDetails[service.EndpointPublic].Route.OverrideSpec.Annotations`

gibizer:
Yeah, I don't see another way to achieve this today.

There is a possible direction in the future: The horizon-operator adds annotation to the service https://github.com/openstack-k8s-operators/horizon-operator/blob/9d2609bb0015818c7eb07a207f5942817c2273c3/controllers/horizon_controller.go#L4[…]412 and the openstack-operator reads them to make decision about the route creation https://github.com/openstack-k8s-operators/openstack-operator/blob/aa91068ef71b267cc9a77cb8950d670a2e292d7b/pkg/openstack/common.go#L232-L233  So I think this channel could be extended to pass extra annotations up to the route created. 
But I hope @stuggi  knows more.

pkg/openstack/horizon.go
bshephar: True, we don't need to actually modify the spec. We can just create a copy of the user provided values and add our own. I've pushed a new revision using that method.

gibizer: could you also merge the new annotations with any existing one coming from the user?

bshephar: It wouldn't need to in this case right? It's just a `map[string]string`, so in the latest revision:
```
	// Set HSTS Headers for Horizon Route
	apiOverrides := instance.Spec.Horizon.APIOverride
	apiOverrides.Route.Annotations[HorizonHSTSHeaderAnnotation] = HorizonHSTSHeaderAnnotationValue

```

We copy any of the user provided values when we duplicate the struct to our `apiOverrides` variable, then we just insert our annotation in addition to whatever already exists in the map.

Maybe the user provided something and the map looks like:
```
map[string]string{
  "this-blah-named-annotation": "this-blah-value",
  "more-blah-annotations": "more-blah-values",
  }
```

Then I come along and make a full copy of that and just add my own in, so now it looks like:
```
map[string]string{
  "this-blah-named-annotation": "this-blah-value",
  "more-blah-annotations": "more-blah-values",
  "haproxy.router.openshift.io/hsts_header": "max-age=31536000;includeSubDomains;preload",
  }
```

Or are you suggesting that the user might also provide a annotation for `"haproxy.router.openshift.io/hsts_header"` and I should merge their provided value and ours?

Maybe if that's the case, then I should just do "if this key does not exists, add annotation."

bshephar: Latest revision looks like this:
```
	// Set HSTS Headers for Horizon Route
	//
	apiOverrides := instance.Spec.Horizon.APIOverride

	if _, ok := apiOverrides.Route.Annotations[HorizonHSTSHeaderAnnotation]; !ok {
		apiOverrides.Route.Annotations[HorizonHSTSHeaderAnnotation] = HorizonHSTSHeaderAnnotationValue
	}

```

gibizer: ahh you are right, I looked at the wrong version. You are clearly adding to it not overriding it. Thanks

gibizer (approved):
Looks good.

PR #823, opened 2024-06-04
[Test-operator] Add test-operator CM 

lpiwowar:
This patch adds the test-operator to the openstackcontroler reconcile
loop. The main purpose of this change is to ensure that during the
deployment of controlplane a config map with test-operator-config name
is created. The CM contains information about:

- which images should be used for testing
- name of secret containing clouds.yaml

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1979


kopecmartin:
recheck the depends on is merged

lpiwowar:
recheck
```
  start: '2024-06-20 04:03:01.173956'
  stderr: |-
    Error from server (InternalError): Internal error occurred: failed calling webhook "mopenstackcontrolplane.kb.io": failed to call webhook: Post "https://openstack-operator-controller-manager-service.openstack-operators.svc:443/mutate-core-openstack-org-v1beta1-openstackcontrolplane?timeout=10s": no endpoints available for service "openstack-operator-controller-manager-service"
    Error from server (InternalError): Internal error occurred: failed calling webhook "mopenstackcontrolplane.kb.io": failed to call webhook: Post "https://openstack-operator-controller-manager-service.openstack-operators.svc:443/mutate-core-openstack-org-v1beta1-openstackcontrolplane?timeout=10s": no endpoints available for service "openstack-operator-controller-manager-service"
    Error from server (InternalError): Internal error occurred: failed calling webhook "mopenstackcontrolplane.kb.io": failed to call webhook: Post "https://openstack-operator-controller-manager-service.openstack-operators.svc:443/mutate-core-openstack-org-v1beta1-openstackcontrolplane?timeout=10s": no endpoints available for service "openstack-operator-controller-manager-service"
    Error from server (InternalError): Internal error occurred: failed calling webhook "mopenstackcontrolplane.kb.io": failed to call webhook: Post "https://openstack-operator-controller-manager-service.openstack-operators.svc:443/mutate-core-openstack-org-v1beta1-openstackcontrolplane?timeout=10s": no endpoints available for service "openstack-operator-controller-manager-service"
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>
```
This doesn't seem to be an error related to the patch.

lpiwowar:
recheck

lpiwowar:
recheck

lpiwowar:
recheck

lpiwowar:
recheck

lpiwowar:
recheck

dprince:
I think this looks fine to me. Only question is should we merge this before GA

pkg/openstack/test.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+func ReconcileTest(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, version *corev1beta1.OpenStackVersion, helper *helper.Helper) (ctrl.Result, error) {
+	customData := map[string]string{
+		TempestImageKey: *version.Status.ContainerImages.TestTempestImage,
+		TobikoImageKey:  *version.Status.ContainerImages.TestTobikoImage,
+		CloudsKey:       CloudsConfigMap,
+	}
+
+	_, _, err := configmap.GetConfigMapAndHashWithName(ctx, helper, CloudsConfigMap, instance.Namespace)
```
dprince, line 34: Why do we wait on the Clouds config map here? I guess it is because test-operator needs it to functionally work. In that case should the wait occur in that operator instead?

lpiwowar: Thank you @dprince for the review!

As I'm reading this now it would probably make a more sense to move this waiting into the test-operator. I'm going to update the PR.

lpiwowar: @dprince the wait is removed.

pkg/openstack/test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	TempestImageKey = "tempest-image"
+	TobikoImageKey  = "tobiko-image"
+	CloudsKey       = "clouds-secret-name"
+)
+
+func ReconcileTest(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, version *corev1beta1.OpenStackVersion, helper *helper.Helper) (ctrl.Result, error) {
+	customData := map[string]string{
+		TempestImageKey: *version.Status.ContainerImages.TestTempestImage,
+		TobikoImageKey:  *version.Status.ContainerImages.TestTobikoImage,
```
lpiwowar, line 28: I think this is potentially unsafe if in downstream we won't have the tobiko variable set (there is no downstream build).

dprince (approved):

pkg/openstack/test.go
```diff
@@ -0,0 +1,72 @@
...
+	TestOperatorConfigMap = "test-operator-config"
+
+	TempestImageKey       = "tempest-image"
+	TobikoImageKey        = "tobiko-image"
+	HorizontestImageKey   = "horizontest-image"
+	AnsibletestImageKey   = "ansibletest-image"
+	CloudsKey             = "clouds-secret-name"
+)
+
+func getString(s *string) string {
```
jpodivin, line 27: Could you use getImg for https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/openstack/version.go#L64

lpiwowar: Let me know please what you think about my update. The reason why we need to check whether the variable contains `nil` is that in downstream it might happen that certain test image does not have a build (RELATED_IMAGE_ env var is not set). But maybe I'm missing something.

jpodivin (approved):

jpodivin (approved):

jpodivin (approved):

PR #828, opened 2024-06-05
Fix make install

Akrog:
Running `make install` fails with error:
```
The CustomResourceDefinition "openstackcontrolplanes.core.openstack.org" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
```
This patch fixes this problem by using `kubectl replace` instead of `kubectl apply`.

Replace was chosen over `kubectl apply --server-side` because the apply may fail to merge with the existing CRD, whereas replace will not fail.


Akrog:
recheck

slagle (approved):

Makefile
```diff
@@ -188,7 +188,7 @@ endif
 
 .PHONY: install
 install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl apply -f -
+	$(KUSTOMIZE) build config/crd | kubectl replace --force -f -
```
karelyatin, line 191: Why force is needed? that also removes CRDs and custom resources

karelyatin (changes requested):

Makefile
Akrog: Because I've seen it fail otherwise during development complaining about some incompatibility.

PR #829, opened 2024-06-06
update config samples to include default cell template

SeanMooney:
This change updates the config samples to inculde the cells templates
so that the default behavior fo the cell 1 creation can be modifed in
the future without impacting docs or ci jobs.

currently cell 0 and 1 are created by nova by default
in a feture operator version we plannign to remove the automatic creation
of cell 1 so that the procedure ot add the first and subsequent comptue
cells are identical.

### DIFF
--- a/config/samples/core_v1beta1_openstackcontrolplane.yaml
+++ b/config/samples/core_v1beta1_openstackcontrolplane.yaml
@@ -109,6 +109,21 @@ spec:
       secret: osp-secret
   nova:
     template:
+      cellTemplates:
+        cell0:
+          cellDatabaseAccount: nova-cell0
+          cellDatabaseInstance: openstack
+          cellMessageBusInstance: rabbitmq
+          conductorServiceTemplate:
+            replicas: 1
+          hasAPIAccess: true
+        cell1:
+          cellDatabaseAccount: nova-cell1
+          cellDatabaseInstance: openstack-cell1
+          cellMessageBusInstance: rabbitmq-cell1
+          conductorServiceTemplate:
+            replicas: 1
+          hasAPIAccess: true
       secret: osp-secret
   heat:
     enabled: false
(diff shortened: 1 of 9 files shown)


config/samples/core_v1beta1_openstackcontrolplane.yaml
```diff
@@ -109,6 +109,19 @@ spec:
...
+            replicas: 1
+          hasAPIAccess: true
+        cell1:
+          cellDatabaseAccount: nova-cell1
+          cellDatabaseInstance: openstack-cell1
+          cellMessageBusInstance: rabbitmq-cell1
+          conductorServiceTemplate:
+            replicas: 1
+          hasAPIAccess: true
       secret: osp-secret
```
SeanMooney, line 125: i belive we coudl also just document 

cellTemplates:
  cell0:{}
  cell1:{}

but i want to show specifying the db and message bus explicitly
which actual speaking of which i am missing both in my cell0 template
ill fix that and respin

gibizer:
We also need to modify https://github.com/openstack-k8s-operators/architecture/blob/27ee3f78472aa65107d4f1b3c6f8fb4a3dd2fa4f/lib/control-plane/openstackcontrolplane.yaml#L151 before we remove the defaulting from the nova-operator

SeanMooney:
> We also need to modify https://github.com/openstack-k8s-operators/architecture/blob/27ee3f78472aa65107d4f1b3c6f8fb4a3dd2fa4f/lib/control-plane/openstackcontrolplane.yaml#L151 before we remove the defaulting from the nova-operator

ah good point.
we don't need to rush to remove the defaulting however ill submit a patch to the architecture repo for that too

gibizer (approved):
Looks good to me

SeanMooney:
> > We also need to modify https://github.com/openstack-k8s-operators/architecture/blob/27ee3f78472aa65107d4f1b3c6f8fb4a3dd2fa4f/lib/control-plane/openstackcontrolplane.yaml#L151 before we remove the defaulting from the nova-operator
> 
> ah good point. we don't need to rush to remove the defaulting however ill submit a patch to the architecture repo for that too

done in https://github.com/openstack-k8s-operators/architecture/pull/272

PR #836, opened 2024-06-10
[envtest]Stabilize version controller tests

gibizer:
There was a race condition between the version controller applying the
finalizer to the version instance and the test case cleanup removing
such finalizer. By waiting for the controller to run at least once
before the test finishes is enough to avoid the race as the controller
only adds the finalizer during the first reconcile run that initialize
the conditions.

Also cleaned up the test case itself a bit and added more precise
assert.

Co-authored-by: Dan Prince <dprince@redhat.com>

### DIFF
--- a/tests/functional/openstackversion_controller_test.go
+++ b/tests/functional/openstackversion_controller_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package functional_test
 
 import (
+	"errors"
 	"os"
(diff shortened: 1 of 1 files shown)


dprince (approved):

gibizer:
From the logs I see that all the kuttl tests are actually passed so something failed in cleanup

PR #838, opened 2024-06-11
Move service templates to be ptr to struct

stuggi:
Moves template parameter in the service sections to be a ptr to have them omitted in case its not provided.
Also only triggers the defaulting for the service in the webhook if the service is enabled or a template provided.

If the service was disabled and gets enabled, at least an empty `template: {}` needs to be passed for correct defaulting to happen.

### As background, current operator behavior

This examples only show keystone and placement

* When deploy with a service disabled, the template section gets added with no proper defaulting:

```
    keystone:
      apiOverride: {}
      enabled: false
      template:
        adminProject: ""
        adminUser: ""
        databaseAccount: ""
        databaseInstance: ""
        enableSecureRBAC: false
        memcachedInstance: ""
        override: {}
        passwordSelectors:
          admin: ""
        preserveJobs: false
        rabbitMqClusterName: ""
        region: ""
        replicas: 1
        resources: {}
        secret: ""
        tls:
          api:
            internal: {}
            public: {}
        trustFlushArgs: ""
        trustFlushSchedule: ""
        trustFlushSuspend: false
    placement:
      apiOverride: {}
      enabled: false
      template:
        customServiceConfig: ""
        databaseAccount: ""
        databaseInstance: ""
        override: {}
        passwordSelectors:
          service: ""
        preserveJobs: false
        replicas: 1
        resources: {}
        secret: ""
        serviceUser: ""
        tls:
          api:
            internal: {}
            public: {}
```

* when enable e.g. placement and provide `template: {}`, the section obviously stays as is, as it was already filled before:

```
    placement:
      apiOverride: {}
      enabled: true
      template:
        customServiceConfig: ""
        databaseAccount: ""
        databaseInstance: ""
        override: {}
        passwordSelectors:
          service: ""
        preserveJobs: false
        replicas: 1
        resources: {}
        secret: ""
        serviceUser: ""
        tls:
          api:
            internal: {}
            public: {}
```

the created placementapi cr gets partially defaulted, no `passwordSelector`, no `databaseAccount`:
```
spec:
  containerImage: quay.io/podified-antelope-centos9/openstack-placement-api@sha256:c36e9c0adf778926963592df947c025b87e05099324c8b9c0a0c5b300727ebf2
  customServiceConfig: ""
  databaseAccount: ""
  databaseInstance: openstack
  override:
    service:
      internal:
        metadata:
          labels:
            osctlplane: ""
            osctlplane-service: placement
      public:
        metadata:
          labels:
            osctlplane: ""
            osctlplane-service: placement
  passwordSelectors:
    service: ""
  preserveJobs: false
  replicas: 1
  resources: {}
  secret: osp-secret
  serviceUser: ""
  tls:
    api:
      internal: {}
      public: {}
    caBundleSecretName: combined-ca-bundle

```

* with keystone/placement enabled from start with `template: {}` the default works properly
```
    keystone:
      apiOverride: {}
      enabled: true
      template:
        adminProject: admin
        adminUser: admin
        databaseAccount: keystone
        databaseInstance: ""
        enableSecureRBAC: true
        memcachedInstance: memcached
        override: {}
        passwordSelectors:
          admin: AdminPassword
        preserveJobs: false
        rabbitMqClusterName: rabbitmq
        region: regionOne
        replicas: 1
        resources: {}
        secret: ""
        tls:
          api:
            internal: {}
            public: {}
        trustFlushArgs: ""
        trustFlushSchedule: 1 * * * *
        trustFlushSuspend: false
    placement:
      apiOverride: {}
      enabled: true
      template:
        customServiceConfig: ""
        databaseAccount: placement
        databaseInstance: ""
        override: {}
        passwordSelectors:
          service: PlacementPassword
        preserveJobs: false
        replicas: 1
        resources: {}
        secret: ""
        serviceUser: placement
        tls:
          api:
            internal: {}
            public: {}
```

### operator behavior with the change
* When osctlplane gets created with service disabled and no `template` provided, no template is defaulted and shown
```
    keystone:
      apiOverride: {}
      enabled: false
    placement:
      apiOverride: {}
      enabled: false
```

* when enable the service and provide `template: {}` , the defaulting happens properly
```
    keystone:
      apiOverride: {}
      enabled: true
      template:
        adminProject: admin
        adminUser: admin
        databaseAccount: keystone
        databaseInstance: ""
        enableSecureRBAC: true
        memcachedInstance: memcached
        override: {}
        passwordSelectors:
          admin: AdminPassword
        preserveJobs: false
        rabbitMqClusterName: rabbitmq
        region: regionOne
        replicas: 1
        resources: {}
        secret: ""
        tls:
          api:
            internal: {}
            public: {}
        trustFlushArgs: ""
        trustFlushSchedule: 1 * * * *
        trustFlushSuspend: false

    placement:
      apiOverride: {}
      enabled: true
      template:
        customServiceConfig: ""
        databaseAccount: placement
        databaseInstance: ""
        override: {}
        passwordSelectors:
          service: PlacementPassword
        preserveJobs: false
        replicas: 1
        resources: {}
        secret: ""
        serviceUser: placement
        tls:
          api:
            internal: {}
            public: {}
```

Also the resulting service gets defaulted as expected
resulting service:
```
spec:
  adminProject: admin
  adminUser: admin
  containerImage: quay.io/podified-antelope-centos9/openstack-keystone@sha256:8616532664fb47a2fc0b6c9176a765d50e5e1d1e983155c6620fb63d16a8510a
  databaseAccount: keystone
  databaseInstance: openstack
  enableSecureRBAC: true
  memcachedInstance: memcached
  override:
    service:
      internal:
        metadata:
          labels:
            osctlplane: ""
            osctlplane-service: keystone
      public:
        endpointURL: https://keystone-public-openstack.apps-crc.testing
        metadata:
          labels:
            osctlplane: ""
            osctlplane-service: keystone
  passwordSelectors:
    admin: AdminPassword
  preserveJobs: false
  rabbitMqClusterName: rabbitmq
  region: regionOne
  replicas: 1
  resources: {}
  secret: osp-secret
  tls:
    api:
      internal:
        secretName: cert-keystone-internal-svc
      public:
        secretName: cert-keystone-public-svc
    caBundleSecretName: combined-ca-bundle
  trustFlushArgs: ""
  trustFlushSchedule: 1 * * * *
  trustFlushSuspend: false
```

* create osctlplane with service enabled + `template: {}` provided, defaults correct:
```
    keystone:
      apiOverride: {}
      enabled: true
      template:
        adminProject: admin
        adminUser: admin
        databaseAccount: keystone
        databaseInstance: ""
        enableSecureRBAC: true
        memcachedInstance: memcached
        override: {}
        passwordSelectors:
          admin: AdminPassword
        preserveJobs: false
        rabbitMqClusterName: rabbitmq
        region: regionOne
        replicas: 1
        resources: {}
        secret: ""
        tls:
          api:
            internal: {}
            public: {}
        trustFlushArgs: ""
        trustFlushSchedule: 1 * * * *
        trustFlushSuspend: false
    placement:
      apiOverride: {}
      enabled: true
      template:
        customServiceConfig: ""
        databaseAccount: placement
        databaseInstance: ""
        override: {}
        passwordSelectors:
          service: PlacementPassword
        preserveJobs: false
        replicas: 1
        resources: {}
        secret: ""
        serviceUser: placement
        tls:
          api:
            internal: {}
            public: {}
```

**Note:** The only remaining scenario which does not result in proper defaulting is when enabling the service without specifying at least a `template: {}`. This will result in the known behavior where the parameters get initialized with their go vars defaults

```
    keystone:
      apiOverride: {}
      enabled: true
      template:
        adminProject: ""
        adminUser: ""
        databaseAccount: ""
        databaseInstance: ""
        enableSecureRBAC: false
        memcachedInstance: ""
        override: {}
        passwordSelectors:
          admin: ""
        preserveJobs: false
        rabbitMqClusterName: ""
        region: ""
        replicas: 1
        resources: {}
        secret: ""
        tls:
          api:
            internal: {}
            public: {}
        trustFlushArgs: ""
        trustFlushSchedule: ""
        trustFlushSuspend: false
    placement:
      apiOverride: {}
      enabled: true
      template:
        customServiceConfig: ""
        databaseAccount: ""
        databaseInstance: ""
        override: {}
        passwordSelectors:
          service: ""
        preserveJobs: false
        replicas: 1
        resources: {}
        secret: ""
        serviceUser: ""
        tls:
          api:
            internal: {}
            public: {}
```

And the resulting services misses details as e.g. empty strings are set:

```
spec:
  adminProject: ""
  adminUser: ""
  containerImage: quay.io/podified-antelope-centos9/openstack-keystone@sha256:8616532664fb47a2fc0b6c9176a765d50e5e1d1e983155c6620fb63d16a8510a
  databaseAccount: ""
  databaseInstance: openstack
  enableSecureRBAC: false
  memcachedInstance: ""
  override:
    service:
      internal:
        metadata:
          labels:
            osctlplane: ""
            osctlplane-service: keystone
      public:
        metadata:
          labels:
            osctlplane: ""
            osctlplane-service: keystone
  passwordSelectors:
    admin: ""
  preserveJobs: false
  rabbitMqClusterName: ""
  region: ""
  replicas: 1
  resources: {}
  secret: osp-secret
  tls:
    api:
      internal: {}
      public: {}
    caBundleSecretName: combined-ca-bundle
  trustFlushArgs: ""
  trustFlushSchedule: ""
  trustFlushSuspend: false
```

Depends-on: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/486


stuggi:
Also should run the defaulting when a service is disabled, but a template provided. e.g. in the sample where manila is disabled and a template provided. if we do not calling the defaulting for that we'll hit:
```
$ oc apply -f core_v1beta1_openstackcontrolplane_network_isolation.yaml                                                                                                           
The OpenStackControlPlane "openstack-network-isolation" is invalid: spec.manila.template.dbPurge.age: Invalid value: 0: spec.manila.template.dbPurge.age in body should be greater than or equal to 1  
```

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -192,55 +212,55 @@ func (r *OpenStackControlPlane) ValidateCreateServices(basePath *field.Path) fie
 	errors = append(errors, r.ValidateServiceDependencies(basePath)...)
 
 	// Call internal validation logic for individual service operators
-	if r.Spec.Keystone.Enabled {
+	if r.Spec.Keystone.Enabled && r.Spec.Keystone.Template != nil {
```
gibizer, line 215: here and below: Do we rely on the defaulting webhook to make the Template non nil if the Enabled is true? It does not seems right to skip validation (and therefore accept) the request if Enabled = True and Template = nil.

// after looking at the defaulting webhook changes

As far as I see the defaulting webhook always ensures the template is not nil. (Note that defaulting webhook runs before the validating hook https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#what-are-admission-webhooks). So instead of ignoring the validation if the template is nil we should return an error if the template is nil as it never should be.

Also I think we should validate the Template even if the service is not enabled now as we make sure that the defaulting is correct for the non enabled services too.

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -254,56 +274,56 @@ func (r *OpenStackControlPlane) ValidateUpdateServices(old OpenStackControlPlane
 	errors = append(errors, r.ValidateServiceDependencies(basePath)...)
 
 	// Call internal validation logic for individual service operators
-	if r.Spec.Keystone.Enabled {
-		errors = append(errors, r.Spec.Keystone.Template.ValidateUpdate(old.Keystone.Template, basePath.Child("keystone").Child("template"))...)
+	if r.Spec.Keystone.Enabled && r.Spec.Keystone.Template != nil && old.Keystone.Template != nil {
```
gibizer, line 277: here and below: When the request chages the template from nil to non-nil we still need to validate the content of the non-nil template.

//later

As the defaulting webhook ensures that the template is never nil we should simply reject the Update if either the old or the new template is nil that simply should not happen
Also as we are defaulting the template even if the service is disabled we should be able to validate it when the service is disabled.

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -445,116 +465,221 @@ func setOverrideSpec(override **route.OverrideSpec, anno map[string]string) {
...
+		initializeOverrideSpec(&r.Spec.Cinder.APIOverride.Route, true)
+		r.Spec.Cinder.Template.SetDefaultRouteAnnotations(r.Spec.Cinder.APIOverride.Route.Annotations)
+	}
 
 	// Galera
-	for key, template := range r.Spec.Galera.Templates {
-		if template.StorageClass == "" {
-			template.StorageClass = r.Spec.StorageClass
+	if r.Spec.Galera.Enabled && r.Spec.Galera.Templates != nil {
+		if r.Spec.Galera.Templates == nil {
```
gibizer, line 479: This seems wrong. It cannot be that both the first and the second' "if" is true at the same time as that would require that the template is nil and not nil at the same time.

stuggi: right thats wrong

apis/core/v1beta1/openstackcontrolplane_webhook.go
stuggi: correct, the defaulting runs before the validation, so we could remove it here for validate create. but the initialized template is not written to etcd as `template: {}`. As mentioned in the note at the bottom of the description, to properly initialize and have defaulting you'd still need to pass in `template: {}` for the service. but for disabled services, especially for the service which are per default disabled, or set as disabled I'd like to not have the details set as they are not necessary.

apis/core/v1beta1/openstackcontrolplane_webhook.go
stuggi: as mentioned above setting the defaulting webhook makes the Template non nil, but this is not stored, so on an update the old template is still nil.

apis/core/v1beta1/openstackcontrolplane_webhook.go
stuggi: let me rephrase what I am trying to achieve. If someone disables the service (or is disabled per default). the resulting service section should not include a defaulted template section:

```
  keystone:
    enabled: false
```

if one enables the service, or provides a template in a disabled service we should defaulted. now I think the validation could happen when the service gets enabled?

stuggi: oh seems there is an issue with the what I planned to do ... if one deploys with disabled and we have no template, then performs an update, we'd not properly validate the update. we'd have to initialize old instead of skip it.

stuggi:
https://github.com/openstack-k8s-operators/data-plane-adoption/pull/486 is needed for the failed adoption job

pkg/openstack/galera.go
```diff
@@ -64,7 +68,7 @@ func ReconcileGaleras(
 			// at the initial deployment because there is no SST involved when the DB is bootstrapped
 			// as there are no data to be transferred yet.
 			Subject: &certmgrv1.X509Subject{
-				Organizations: []string{fmt.Sprintf("%s.cluster.local", instance.Namespace)},
+				Organizations: []string{fmt.Sprintf("%s.%s", instance.Namespace, ClusterInternalDomain)},
```
gibizer, line 71: this seems like an unrelated change.

pkg/openstack/nova.go
```diff
@@ -58,6 +58,10 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
 		return ctrl.Result{}, nil
 	}
 
+	if instance.Spec.Nova.Template == nil {
+		instance.Spec.Nova.Template = &novav1.NovaSpec{}
```
gibizer, line 62: nit: NovaSpec is overkill, only the NovaSpecCore part of it will be used after the assignment so you can simply create NovaSpecCore right away.

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -247,55 +267,94 @@ func (r *OpenStackControlPlane) ValidateUpdateServices(old OpenStackControlPlane
...
+		if old.Ironic.Template == nil {
+			old.Ironic.Template = &ironicv1.IronicSpecCore{}
+		}
+		errors = append(errors, r.Spec.Ironic.Template.ValidateUpdate(*old.Ironic.Template, basePath.Child("ironic").Child("template"))...)
 	}
 
 	if r.Spec.Nova.Enabled {
-		errors = append(errors, r.Spec.Nova.Template.ValidateUpdate(old.Nova.Template, basePath.Child("nova").Child("template"))...)
+		if old.Nova.Template == nil {
+			old.Nova.Template = &novav1.NovaSpec{}
```
gibizer, line 285: NovaSpecCore is enough

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -423,116 +482,221 @@ func setOverrideSpec(override **route.OverrideSpec, anno map[string]string) {
...
+		}
+		r.Spec.Neutron.Template.Default()
+		setOverrideSpec(&r.Spec.Neutron.APIOverride.Route, r.Spec.Neutron.Template.GetDefaultRouteAnnotations())
+	}
 
 	// Nova
-	r.Spec.Nova.Template.Default()
+	if r.Spec.Nova.Enabled || r.Spec.Nova.Template != nil {
+		if r.Spec.Nova.Template == nil {
+			r.Spec.Nova.Template = &novav1.NovaSpec{}
```
gibizer, line 600: NovaSpecCore

apis/core/v1beta1/openstackcontrolplane_webhook.go
gibizer: > let me rephrase what I am trying to achieve. If someone disables the service (or is disabled per default). the resulting service section should not include a defaulted template section:
> 
> ```
>   keystone:
>     enabled: false
> ```

OK I see now. I think I misunderstood the condition:
```golang
	if r.Spec.Keystone.Enabled || r.Spec.Keystone.Template != nil {
		if r.Spec.Keystone.Template == nil {
			r.Spec.Keystone.Template = &keystonev1.KeystoneAPISpecCore{}
		}
		r.Spec.Keystone.Template.Default()
	}
```
So if the keystone is disabled and the template is not provided then the defaulting is skipped. 

I won't ask to change this to the following as it is complicated in a different way. 
```golang
	if r.Spec.Keystone.Enabled && r.Spec.Keystone.Template == nil {
		r.Spec.Keystone.Template = &keystonev1.KeystoneAPISpecCore{}
        }
        if r.Spec.Keystone.Template != nil {
		r.Spec.Keystone.Template.Default()
        }
```

So this is good as is.

gibizer (commented):
I understand the logic better now. Thanks. I have couple of nits inline still.

Also one more question. Is it intentional that the template is only validated if the service is enabled? I mean there is a case when the user provides a template but disables the service during create. The validateCreate will not run on the template provided by the user. It is probably OK as nobody will see that template as the controller will not pass it to the service operator. And when the service is enabled later, the validateUpdate will check the template.

pkg/openstack/nova.go
stuggi: right, but then we'd have to change https://github.com/openstack-k8s-operators/openstack-operator/blob/main/apis/core/v1beta1/openstackcontrolplane_types.go#L537 . I did not want to do it as part of this PR. It also requires a change on how the template gets assigned to the service spec in 
 https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/openstack/nova.go#L321 . We'd then should add a func which converts the NovaSpecCore to a NovaSpec

pkg/openstack/galera.go
stuggi: yes, it was a drive by to use the const we have. I can to a separate PR, but should be trivial enough to keep it?

stuggi:
> I understand the logic better now. Thanks. I have couple of nits inline still.
> 
> Also one more question. Is it intentional that the template is only validated if the service is enabled? I mean there is a case when the user provides a template but disables the service during create. The validateCreate will not run on the template provided by the user. It is probably OK as nobody will see that template as the controller will not pass it to the service operator. And when the service is enabled later, the validateUpdate will check the template.

right, this is what we have and had for all the previews. need to check if there are issues with the current implemented validations if we change it.

stuggi:
> > I understand the logic better now. Thanks. I have couple of nits inline still.
> > Also one more question. Is it intentional that the template is only validated if the service is enabled? I mean there is a case when the user provides a template but disables the service during create. The validateCreate will not run on the template provided by the user. It is probably OK as nobody will see that template as the controller will not pass it to the service operator. And when the service is enabled later, the validateUpdate will check the template.
> 
> right, this is what we have and had for all the previews. need to check if there are issues with the current implemented validations if we change it.

@gibizer this require changes, especially in the functional tests. I suggest that we'll do that in a follow up

stuggi:
rebased

gibizer:
> > > I understand the logic better now. Thanks. I have couple of nits inline still.
> > > Also one more question. Is it intentional that the template is only validated if the service is enabled? I mean there is a case when the user provides a template but disables the service during create. The validateCreate will not run on the template provided by the user. It is probably OK as nobody will see that template as the controller will not pass it to the service operator. And when the service is enabled later, the validateUpdate will check the template.
> > 
> > 
> > right, this is what we have and had for all the previews. need to check if there are issues with the current implemented validations if we change it.
> 
> @gibizer this require changes, especially in the functional tests. I suggest that we'll do that in a follow up

follow up is OK to me.

pkg/openstack/galera.go
gibizer: OK to keep it.

pkg/openstack/nova.go
gibizer: Hm. It feels like a bug from the PR that added the usage of the Core Specs. I agree to do a fix separately.

fmount:
+1 from me on this change, I understand that you're going to have a follow up patch. Thanks @gibizer for the multiple reviews on this one, I'm in favor of moving to ptr and adoption would benefit of this change!

pkg/openstack/memcached.go
```diff
@@ -54,17 +54,23 @@ func ReconcileMemcacheds(
...
 			// Check owner UID to ensure the memcached instance is owned by this OpenStackControlPlane instance
 			if ref.UID == instance.GetUID() {
 				owned := false
 
 				// Check whether the name appears in spec
-				for name := range instance.Spec.Memcached.Templates {
-					if name == memcached.GetName() {
-						owned = true
-						break
+				if instance.Spec.Memcached.Templates != nil {
```
abays, line 68: Isn't this covered by L57-59?

stuggi: yes, can remove this check here

abays (approved):
/lgtm

PR #839, opened 2024-06-11
Revert "Check for Ceilometer and Autoscaling dependencies"

vyzigold:
Reverts openstack-k8s-operators/openstack-operator#837

We have a PR  https://github.com/openstack-k8s-operators/telemetry-operator/pull/404 in telemetry-operator to change the "Enabled" values to *bool. Because of this code, we'd need to simultaneously merge that PR along with a PR here to change how we access the variables. Instead the plan is to revert this PR, merge the PR in telemetry-operator and then create a new PR here, which would reintroduce this code but would access the variables correctly as pointers.


stuggi (approved):
/lgtm

PR #843, opened 2024-06-12
Bump Glance with the new storage interface

fmount:
We currently provide both `StorageClass` and `StorageRequest` that can be propagated to the `Glance` underlying instances.
However, it might be possible that operators use `extraMounts` to plug a `NFS` share mapped to  `/var/lib/glance/os_glance_staging_store`.
In this case the `API` should be updated to not provision any `PVC` and let the human operator take care about it via the `extraMounts` interface.

This patch updates the `Glance` bits to reflect the new behavior introduced in the `glance-operator`.

Jira: https://issues.redhat.com/browse/OSPRH-6719

### DIFF
--- a/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
+++ b/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
@@ -4373,8 +4373,6 @@ spec:
                                     type: object
                                   type: object
                               type: object
-                            pvc:
-                              type: string
                             replicas:
                               default: 1
                               format: int32
@@ -4412,10 +4410,15 @@ spec:
                                     x-kubernetes-int-or-string: true
                                   type: object
                               type: object
-                            storageClass:
-                              type: string
-                            storageRequest:
-                              type: string
+                            storage:
+                              properties:
+                                external:
+                                  type: boolean
+                                storageClass:
+                                  type: string
+                                storageRequest:
+                                  type: string
+                              type: object
                             tls:
                               properties:
                                 api:
@@ -4443,7 +4446,6 @@ spec:
                               type: string
                           required:
                           - containerImage
-                          - storageRequest
                           type: object
                         type: object
                       imageCache:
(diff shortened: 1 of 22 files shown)


konan-abhi (approved):
Looks good!!

fmount:
@fao89 do you know what's going on w/ `make docs` ? I tried it locally and it doesn't change anything compared to what I already pushed here, but I'm wondering if I'm missing something.

fao89:
@fmount I don't know what is happening, but I checkout this PR, ran make docs and it did change. My best guess is something with the binaries, I would try deleting what you have on openstack-operator/bin and run make docs again to see if it changes the docs for you

fmount:
> @fmount I don't know what is happening, but I checkout this PR, ran make docs and it did change. My best guess is something with the binaries, I would try deleting what you have on openstack-operator/bin and run make docs again to see if it changes the docs for you



> @fmount I don't know what is happening, but I checkout this PR, ran make docs and it did change. My best guess is something with the binaries, I would try deleting what you have on openstack-operator/bin and run make docs again to see if it changes the docs for you

I think it's `bundle` that makes the difference, and it's not in `openstack-operator/bin`. I'll look into that but so far I'm not able to reproduce (locally) that issue.

fao89:
it makes sense, try to replace `manifests` with `bundle` on the docs dependencies: https://github.com/openstack-k8s-operators/openstack-operator/blob/main/Makefile#L94

fmount:
> it makes sense, try to replace `manifests` with `bundle` on the docs dependencies: https://github.com/openstack-k8s-operators/openstack-operator/blob/main/Makefile#L94
> 
> So there are weird things happening with ruby and bundle: do we enforce a specific version?

ok the combination of fedora40 with ruby-v3.3 did the trick, but if I slightly change version(s) (even for the asciidoctor) gem I generate a diff output.
@fao89 now pre-commit check passes, thanks for the help!

fmount:
@abays @stuggi @konan-abhi we have green CI here and I'm wondering if we can land this

stuggi (approved):
/lgtm

konan-abhi (approved):

fmount:
mm rechecking kuttl as it seems failing on `FIPS` check :/

PR #847, opened 2024-06-12
Use non static name for the Cinder and Glance CRs

Akrog:
When the openstack operator creates the cinder CR it always uses the
"cinder" name and there is no way to change it.

This means that the names of the pods, and therefore the reported
hostname inside them, will always be the same for the backup service
(cinder-backup-0) and for the volume service in 2 deployments if they
have the same key in the volumes map.

This may be fine for some deployments, but it creates problems in
others, because many cinder drivers expect the hostnames of the hosts
attaching volumes to be unique.

Because they expect them to be unique they use the hostname as a key to
store information on the storage array such as the iSCSI initiator name
or the HBAs addresses.

If they are not unique there can be unexpected behavior like hosts being
unable to attach.

This is very common in CI systems, but is not limited to them as it can
also happen in production deployments using the same storage arrays.

A similar thing would happen when Glance is using Cinder as a backend.

With this PR the cinder and glance sections have a new field called
`randomPodNames` each so the openstack-operator can generate a non
static name for the `Cinder` and `Glance` CRs.

This new name is the same as before but with a suffix of a dash (`-`)
followed by the first 5 characters of the `OpenStackControlPlane` UID.

The is still a small chance to have duplicated pod names, but the chance
should be small enough to be safe to ignore.

Related PRs:
- https://github.com/openstack-k8s-operators/cinder-operator/pull/399
- https://github.com/openstack-k8s-operators/cinder-operator/pull/400
- https://github.com/openstack-k8s-operators/glance-operator/pull/555

Jira: [OSPRH-7396](https://issues.redhat.com//browse/OSPRH-7396)

### DIFF
--- a/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
+++ b/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
@@ -4528,6 +4528,9 @@ spec:
                     - memcachedInstance
                     - secret
                     type: object
+                  uniquePodNames:
+                    default: false
+                    type: boolean
                 type: object
               heat:
                 properties:
--- a/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
+++ b/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
@@ -1661,6 +1661,9 @@ spec:
                     - rabbitMqClusterName
                     - secret
                     type: object
+                  uniquePodNames:
+                    default: false
+                    type: boolean
                 type: object
               designate:
                 properties:
--- a/apis/core/v1beta1/openstackcontrolplane_types.go
+++ b/apis/core/v1beta1/openstackcontrolplane_types.go
@@ -414,6 +414,13 @@ type GlanceSection struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// APIOverride, provides the ability to override the generated manifest of several child resources.
 	APIOverride map[string]Override `json:"apiOverrides,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// UniquePodNames - Use a unique prefix for glance CRs to have unique pod names.
+	// Convenient to avoid podname (and thus hostname) collision between different deployments.
+	// Useful for CI jobs as well as preproduction and production environments that use the same storage backend, etc.
+	UniquePodNames bool `json:"uniquePodNames"`
 }
 
 // CinderSection defines the desired state of Cinder service
--- a/apis/core/v1beta1/openstackcontrolplane_types.go
+++ b/apis/core/v1beta1/openstackcontrolplane_types.go
@@ -433,6 +433,13 @@ type CinderSection struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// APIOverride, provides the ability to override the generated manifest of several child resources.
 	APIOverride Override `json:"apiOverride,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// UniquePodNames - Use a unique prefix for cinder CRs to have unique pod names.
+	// Convenient to avoid podname (and thus hostname) collision between different deployments.
+	// Useful for CI jobs as well as preproduction and production environments that use the same storage backend, etc.
+	UniquePodNames bool `json:"uniquePodNames"`
 }
 
 // GaleraSection defines the desired state of Galera services
--- a/pkg/openstack/cinder.go
+++ b/pkg/openstack/cinder.go
@@ -21,9 +21,27 @@ import (
 
 // ReconcileCinder -
 func ReconcileCinder(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, version *corev1beta1.OpenStackVersion, helper *helper.Helper) (ctrl.Result, error) {
+	cinderName := "cinder"
+	altCinderName := fmt.Sprintf("cinder-%s", instance.UID[:5])
+
+	if instance.Spec.Cinder.UniquePodNames {
+		cinderName, altCinderName = altCinderName, cinderName
+	}
+
+	// Ensure the alternate cinder CR doesn't exist, as the ramdomPodNames flag may have been toggled
 	cinder := &cinderv1.Cinder{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "cinder",
+			Name:      altCinderName,
+			Namespace: instance.Namespace,
+		},
+	}
+	if res, err := EnsureDeleted(ctx, helper, cinder); err != nil {
+		return res, err
+	}
+
+	cinder = &cinderv1.Cinder{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cinderName,
 			Namespace: instance.Namespace,
 		},
 	}
@@ -58,7 +76,7 @@ func ReconcileCinder(ctx context.Context, instance *corev1beta1.OpenStackControl
 	}
 
 	// When component services got created check if there is the need to create a route
-	if err := helper.GetClient().Get(ctx, types.NamespacedName{Name: "cinder", Namespace: instance.Namespace}, cinder); err != nil {
+	if err := helper.GetClient().Get(ctx, types.NamespacedName{Name: cinderName, Namespace: instance.Namespace}, cinder); err != nil {
 		if !k8s_errors.IsNotFound(err) {
 			return ctrl.Result{}, err
 		}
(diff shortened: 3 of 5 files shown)


apis/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -188,6 +188,13 @@ type OpenStackControlPlaneSpec struct {
 	// template Section, the globally defined ExtraMounts are ignored and
 	// overridden for the operator which has this section already.
 	ExtraMounts []OpenStackExtraVolMounts `json:"extraMounts,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// RandomPodNames - Use a unique prefix for cinder and Glance CRs to have unique pod names.
+	// Convenient to avoid podname (and thus hostname) collision between different deployments.
+	// Useful for CI jobs as well as preproduction and production environments that use the same storage backend, etc.
+	RandomPodNames bool `json:"randomPodNames"`
```
gibizer, line 197: I would put this into the cinder and the glance section of the Spec respectively that would:
* signal that this is only applied to the give subresource
* allow to enable the feature selectively per subresource

apis/bases/core.openstack.org_openstackcontrolplanes.yaml
```diff
@@ -564,6 +564,9 @@ spec:
                   enabled:
                     default: true
                     type: boolean
+                  randomPodNames:
```
bogdando, line 567: this looks like getting away from standard kubernetes resources. Statefulsets guarantee stable names, deployments, dsets, and replicasets are about to use dynamic names for pods.
Please do not introduce artifical CRD parameters, maybe instead allow choosing the kind of the resource it need to create, like deployment, ds, or sts?

bogdando (changes requested):
please see the random naming spec concerns

apis/bases/core.openstack.org_openstackcontrolplanes.yaml
Akrog: These are stable names, the `random` part is not really true.
The value is based on the 5 first characters of the UID of the `OpenStackControlPlane` CR.
It doesn't change even if everything is manually destroyed and the openstack-operator recreates it.
The pods and CRs will have a stable name.

I'm open to changing the name if you feel like it's not clear, but to be honest I don't know of a better name.

bogdando: The name should be connected to the kind of the managed resource, for pods, either sts, or ds, or deployment.
We shouldn't invent custom resouces controllers with custom nameing schemes

jpodivin: How about a `boundName`? As it's bound to the control plane?

Akrog: All those names sound theoretically nice, but I doubt customers will understand what the field is meant to do, because even I have trouble figuring it out just from the name.

Since the `random` part of the name sounds weird, we could change it to `uniquePodNames`.

fmount: I agree that the primary goal is to have customers understand what the boolean does. We could change to `uniquePodNames` if it sounds better (I'm bad at that kind of things :/) but I think the patch goes to the right direction.
Only `Glance` and `Cinder` sections have the parameter, which is set to `false` (and the job executions prove that there are no regressions associated to this code).

pkg/openstack/cinder.go
```diff
@@ -21,9 +21,27 @@ import (
...
+
+	// Ensure the alternate cinder CR doesn't exist, as the ramdomPodNames flag may have been toggled
 	cinder := &cinderv1.Cinder{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "cinder",
+			Name:      altCinderName,
+			Namespace: instance.Namespace,
+		},
+	}
+	if res, err := EnsureDeleted(ctx, helper, cinder); err != nil {
```
fmount, line 38: I assume this will cause a redeploy in case an old `cinder` CR exists, correct?

fmount (approved):
/lgtm

fmount:
I'm ok with the current patch, and I don't have a strong option on using `uniquePodNames` instead of `randomPodNames`, so we can move forward with this unless there are strong concerns to address.

fmount:
As discussed, let's have an additional iteration to move `randomPodNames` into `uniquePodNames` and we'll review again.

pkg/openstack/cinder.go
Akrog: Yes, if the alternative CR exists a redeploy is necessary to have the right naming scheme.

fmount (approved):
/lgtm

fmount:
@stuggi @abays not sure you can give another pass to this and land to solve the nfs related issue. I think the main point was about the CRD parameter name and it has been added under Cinder and Glance templates.
Just looking at kuttl we can see the old behavior is preserved [1], so I think it's good to go.

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_openstack-operator/847/pull-ci-openstack-k8s-operators-openstack-operator-main-openstack-operator-build-deploy-kuttl/1805679081946091520/artifacts/openstack-operator-build-deploy-kuttl/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-239f9b57d856c30ba157bcdfd75edb27270a8ba8613199defb1cfecf726f30c5/namespaces/openstack/crs/glances.glance.openstack.org/glance.yaml

fmount:
> > @stuggi @abays not sure you can give another pass to this and land to solve the nfs related issue. I think the main point was about the CRD parameter name and it has been added under Cinder and Glance templates. Just looking at kuttl we can see the old behavior is preserved [1], so I think it's good to go.
> > [1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_openstack-operator/847/pull-ci-openstack-k8s-operators-openstack-operator-main-openstack-operator-build-deploy-kuttl/1805679081946091520/artifacts/openstack-operator-build-deploy-kuttl/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-239f9b57d856c30ba157bcdfd75edb27270a8ba8613199defb1cfecf726f30c5/namespaces/openstack/crs/glances.glance.openstack.org/glance.yaml
> 
> I am ok to land this just wondering if it would be better to keep this in the component api as a parameter to have it transparent for the openstack-operator. like the openstack-operator always creates with the static name for a cinder or glance cr in the namespace, but the service operator could still use unique names for the specific deployments which need it (c-v, c-b,.). iiuc c-a and c-s won't need this.

I think the main goal is the opposite: the service operator can accept whatever is passed (if we build a CR with a name != `glance` or `cinder`, it will work), but considering the main CR is generated by the `openstack-operator`, this kind of features applies only when the deployment is coordinated from a CR where both glance and cinder are defined in this context.
The service operator will work, but we can keep it generic enough to work with and without this logic.
In addition, having this code living here makes any potential maintenance easy (because we can keep the behavior aligned for multiple operators from the same place).
Not sure if you want to discuss and dig more into this problem, but if you think it's problematic we can discuss about changing it (after GA?)

stuggi:
> > > @stuggi @abays not sure you can give another pass to this and land to solve the nfs related issue. I think the main point was about the CRD parameter name and it has been added under Cinder and Glance templates. Just looking at kuttl we can see the old behavior is preserved [1], so I think it's good to go.
> > > [1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_openstack-operator/847/pull-ci-openstack-k8s-operators-openstack-operator-main-openstack-operator-build-deploy-kuttl/1805679081946091520/artifacts/openstack-operator-build-deploy-kuttl/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-239f9b57d856c30ba157bcdfd75edb27270a8ba8613199defb1cfecf726f30c5/namespaces/openstack/crs/glances.glance.openstack.org/glance.yaml
> > 
> > 
> > I am ok to land this just wondering if it would be better to keep this in the component api as a parameter to have it transparent for the openstack-operator. like the openstack-operator always creates with the static name for a cinder or glance cr in the namespace, but the service operator could still use unique names for the specific deployments which need it (c-v, c-b,.). iiuc c-a and c-s won't need this.
> 
> I think the main goal is the opposite: the service operator can accept whatever is passed (if we build a CR with a name != `glance` or `cinder`, it will work), but considering the main CR is generated by the `openstack-operator`, this kind of features applies only when the deployment is coordinated from a CR where both glance and cinder are defined in this context. The service operator will work, but we can keep it generic enough to work with and without this logic. In addition, having this code living here makes any potential maintenance easy (because we can keep the behavior aligned for multiple operators from the same place). Not sure if you want to discuss and dig more into this problem, but if you think it's problematic we can discuss about changing it (after GA?)

thanks for the additional details, from my side we can keep it as is. Lets wait for  @abays review and get it in

abays (approved):
/lgtm

abays:
```
error: build error: Failed to push image: trying to reuse blob sha256:e394ea8406c7ed85ddc862f882ec77dcfd3863de3cb90fd8006238d521d22d44 at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout
```

PR #851, opened 2024-06-14
Add dataplane docs

fao89:
https://logserver.rdoproject.org/51/851/d054eec5ab28ee48bb5e4103546d1604a63673ed/github-check/openstack-operator-docs-preview/4fb110a/docs_build/docs_build/openstack/


.github/workflows/docs.yaml
```diff
@@ -35,21 +35,23 @@ jobs:
...
 
       - name: Prepare gh-pages branch
         run: |
-          git restore docs/assemblies/custom_resources.adoc
           git config user.name github-actions
           git config user.email github-actions@github.com
 
           git branch -D gh-pages &>/dev/null || true
           git checkout -b gh-pages 4cd0193fc6c5bc7e76f3a0148d0447fb0d7fbe6a
+          git revert 4cd0193fc6c5bc7e76f3a0148d0447fb0d7fbe6a
```
jpodivin, line 47: Please add a comment about why are we reverting this.

jpodivin (changes requested):

.github/workflows/docs.yaml
fao89: it is just for creating an empty branch, but I found a better way

rebtoor (approved):

.github/workflows/docs.yaml
rebtoor: ++ for `--orphan` usage

fao89:
```
Waiting the cluster to be up
..........................................................................................make: *** [Makefile:2098: ceph] Error 1
```

fao89:
recheck

PR #852, opened 2024-06-14
Drop DataPlaneService ConfigMaps and Secrets

fao89:
- Drops the ConfigMaps and Secrets fields from OpenStackDataPlaneService.
- Updates all services under config/services to use the OpenStackDataPlaneService.Spec.DataSources field instead.
- Update docs for the switch to DataSources.

Copied: https://github.com/openstack-k8s-operators/dataplane-operator/pull/923


docs/assemblies/custom_resources.adoc
```diff
@@ -15397,7 +15397,7 @@ AutoscalingSection defines the desired state of the autoscaling service
 
 | enabled
 | Enabled - Whether OpenStack autoscaling service should be deployed and managed
-| bool
+| *bool
```
jpodivin, line 15400: These changes to the docs appear completely unrelated to the PR.

jpodivin (changes requested):
I don't agree with dropping of the labels, at very least not all of them. There is still need for fine grained resource selection and I don't think removing labels supporting that is especially useful.

docs/assemblies/custom_resources.adoc
fao89: make docs wasn't added into pre-commit hook, I added it here: https://github.com/openstack-k8s-operators/openstack-operator/pull/851

fao89:
> I don't agree with dropping of the labels, at very least not all of them. There is still need for fine grained resource selection and I don't think removing labels supporting that is especially useful.

I just copied @slagle PRs, but I believe the labels were removed because it didn't have a pattern among the services.
I think we can add labels using the webhook, this way we would ensure a pattern

rabi (commented):
/lgtm

rabi:
Needs a rebase after the docs patch it seems.

PR #853, opened 2024-06-14
Introducing spec level validation of dataplane/controlplane TLS consistency

jpodivin:
Verifies that TLS settings for nodeset are consistent with those of existing control plane, if there is one and only one.

If there are multiple control planes the process will result in error, same if it isn't possible to retrieve list of control planes.

Tests are included

I'm porting this now to the joint repo.

### DIFF
--- a/apis/dataplane/v1beta1/openstackdataplanenodeset_types.go
+++ b/apis/dataplane/v1beta1/openstackdataplanenodeset_types.go
@@ -17,9 +17,11 @@ limitations under the License.
 package v1beta1
 
 import (
+	"context"
 	"fmt"
 
 	"golang.org/x/exp/slices"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	infranetworkv1 "github.com/openstack-k8s-operators/infra-operator/apis/network/v1beta1"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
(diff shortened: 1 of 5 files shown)


apis/dataplane/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -290,3 +292,49 @@ func (r *OpenStackDataPlaneNodeSetSpec) duplicateNodeCheck(nodeSetList *OpenStac
...
+		Namespace: namespace,
+	}
+
+	// Attempt to get list of all ControlPlanes fail if that isn't possible
+	if err = reconcilerClient.List(ctx, &controlPlanes, &opts); err != nil {
+		return err
+	}
+	// Verify TLS status of control plane only if there is a single one
+	// report error if there are multiple, or proceed if there are none
+	if len(controlPlanes.Items) > 1 {
```
stuggi, line 312: fyi, the ctlplane has a validation webhook, which only allows a single ctlplane per namespace.

apis/dataplane/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -290,3 +292,49 @@ func (r *OpenStackDataPlaneNodeSetSpec) duplicateNodeCheck(nodeSetList *OpenStac
...
+		}
+	}
+
+	return err
+}
+
+// Do TLS flags match in control plane ingress, pods and data plane
+func (r *OpenStackDataPlaneNodeSetSpec) TLSMatch(controlPlane openstackv1.OpenStackControlPlane) *field.Error {
+
+	if controlPlane.Spec.TLS.Ingress.Enabled != r.TLSEnabled || controlPlane.Spec.TLS.PodLevel.Enabled != r.TLSEnabled {
```
stuggi, line 328: `controlPlane.Spec.TLS.Ingress.Enabled` reflects tls on public endpoints. This can be different then the dataplane tls setting, why you think its an issue if those are different?

apis/dataplane/v1beta1/openstackdataplanenodeset_types.go
jpodivin: That is a good point. I didn't know that. In that case we can dispose of the conditional.

apis/dataplane/v1beta1/openstackdataplanenodeset_types.go
stuggi: I think we can ignore `TLS.Ingress.Enabled` for dataplane tls validation as tls on the public endpoints does not reflect if tls should be enabled for the dataplane. only if `TLS.Podified.Enabled` is enabled, the dataplane must also use tls.

jpodivin: @vakwetu wdyt?

jpodivin:
recheck

jpodivin:
recheck

Deydra71:
@jpodivin  Hi! 
I'm curious why this has been closed.

PR #860, opened 2024-06-18
Remove the deployed attribute from deployment

rabi:
We reset the conditions and `deployed: false` for every reconcile now and code to check the deployed flag during reconcile is not relevant and dead code.

We have also made the deployment spec immutable which won't allow for any changes to the spec.


fao89 (approved):

slagle:
Revert proposed at https://github.com/openstack-k8s-operators/openstack-operator/pull/970 due to https://issues.redhat.com/browse/OSPRH-8923

PR #864, opened 2024-06-18
Drop make docs from pre-commit hooks

fao89:
Different versions of ruby/bundler may yield different outputs, removing it for now


fmount (approved):
/lgtm

jpodivin (approved):

PR #868, opened 2024-06-19
Fix duplicateNode check

rabi:
- Check for both create and update/patch validation
- Check for duplicates in the spec itself


bshephar (approved):
Yeah, makes sense.

I think we're going to need to revisit this whole thing soon because it's going to be quite burdensome with hundreds of nodes. But I'm not sure what the best solution to that problem is right now.

It makes sense to check for duplicates in the same NodeSet, so running on Create and Update makes sense. lgtm

PR #870, opened 2024-06-19
Cinder: Fix `cinderVolumes` changes

Akrog:
A regression was introduced on the handling of the `cinderVolumes` on commit 4930eed3380bc945279173c989d1a4c95cdda90e.

The current code is not able to remove entries from the `cinderVolumes` section so backends cannot be removed and they cannot be renamed.

When we try to remove a volume backend, the volume remains in the existing `Cinder` CR.  And when we try to rename a volume backend, we end up with 2 backends in the `Cinder` CR, one with the old name and one with the new name.


ASBishop (approved):
/lgtm

PR #873, opened 2024-06-20
Don't use watcher function for logging job status

rabi:
This also optimizes to only query failed pods when the deployment has failed.

### DIFF
--- a/controllers/dataplane/openstackdataplanenodeset_controller.go
+++ b/controllers/dataplane/openstackdataplanenodeset_controller.go
@@ -385,7 +385,7 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 		}
 	}
 
-	isDeploymentReady, isDeploymentRunning, isDeploymentFailed, err := checkDeployment(helper, instance)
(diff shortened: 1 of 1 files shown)


rabi:
Last failure `Failed to run 'make crc_storage' target. Aborting`

controllers/dataplane/openstackdataplanenodeset_controller.go
```diff
@@ -436,31 +436,38 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
...
+		if _err != nil {
+			Log.Error(err, "unable to retrieve list of pods for dataplane diagnostic")
+		} else {
+			for _, pod := range podsList.Items {
+				Log.Info(fmt.Sprintf("openstackansibleee job %s failed due to %s with message: %s", pod.Name, pod.Status.Reason, pod.Status.Message))
+			}
 		}
-		instance.Status.Conditions.MarkFalse(condition.DeploymentReadyCondition,
-			condition.ErrorReason, condition.SeverityError,
-			deployErrorMsg)
```
fao89, line 446: I think we should keep the status update

rabi: oops that was unintentional.

PR #877, opened 2024-06-21
Bump galera storageRequest to 5G

stuggi:
Use a more production valid storageRequest in the sample files for galera.


stuggi:
not sure if this is to high for ci, but lets see

PR #881, opened 2024-06-24
Dedupe services to be deployed for nodeset

rabi:
- Multiple Services of same ServiceType/ServiceName in nodeset
- Global Services in multiple NodeSets for a deployment

Jira: https://issues.redhat.com/browse/OSPRH-8104

### DIFF
--- a/pkg/dataplane/const.go
+++ b/pkg/dataplane/const.go
@@ -71,4 +71,8 @@ const (
 
 	//HostnameLabel label for marking secrets to be watched for changes
 	HostnameLabel = "hostname"
+
+	//AllNodeSets Key used in nodeset service map for all nodesets when
+	//using serviceOverride.
+	AllNodeSets = "AllNodeSets"
 )
--- a/pkg/dataplane/service.go
+++ b/pkg/dataplane/service.go
@@ -25,6 +25,7 @@ import (
 
 	"golang.org/x/exp/slices"
 	yaml "gopkg.in/yaml.v3"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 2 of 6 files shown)


rabi:
Marking WIP to add functional tests.

slagle:
Is this something you view as blocker for GA? It was shared in the podified call yesterday that "We should consider ourselves in blockers-only mode at this point". If we intend to proceed with this for GA, it should likely have an associated blocking Jira.

rabi:
> Is this something you view as blocker for GA?

I think this is a good UX improvement. I was not aware of the decision and also started this before Tuesday's podified call. Are we applying this rule for all PRs as I notice PRs[1] merged yesterday without blocker flags? Having said that, if we think we should not merge this before GA, it can wait.

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/847

pkg/dataplane/service.go
```diff
@@ -160,38 +162,60 @@ func EnsureServices(ctx context.Context, helper *helper.Helper, instance *datapl
...
 		if err != nil {
-			helper.GetLogger().Error(err, fmt.Sprintf("error getting service %s for consistency check", svc))
-			return err
+			return nil, err
 		}
-
-		if service.Spec.DeployOnAllNodeSets {
-			if serviceInList(service.Name, globalServices) {
-				return fmt.Errorf("global service %s defined multiple times", service.Name)
+		nodeSetServiceMap[AllNodeSets] = services
```
jpodivin, line 178: This line is missing coverage in functional tests. Since it's such an important feature, I think it does deserve it.

pkg/dataplane/const.go
```diff
@@ -71,4 +71,7 @@ const (
 
 	//HostnameLabel label for marking secrets to be watched for changes
 	HostnameLabel = "hostname"
+
+	//AllNodeSets
```
jpodivin, line 75: Please give this constant further annotation, just to make clear what is it for.

jpodivin (changes requested):

slagle:
> > Is this something you view as blocker for GA?
> 
> I think this is a good UX improvement. I was not aware of the decision and also started this before Tuesday's podified call. Are we applying this rule for all PRs as I notice PRs[1] merged yesterday without blocker flags? Having said that, if we think we should not merge this before GA, it can wait.
> 
> [1] #847

we can check with @cmsj, the guidance came from him

pkg/dataplane/service.go
rabi: Well, there was not functional test for `serviceOverride` in deployment, so I did not consider it. Anyway, added one now to cover this.

pkg/dataplane/const.go
rabi: OK, this constant is mostly internal usage, added a verbose comment.

rabi:
`error: build error: Failed to push image: trying to reuse blob sha256:e394ea8406c7ed85ddc862f882ec77dcfd3863de3cb90fd8006238d521d22d44 at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout`

pkg/dataplane/service.go
jpodivin: The code you have introduced substantially increases complexity of the scenario. We need to ensure everything behaves properly. Thanks for taking care of that.

jpodivin:
The code has +2 from me, even though we can't merge it right away. Couple of things to consider in the meantime.
* add JIRA epic/task for tracking
* set up followup PR for documentation of this behavior

rabi:
> even though we can't merge it right away. 

Why is that (as discussed yesterday we're not in blocker only mode yet and this is an UX improvement)? As per documentation is concerned, we need to just document that:
-  For duplicate services, later  services in the service list would be ignored
-  For global services in multiple nodesets, we won't fail like before.

I'll will do a doc only follow-up next week.

jpodivin:
This isn't a blocker so I'm putting DNM until we either get the priority or the GA is out.

rabi:
recheck

slagle:
recheck

https://github.com/openstack-k8s-operators/ci-framework/pull/2009 is merged now, let's see if that fixes the docs job

slagle (approved):

slagle:
recheck

dropped the outdated Depends-On from the description

rabi:
`Failed to run 'make crc_storage' target. Aborting`

PR #882, opened 2024-06-24
Validate galera templates storageRequest via validation webhook

stuggi:
Notify the enduser if the configured DB storageRequest is too low for production use. It is not blocking to allow lower settings for test/ci systems. The output would be like:

```
$ oc apply -n openstack -f core_v1beta1_openstackcontrolplane_galera_network_isolation_3replicas.yaml
Warning: spec.galera.template[openstack-cell1].storageRequest: 1G is not appropriate for production! For production use at least 5G!
Warning: spec.galera.template[openstack].storageRequest: 500M is not appropriate for production! For production use at least 5G!

```
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/522
Depends-On: https://github.com/openstack-k8s-operators/mariadb-operator/pull/234


stuggi:
recheck

PR #883, opened 2024-06-24
Standardize config manager defaults

yazug:


dprince (commented):
I am open to trying this, but it means we'll have to manually maintain it. This means all DFGs who maintain the service operators for these openstack services will need to bump changes to the settings here which will override anything they have set within their service operator repos directly. Today these settings are merged in by https://github.com/dprince/openstack-operator/blob/main/custom-bundle.Dockerfile#L17 from the bundle_extra_data cache that is generated locally from the Makefile target. What I like about inlining these manually is that we would decrease complexity and simplify the bundle generation

yazug:
Another advantage is that we can see up front if there are any warning from operator-sdk when generating the bundle.  Currently there are 3 warnings that are only visible in the bundle creation job but likely missed or glossed over.

WARN[0002] warning: multiple related images with the same image ref, "aodh-api-image-url-default", and different names found.The image will only be listed once with an empty name.It is recmmended to either remove the duplicate or use the exact same name.
WARN[0002] warning: multiple related images with the same image ref, "apache-image-url-default", and different names found.The image will only be listed once with an empty name.It is recmmended to either remove the duplicate or use the exact same name.
WARN[0002] warning: multiple related images with the same image ref, "infra-dnsmasq-image-url-default", and different names found.The image will only be listed once with an empty name.It is recmmended to either remove the duplicate or use the exact same name.

jpodivin (changes requested):
While the change itself is sensible, it shouldn't be done without including maintainers of other operators, as they will bear some of the costs.

Furthermore, if we merge this, there should be an associated change to documentation in affected operators, ensuring that the new de facto dependency is known and recognized.

dprince:
@yazug I pushed a PR https://github.com/openstack-k8s-operators/openstack-operator/pull/928 that has these changes (rebased) and also simplifies how bundles are built by getting rid of csv_merger

PR #890, opened 2024-06-27
Create cloudrc file in `openstackclient` pod

Akrog:
We are currently only creating the `clouds.yaml` and `secure.yaml` in the `openstackclient` pod, and that is enough for the openstack client, but unfortunately this is not enough since there is no parity between that client and the individual clients.

For example it is missing (though it has upstream patches in progress):

- The `openstack volume service list` cannot show the cluster.
- The `openstack volume service list` cannot show the backend state.
- There is no support for any of the default volume type commands: set, show, list, unset.

These are some of the ones we know about, but there could still be some other missing.

Users will need to use the old `cinder` client, so we should make it convenient for them by having the `cloudrc` available in the `openstackclient` pod.

Depends-On: https://github.com/openstack-k8s-operators/keystone-operator/pull/432

### DIFF
--- a/pkg/openstackclient/funcs.go
+++ b/pkg/openstackclient/funcs.go
@@ -106,6 +106,11 @@ func clientPodVolumeMounts() []corev1.VolumeMount {
 			MountPath: "/home/cloud-admin/.config/openstack/secure.yaml",
 			SubPath:   "secure.yaml",
 		},
+		{
+			Name:      "openstack-config-secret",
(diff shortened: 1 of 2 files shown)


bshephar (commented):
/lgtm

fmount:
This patch works for me as well and I see the dependency already merged.

fmount (approved):
/lgtm

PR #891, opened 2024-06-27
Don't mount on /runner/env/ssh_key

rabi:
Though this is the default from which ansible-runner uses the ssh_key, it does not match with `ansible_ssh_private_key_file` var in the inventrory. This results in ambiguous error when ssh_key is not valid as it looks for `ansible_ssh_private_key_file` as fallback and the location does not exist.

Let's not mount on the default runner location and instead in location that matches with `ansible_ssh_private_key_file`.

Related Jira: https://issues.redhat.com/browse/OSPRH-8085
Related Jira: https://issues.redhat.com/browse/OSPRH-8057
Signed-off-by: rabi <ramishra@redhat.com>

### DIFF
--- a/pkg/dataplane/util/ansible_execution.go
+++ b/pkg/dataplane/util/ansible_execution.go
@@ -142,18 +142,12 @@ func AnsibleExecution(
 
 		for _, sshKeyNodeName := range sshKeys {
 			sshKeySecret := sshKeySecrets[sshKeyNodeName]
-			if service.Spec.DeployOnAllNodeSets {
-				sshKeyName = fmt.Sprintf("ssh-key-%s", sshKeyNodeName)
-				sshKeyMountSubPath = fmt.Sprintf("ssh_key_%s", sshKeyNodeName)
-				sshKeyMountPath = fmt.Sprintf("/runner/env/ssh_key/%s", sshKeyMountSubPath)
-			} else {
-				if sshKeyNodeName != nodeSet.GetName() {
-					continue
-				}
-				sshKeyName = "ssh-key"
-				sshKeyMountSubPath = "ssh_key"
-				sshKeyMountPath = "/runner/env/ssh_key"
+			if !service.Spec.DeployOnAllNodeSets && sshKeyNodeName != nodeSet.GetName() {
+				continue
 			}
+			sshKeyName = fmt.Sprintf("ssh-key-%s", sshKeyNodeName)
+			sshKeyMountSubPath = fmt.Sprintf("ssh_key_%s", sshKeyNodeName)
+			sshKeyMountPath = fmt.Sprintf("/runner/env/ssh_key/%s", sshKeyMountSubPath)
 			sshKeyVolume := corev1.Volume{
 				Name: sshKeyName,
 				VolumeSource: corev1.VolumeSource{
--- a/tests/kuttl/tests/dataplane-deploy-global-service-test/01-assert.yaml
+++ b/tests/kuttl/tests/dataplane-deploy-global-service-test/01-assert.yaml
@@ -150,18 +150,18 @@ spec:
   backoffLimit: 6
   extraMounts:
   - mounts:
-    - mountPath: /runner/env/ssh_key
-      name: ssh-key
-      subPath: ssh_key
+    - mountPath: /runner/env/ssh_key/ssh_key_edpm-compute-global
+      name: ssh-key-edpm-compute-global
+      subPath: ssh_key_edpm-compute-global
     - mountPath: /runner/inventory/hosts
       name: inventory
       subPath: inventory
     volumes:
-    - name: ssh-key
+    - name: ssh-key-edpm-compute-global
       secret:
         items:
         - key: ssh-privatekey
-          path: ssh_key
+          path: ssh_key_edpm-compute-global
         secretName: dataplane-ansible-ssh-private-key-secret
     - name: inventory
       secret:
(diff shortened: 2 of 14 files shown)


pkg/dataplane/util/ansible_execution.go
```diff
@@ -142,18 +142,12 @@ func AnsibleExecution(
...
-				}
-				sshKeyName = "ssh-key"
-				sshKeyMountSubPath = "ssh_key"
-				sshKeyMountPath = "/runner/env/ssh_key"
+			if !service.Spec.DeployOnAllNodeSets && sshKeyNodeName != nodeSet.GetName() {
+				continue
 			}
+			sshKeyName = fmt.Sprintf("ssh-key-%s", sshKeyNodeName)
+			sshKeyMountSubPath = fmt.Sprintf("ssh_key_%s", sshKeyNodeName)
+			sshKeyMountPath = fmt.Sprintf("/runner/env/ssh_key/%s", sshKeyMountSubPath)
```
slagle, line 150: we probably shouldn't be using /runner/env/ssh_key as a directory at all, as it's meant to be a file for ansible-runner:

https://ansible.readthedocs.io/projects/runner/en/stable/intro/#env-ssh-key

thankfully ansible-runner handles this gracefully if it can't load a file, and just ignores whatever is there. that's why I suggested putting these keys in a different directory on the initial PR that added this code: https://github.com/openstack-k8s-operators/dataplane-operator/pull/624#discussion_r1491635975

Regardless I don't know what issue this is fixing because if /runner/env/ssh_key *is* a file (as in the case of a non-global service), then ansible-runner automatically adds the key there to the ssh-agent. At that point, it doesn't matter that ansible_ssh_private_key in the inventory is pointing to a non-existant file. I agree that would be good to clean up so it's more obvious what key actually gets used, I'm just not sure that this fixes the issue.

rabi: > we probably shouldn't be using /runner/env/ssh_key as a directory at all, as it's meant to be a file for ansible-runner

Agreed, but we should do that in a separate PR as we hardcode the path in kuttl asserts and would need changes to all kuttl tests and this is trying for to fix a specific issue reported downstream.

>  At that point, it doesn't matter that ansible_ssh_private_key in the inventory is pointing to a non-existant file.

Probably it matters as reported in downstream testing in https://redhat-internal.slack.com/archives/C063J9ZJ7N3/p1719482555872529 (Note: I've not checked the ansible-runner/ansible combination causing the issue), but I think this PR would fix the issue if encountered. I've added a jira too.

rabi:
`error: build error: Failed to push image: trying to reuse blob sha256:e394ea8406c7ed85ddc862f882ec77dcfd3863de3cb90fd8006238d521d22d44 at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout`

pkg/dataplane/util/ansible_execution.go
slagle: The error there:

`
"msg": "Failed to connect to the host via ssh: Warning: Permanently added '192.168.122.100' (ED25519) to the list of known hosts.\r\nno such identity: /runner/env/ssh_key/ssh_key_openstack-edpm-ipam: Not a directory\r\ncloud-admin@192.168.122.100: Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password)."
`

is because `/runner/env/ssh_key` is not correct, so it falls back to try the path defined in the inventory. The cause sounds like ci-framework keeps recreating the ssh key secret, and it no longer contains the key configured in authorized_keys on the edpm node. A way to confirm this is to check those keys in the secret and in authorized_keys and see if they match, if they don't, that's the problem.

rabi: Did you see the workaround mentioned in https://issues.redhat.com/browse/OSPRH-8057?focusedId=25012795&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25012795. That sounds like the secret in `/runner/env/ssh_key` is good, though it's good to validate if the key is good or not.

slagle: Yes, I did see that. The workaround doesn't make sense and I question if it actually works. Editing the inventory isn't going to persist, the inventory will just be regenerated on the next NodeSet reconcile which is going to keep happening during a Deployment.

the comment also says "Hopefully on the next restart...", I doubt it actually does.

slagle: What's the reasoning behind the fix is why I'm questioning this. 

The PR implies that  `/runner/env/ssh_key` can be correct, but if  `ansible_ssh_private_key_file` points to a non-existent path, then the Deployment doesn't work. If that was the case then no Deployment would ever work as things stand.

rabi: > The PR implies that `/runner/env/ssh_key` can be correct, but if `ansible_ssh_private_key_file` points to a non-existent path, then the Deployment doesn't work.  If that was the case then no Deployment would ever work as things stand.

Well, my initial suspicion as mentioned in the slack thread was that it could be some recent change in ansible-runner/ansible on why it's not working now, but I did not dig into it. This PR may not fix the issue reported, but it would ensure that we don't get misleading error like `no such identity: /runner/env/ssh_key/ssh_key_openstack-edpm-ipam: Not a directory` when the ssh key is not valid. That would at least be fixed with this PR as the end user won't be aware of the fallback and 'ansible_ssh_private_key_file' pointing to an non-existent location.

rabi:
Do we still have reasons/concerns to not merge this?

slagle:
> Do we still have reasons/concerns to not merge this?

Yes. `/runner/env/ssh_key` should always be a file, never a directory according to the ansible-runner runner input directory specification https://ansible.readthedocs.io/projects/runner/en/stable/intro/#env-ssh-key

While this PR isn't the first to treat it as a directory,  it does change it to be a directory every time, thus building upon an initial mistake.

We need to store the keys in a different location.

rabi:
> While this PR isn't the first to treat it as a directory, it does change it to be a directory every time, thus building upon an initial mistake.

OK, the current approach we use anyway looks to me like a hack without https://github.com/ansible/ansible-runner/issues/51. My intention was to not build on a mistake (which anyway does not cause any errors), but to improve the user experience avoiding the misleading error that was reported. But it's ok if we want to live with it till we change the locations.

PR #893, opened 2024-06-28
Introducing spec level validation TLS consistency

jpodivin:
Verifies that TLS settings for nodeset are consistent with those of existing control plane, if there is one and only one.

If there are multiple control planes the process will result in error, same if it isn't possible to retrieve list of control planes.

Tests are included

Original PR was opened with branch on upstream due to faulty local automation, I'm reopening it here.

### DIFF
--- a/apis/dataplane/v1beta1/openstackdataplanenodeset_types.go
+++ b/apis/dataplane/v1beta1/openstackdataplanenodeset_types.go
@@ -17,9 +17,11 @@ limitations under the License.
 package v1beta1
 
 import (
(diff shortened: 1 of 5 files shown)


fao89 (approved):

vakwetu (approved):

jpodivin:
/retest

```
502 Bad Gateway, image \"quay.io/openstack-k8s-operators/heat-operator-bundle:dfcac6c57e461339e5edce71bf8348cb74165b5f\": not found
```

PR #897, opened 2024-07-01
Use oc kustomize instead of kustomize to build the docs

fao89:
For not mixing the kustomize versions (CRDs and docs), use oc kustomize for docs


fao89:
recheck

fao89:
recheck

slagle (approved):

PR #898, opened 2024-07-01
Validate dataplane volume names

fao89:
and use kubernetes constants

[OSPRH-8063](https://issues.redhat.com//browse/OSPRH-8063)


pkg/dataplane/util/ansible_execution.go
```diff
@@ -225,6 +226,15 @@ func AnsibleExecution(
...
 		ansibleEE.Spec.Env = aeeSpec.Env
 
+		// Validate volume names
+		for _, emount := range ansibleEE.Spec.ExtraMounts {
+			for _, vol := range emount.Volumes {
+				if msgs := apimachineryvalidation.IsDNS1123Label(vol.Name); len(msgs) != 0 {
+					return fmt.Errorf("volume name %q is not a valid: %s", vol.Name, strings.Join(msgs, ";"))
+				}
+			}
+		}
```
bshephar, line 236: We should validate this with the webhook right? Rather than accepting values that we can't work with and then raising errors during the reconcile process?

bshephar (commented):
If it's not going to be valid,

pkg/dataplane/util/ansible_execution.go
fao89: just moved it to the webhook

apis/dataplane/v1beta1/common.go
```diff
@@ -102,10 +102,6 @@ type AnsibleOpts struct {
 
 // NodeSection defines the top level attributes inherited by nodes in the CR.
 type NodeSection struct {
-	// ExtraMounts containing files which can be mounted into an Ansible Execution Pod
-	// +kubebuilder:validation:Optional
-	ExtraMounts []storage.VolMounts `json:"extraMounts,omitempty"`
```
stuggi, line 107: this is changing the API? isn't it a breaking API change?

fao89: currently, our code only uses the `ExtraMounts` from `NodeTemplate`: https://github.com/openstack-k8s-operators/openstack-operator/blob/main/apis/dataplane/v1beta1/openstackdataplanenodeset_types.go#L199
We don't have any mechanism to custom the `ExtraMounts` per node, that's why I'm dropping it from the `NodeSection`

slagle (approved):

apis/dataplane/v1beta1/common.go
stuggi: ack, thanks for the explanation.

PR #899, opened 2024-07-01
Set NodeSet.Status.DeployedVersion when update service is complete

slagle:
Set NodeSet.Status.DeployedVersion when update service is complete

Changes setting NodeSet.Status.DeployedVersion to be only when a
Deployment has completed that contains a service where EDPMServiceType
== 'update'. When satisfied, the Deployment's Status.DeployedVersion
will be copied to NodeSet.Status.DeployedVersion.

Jira: https://issues.redhat.com/browse/OSPRH-8176

### DIFF
--- a/controllers/dataplane/openstackdataplanenodeset_controller.go
+++ b/controllers/dataplane/openstackdataplanenodeset_controller.go
@@ -529,9 +529,41 @@ func checkDeployment(helper *helper.Helper,
 					instance.Status.ContainerImages[k] = v
 				}
 				instance.Status.DeployedConfigHash = deployment.Status.NodeSetHashes[instance.Name]
-				instance.Status.DeployedVersion = deployment.Status.DeployedVersion
-			}
 
+				// Get list of services by name, either from ServicesOverride or
+				// the NodeSet.
+				var services []string
+				if len(deployment.Spec.ServicesOverride) != 0 {
+					services = deployment.Spec.ServicesOverride
+				} else {
+					services = instance.Spec.Services
+				}
+
+				// For each service, check if EDPMServiceType is "update", and
+				// if so, copy Deployment.Status.DeployedVersion to
+				// NodeSet.Status.DeployedVersion
+				for _, serviceName := range services {
+					service := &dataplanev1.OpenStackDataPlaneService{}
+					name := types.NamespacedName{
+						Namespace: instance.Namespace,
+						Name:      serviceName,
+					}
+					err := helper.GetClient().Get(context.Background(), name, service)
+					if err != nil {
+						helper.GetLogger().Error(err, "Unable to retrieve OpenStackDataPlaneService %v")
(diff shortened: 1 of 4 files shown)


pkg/openstack/dataplane.go
```diff
@@ -65,7 +66,65 @@ func DataplaneNodesetsDeployed(version *corev1beta1.OpenStackVersion, dataplaneN
 		if nodeset.Status.DeployedVersion != version.Spec.TargetVersion {
 			return false
 		}
+	}
+	return true
+}
 
+// DataplaneNodesetsUpdated returns true if all nodesets are deployed with the latest version
+func DataplaneNodesetsUpdated(ctx context.Context, helper *helper.Helper, version *corev1beta1.OpenStackVersion, dataplaneNodesets *dataplanev1.OpenStackDataPlaneNodeSetList) bool {
```
dprince, line 74: this will get called repeatedly during a minor update until it reconciles.  while this may work in some cases it also depends on DataplaneNodeSet to trigger a reconcile loop on OpenStackVersion (this is the only resource we are watching) and it is not as efficient as perhaps just having the OpenStackDataplaneNodeset controller do it (since it already watches Deployment anyway).

slagle: Ok, I have reworked it to set a condition on the NodeSet instead, and just updated the func here to check that condition.

pkg/openstack/dataplane.go
```diff
@@ -62,7 +62,7 @@ func DataplaneNodesetsDeployed(version *corev1beta1.OpenStackVersion, dataplaneN
 		if !nodeset.IsReady() {
 			return false
 		}
-		if nodeset.Status.DeployedVersion != version.Spec.TargetVersion {
+		if !nodeset.Status.Conditions.IsTrue(dataplanev1.NodeSetMinorUpdateReadyCondition) {
```
dprince, line 65: I think we might need to check both of these right?

slagle: The version check has already been done in the nodeset controller before the condition is set.

bshephar, line 65: The linter is complaining now because we don't use the `version *corev1beta1.OpenStackVersion` that is the first arg to the function.  We can remove it from the functions args if it's not needed any more.

slagle: Fixed.

dprince: while that may be true, this function gets executed from the openstackversion side of things which could bump targetVersion. I am trying to guard against the case where the version changes, but in the interim this condition hasn't yet been reset

slagle: This code isn't changed in the latest revision.

slagle: Reworked so that this "if" condition will work without any changes.

dprince (approved):

PR #901, opened 2024-07-02
Add configuration of --procs for ginkgo

slagle:
${PROC_CMD} is used in the call to ginkgo, but never set anywhere.


slagle:
It seems there's something in the functional tests that breaks when trying to run in parallel. I'll close this for now.

PR #903, opened 2024-07-02
Cleanup apiOverrides when a GlanceAPI is removed from ctlplane

fmount:
Currently, if a `GlanceAPI` is added to the `ctlplane` and then removed, the human operator needs to manually clean up the `Glance` section in the `ctlplane CR` and remove the generated `APIOverrides`.
In addition to that, overrides are generated regardless of the nature of the `glanceAPI` (e.g. `Edge`, that is not supposed to have any `Route` as only an `internal` Svc is provided).
This patch solves this problem by removing the (leftover) overrides for the `APIs` that are no longer present in the ctlplane CR.

Jira: [OSPRH-8072](https://issues.redhat.com//browse/OSPRH-8072)


stuggi:
looks good to me, thanks

stuggi (approved):
/lgtm

PR #908, opened 2024-07-04
Set Designate API TLS cert secret

omersch381:
This patch provides the cert secret to the Designate API TLS field.

Depends-On: openstack-k8s-operators/designate-operator#184


omersch381:
recheck

omersch381:
oops, manual rebase got really wrong. I will create a new PR.

PR #915, opened 2024-07-08
Pass context based on signal handlers to managers

rabi:
And don't use context.Background() which is an empty context when context is available.


fao89 (commented):
+2

abays (approved):
/lgtm

PR #919, opened 2024-07-09
[18.0.0-proposed]Use branch specific renovate config

gibizer:
It depends on https://github.com/openstack-k8s-operators/renovate-config/pull/17 (but I cannot add a Depends-On as that hurts RDO zuul)
Related: [OSPRH-8355](https://issues.redhat.com//browse/OSPRH-8355)


gibizer:
we won't try this on 18.0.0-proposed

PR #920, opened 2024-07-09
Enable renovate on 18.0.0-proposed

gibizer:
I uses  openstack-k8s-operators/renovate-config#17 (but cannot add Depends-On as that hurts RDO zuul)
Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/919 
Related: [OSPRH-8355](https://issues.redhat.com//browse/OSPRH-8355)


gibizer:
We won't try this on 18.0.0-propsed

PR #931, opened 2024-07-15
Adding CI target for functional tests and related vars

jpodivin:
Dependency of https://github.com/openshift/release/pull/53105

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,11 @@ SHELL = /usr/bin/env bash -o pipefail
 # Extra vars which will be passed to the Docker-build
 DOCKER_BUILD_ARGS ?=
 
+# Ginkgo args
+GINKGO_REPEAT = --repeat 5
+GINKGO_RAND_ALL = --randomize-all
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -170,6 +175,10 @@ golangci-lint:
 .PHONY: test
 test: manifests generate gowork fmt vet envtest ginkgo ginkgo-run ## Run ginkgo tests with dependencies.
 
+.PHONY: test-ci
+test-ci: GINKGO_ARGS = $(GINKGO_CI_ARGS)
```
stuggi, line 179: do we need a new target for this? can't we just set different  GINKGO_ARGS in the CI job?

jpodivin: That was my original suggestion. But it was pointed out that we would have to submit PR to release repo every time we would want to change anything about the way ginkgo runs, say number of repetitions.

stuggi: ah ok, so maybe we should get some review of those who requested it?

jpodivin: They are on the list.

PR #934, opened 2024-07-15
[rabbitmq] remove CommonName from tls cert

stuggi:
The CommonName has a max length of 64 bytes. Since the CommonName value is already in the DnsNames of the certificate it is not required.

Jira: https://issues.redhat.com/browse/OSPRH-8652


stuggi:
manually tested in a fips enabled env that rabbitmq properly starts and services connect.

lmiccini:
seems to be fine for FIPS so it should be OK.

vakwetu (commented):
+2 pending CI

PR #935, opened 2024-07-15
Add TLS cert to nova service

fao89:


bogdando (commented):
I believe this should become a default, right

config/services/dataplane_v1beta1_openstackdataplaneservice_nova.yaml
```diff
@@ -12,6 +12,14 @@ spec:
         name: nova-extra-config
         optional: true
   playbook: osp.edpm.nova
+  tlsCerts:
```
SeanMooney, line 15: nova does not requrie a tls cert on the edpm node

it has no API or similar that would use it.

it just need the ca bundle to be able to talk to other services and trust there cert

SeanMooney (changes requested):

config/services/dataplane_v1beta1_openstackdataplaneservice_nova.yaml
bogdando: I am confused now. Whata should be there in https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc?plain=1#L143 then?

bogdando: I believe we still need this PR, maybe adding only
```
tlsCerts:
  caCerts: combined-ca-bundle
  edpmServiceType: nova
``` ?

SeanMooney: caCerts: combined-ca-bundle is already incldued 

it is not a sub key of the tlsCerts section

https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/crd/bases/dataplane.openstack.org_openstackdataplaneservices.yaml#L37

edpmServiceType: nova is required because we are currently relying on the name of the service to set the default value of edpmServiceType when it not set in the OpenStackDataPlaneService cr

SeanMooney: edpmServiceType is also not a subkey of tlsCerts its directly under spec
https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/crd/bases/dataplane.openstack.org_openstackdataplaneservices.yaml#L70

SeanMooney: this is now addressed by https://github.com/openstack-k8s-operators/openstack-operator/pull/943

PR #936, opened 2024-07-15
[nova] use short service name for CommonName vencrypt cert

stuggi:
The CommonName has a max length of 64 bytes.

From https://docs.openstack.org/nova/latest/admin/remote-console-access.html#vnc-proxy-security

~~~
An x509 certificate to be presented to the VNC server. While libvirt/QEMU will not currently do any validation of the CommonName field, future versions will allow for setting up access controls based on the CommonName. The CommonName field should match the primary hostname of the controller node.
~~~

Related: https://issues.redhat.com/browse/OSPRH-8652


stuggi:
pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout

olliewalsh (approved):
/lgtm

PR #937, opened 2024-07-15
Mount under /var/lib/openstack/configs using EDPMServiceType

slagle:
Instead of using the service name when creating the mount path under
/var/lib/openstack/configs, use the value of EDPMServiceType, which
defaults to the service name, but would be overridden for custom
services. The roles in edpm-ansible can then rely on a consistent path
that doesn't change when using a custom service.

Jira: https://issues.redhat.com/browse/OSPRH-8651


krcmarik (approved):
lgtm

fao89 (approved):

PR #949, opened 2024-07-17
Fix runtime error when CtlplaneGateway is nil

fao89:
closes [OSPRH-8645](https://issues.redhat.com//browse/OSPRH-8645)

### DIFF
--- a/pkg/dataplane/baremetal.go
+++ b/pkg/dataplane/baremetal.go
@@ -71,6 +71,9 @@ func DeployBaremetalSet(
 						}
 						ipPrefix, _ := ipNet.Mask.Size()
 						instanceSpec.CtlPlaneIP = fmt.Sprintf("%s/%d", res.Address, ipPrefix)
+						if res.Gateway == nil {
+							return fmt.Errorf("%s gateway is missing", CtlPlaneNetwork)
+						}
 						baremetalSet.Spec.CtlplaneGateway = *res.Gateway
 						baremetalSet.Spec.BootstrapDNS = dnsAddresses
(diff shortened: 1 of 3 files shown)


pkg/dataplane/baremetal.go
```diff
@@ -71,7 +71,9 @@ func DeployBaremetalSet(
 						}
 						ipPrefix, _ := ipNet.Mask.Size()
 						instanceSpec.CtlPlaneIP = fmt.Sprintf("%s/%d", res.Address, ipPrefix)
-						baremetalSet.Spec.CtlplaneGateway = *res.Gateway
+						if res.Gateway != nil {
```
rabi, line 74: What is the use case for not having ctlplane gateway in the netconfig?  Do they use static routes for controlplane? If they have gateway on external network, it won't work because we download stuff (download-cache) before configuring networks on edpm nodes. If it's something that the user missed, then we should throw an error here...

jpodivin: I'm not sure about it either. Could be useful for air-gaped environments maybe?

In any case we should log the fact that there is no gateway set. Even if it's not going to be an error.

jpodivin (approved):

pkg/dataplane/baremetal.go
jpodivin: Way better, now we don't have to worry about it at all.

rabi: > yeah, without the gateway, it fails:

That's because you've `defaultRoute: true` for ctlplane network in ospdns. I was wondering if there is an use-case with gateway only on external network that we need to address. That would not work unless we configure network on nodes before downloading container images/packages.

PR #953, opened 2024-07-18
Add InventorySecretName to NodeSet.Status

fao89:
closes OSPRH-6094


bshephar (approved):
Ah nice one. This is a nicer user experience for sure compared to `oc get secret | grep blah`.

PR #955, opened 2024-07-19
[validation] check glanceAPI names are valid

stuggi:
The glanceAPI controller creates StatefulSet for glanceapi to run. This adds a StatefulSet pod's label
"controller-revision-hash": "<statefulset_name>-<hash>" to the pod.
The kubernetes label is restricted under 63 char and the revision hash is an int32, 10 chars + the hyphen. Which results in a default statefulset max len of 52 chars. The statefulset name also contain the glance name and the glanceAPI type + 2 hyphens. So the max len also need to be reduced bye the length of those.

Also the name of the created rabbitmq instance must match a lowercase RFC 1123.

Depends-On: openstack-k8s-operators/lib-common#532
Depends-On: https://github.com/openstack-k8s-operators/glance-operator/pull/587

Jira: https://issues.redhat.com/browse/OSPRH-8063

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_types.go
+++ b/apis/core/v1beta1/openstackcontrolplane_types.go
@@ -17,6 +17,8 @@ limitations under the License.
 package v1beta1
 
 import (
+	"fmt"
+
 	barbicanv1 "github.com/openstack-k8s-operators/barbican-operator/api/v1beta1"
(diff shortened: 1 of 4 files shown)


abays (approved):
/lgtm

gibizer (approved):

stuggi:
rebased

stuggi:
once more a rebase

PR #957, opened 2024-07-19
[validation] check cinderVolumes names are valid

stuggi:
CinderVolume name is <cinder name>-volume-<volume name>
The CinderVolume controller creates StatefulSet for volume service to run.
This adds a StatefulSet pod's label
"controller-revision-hash": "<statefulset_name>-<hash>"
to the pod.
The kubernetes label is restricted under 63 char and the revision
hash is an int32, 10 chars + the hyphen. Which results in a default
statefulset max len of 52 chars. The statefulset name also
contain the cinder name and -volume-. So the max len also need to
be reduced bye the length of those.
    
Also the name of the created cinderVolume name match a lowercase RFC 1123.

Depends-On: openstack-k8s-operators/lib-common#532
Depends-On: https://github.com/openstack-k8s-operators/cinder-operator/pull/420
Jira: https://issues.redhat.com/browse/OSPRH-8063

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_types.go
+++ b/apis/core/v1beta1/openstackcontrolplane_types.go
@@ -67,6 +67,8 @@ const (
 
 	// GlanceName - Default Glance name
 	GlanceName = "glance"
+	// CinderName - Default Cinder name
+	CinderName = "cinder"
 )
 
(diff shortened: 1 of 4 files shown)


stuggi:
need to rebase after https://github.com/openstack-k8s-operators/openstack-operator/pull/955

abays (approved):
/lgtm

PR #958, opened 2024-07-19
Add labels and defaults to DataPlaneService from EnsureServices

slagle:
Fixes a reconcile loop by not modifying existing DataPlaneServices in
the CreateOrPatch call from EnsureService. If the webhook modifies the
service instead, that seems to trigger another NodeSet reconcile which
creates a loop.

Jira: [OSPRH-8811](https://issues.redhat.com//browse/OSPRH-8811)

### DIFF
--- a/controllers/dataplane/openstackdataplanenodeset_controller.go
+++ b/controllers/dataplane/openstackdataplanenodeset_controller.go
@@ -716,6 +716,7 @@ func (r *OpenStackDataPlaneNodeSetReconciler) deploymentWatcherFn(
 	ctx context.Context, //revive:disable-line
 	obj client.Object,
 ) []reconcile.Request {
+	Log := r.GetLogger(ctx)
 	namespace := obj.GetNamespace()
 	deployment := obj.(*dataplanev1.OpenStackDataPlaneDeployment)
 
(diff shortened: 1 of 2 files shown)


bshephar (approved):

rabi:
>  that seems to trigger another NodeSet reconcile which creates a loop.

Hmm.. Not sure why would service update/patch would trigger a nodeset reconciliation as nodeset controller neither owns or watches services? There is something strange going on for sure, if deleting the MutatingWebhookConfiguration stops the reconcile loop as the webhook configuration would be created again after deletion.

gibizer:
recheck

rabi:
> I deleted the webhook from the CSV, so it does not come back, unless you delete/recreate the Subscription.

Deleting the webhook with `oc delete` would bring it back and it also seems to stop the reconcile loop in spite of the webhook re-appearing. Also, if you patch a service manually (not in the nodeset controller as we do), mutating webhook is called and the fields are defaulted, but it does not do the nodeset reconcile and hence the loop.

jpodivin (approved):

rabi:
> I agree this is an unexplained behavior as I didn't think a webhook should trigger another reconcile loop, but it seems to based on the observations.

https://github.com/openstack-k8s-operators/openstack-operator/pull/968

PR #961, opened 2024-07-22
Allow firewall-cmd to fail gracefully

slagle:
firewalld is not always installed and running on the machine where "make
run-with-webhooks" might be used.


jpodivin:
` error: build error: Failed to push image: trying to reuse blob sha256:e394ea8406c7ed85ddc862f882ec77dcfd3863de3cb90fd8006238d521d22d44 at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout`

PR #962, opened 2024-07-22
Add edpm_service_types

slagle:
The list of service types is added as an ansible variable to the
inventory as edpm_service_types. This will be used instead of
edpm_services since service names could be any customer values.

edpm-ansible roles such as edpm_update, need a list of consistent
service identifiers (types instead of names), so it knows which other
roles to include for update tasks.

Jira: https://issues.redhat.com/browse/OSPRH-8107

### DIFF
--- a/pkg/dataplane/inventory.go
+++ b/pkg/dataplane/inventory.go
@@ -105,6 +105,19 @@ func GenerateNodeSetInventory(ctx context.Context, helper *helper.Helper,
 	// add services list
 	nodeSetGroup.Vars["edpm_services"] = instance.Spec.Services
 
+	// add service types list
+	serviceTypes := []string{}
+	for _, serviceName := range instance.Spec.Services {
+		service, err := GetService(ctx, helper, serviceName)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


fao89 (commented):
should we also get the serviceOverride list?
https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/dataplane/util/ansible_execution.go#L127-L135

rebtoor (approved):

slagle:
> should we also get the serviceOverride list? https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/dataplane/util/ansible_execution.go#L127-L135

do you mean something like `edpm_service_override_types`? I don't have a specific use case for that right now. but we can add it in a follow up.

fao89:
> > should we also get the serviceOverride list? https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/dataplane/util/ansible_execution.go#L127-L135
> 
> do you mean something like `edpm_service_override_types`? I don't have a specific use case for that right now. but we can add it in a follow up.

yeah, I was curious if you had a use case for it

fao89:
recheck

fao89 (approved):

jpodivin (approved):

PR #966, opened 2024-07-23
Handle errors in pin-bundle-images

karelyatin:
Even if there are errors in pin-bundle-image a broken index image get build and pushed to registry. Handle such errors to avoid push for broken index image.

Resolves: [OSPRH-5354](https://issues.redhat.com//browse/OSPRH-5354)


fao89 (approved):

jpodivin (approved):
I like this. Exposing the errors rather than hiding them.

PR #967, opened 2024-07-23
[OSPRH-8473] Warning: unknown field "spec.telemetry.template.logging.…

jlarriba:
…network"

### DIFF
--- a/config/samples/base/openstackcontrolplane/core_v1beta1_openstackcontrolplane.yaml
+++ b/config/samples/base/openstackcontrolplane/core_v1beta1_openstackcontrolplane.yaml
@@ -160,7 +160,6 @@ spec:
         secret: osp-secret
       logging:
         enabled: false
-        network: internalapi
         ipaddr: 172.17.0.80
         port: 10514
         cloNamespace: openshift-logging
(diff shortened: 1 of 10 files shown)


snecklifter:
@jlarriba rather than just remove the line, would it be possible to replace it with an annotation example?

jlarriba:
@snecklifter I dont think I like that idea: the annotations are configured exactly identical in all the operators (we are all reusing the same framework) and none of the rest have them in these examples. I think the idea is to try to keep the yaml example files not too verbose. The annotation usage could be documented in the official documentation, in a common section.

abays:
> Sorry, i have realized that there are actually examples where they have them. We could do it to show that there is that possibility. I am going definitively to add the annotations field to some of the samples.

It looks good to me.  Should we wait for this additional change before approving?

jlarriba:
> > Sorry, i have realized that there are actually examples where they have them. We could do it to show that there is that possibility. I am going definitively to add the annotations field to some of the samples.
> 
> It looks good to me. Should we wait for this additional change before approving?

Yes please

jlarriba:
@abays now its ready

abays (approved):
/lgtm

jlarriba:
recheck

jlarriba:
recheck

jlarriba:
Kuttl fix: https://github.com/openstack-k8s-operators/telemetry-operator/pull/451

PR #968, opened 2024-07-23
Don't reconcile for status updates

rabi:
We do nodeset/deployment status update very frequently and reconciling them again after every status update is expensive. This also sometimes leads to infinite loops and results in very verbose controller logs.

Also removes the changes to duplicate multating webhook logic when calling CreateOrPatch() from EnsureServices().

### DIFF
--- a/controllers/dataplane/openstackdataplanedeployment_controller.go
+++ b/controllers/dataplane/openstackdataplanedeployment_controller.go
@@ -28,8 +28,10 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"github.com/go-logr/logr"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
@@ -426,7 +428,11 @@ func (r *OpenStackDataPlaneDeploymentReconciler) setHashes(
 // SetupWithManager sets up the controller with the Manager.
 func (r *OpenStackDataPlaneDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&dataplanev1.OpenStackDataPlaneDeployment{}).
+		For(&dataplanev1.OpenStackDataPlaneDeployment{},
+			builder.WithPredicates(predicate.Or(
+				predicate.GenerationChangedPredicate{},
+				predicate.AnnotationChangedPredicate{},
+				predicate.LabelChangedPredicate{}))).
 		Owns(&ansibleeev1.OpenStackAnsibleEE{}).
 		Complete(r)
 }
--- a/controllers/dataplane/openstackdataplanenodeset_controller.go
+++ b/controllers/dataplane/openstackdataplanenodeset_controller.go
@@ -628,7 +628,11 @@ func (r *OpenStackDataPlaneNodeSetReconciler) SetupWithManager(
 		return err
 	}
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&dataplanev1.OpenStackDataPlaneNodeSet{}).
+		For(&dataplanev1.OpenStackDataPlaneNodeSet{},
+			builder.WithPredicates(predicate.Or(
+				predicate.GenerationChangedPredicate{},
+				predicate.AnnotationChangedPredicate{},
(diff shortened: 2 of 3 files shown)


abays:
I tested this with VA1 and the deployment succeeded, and the OpenStack operator logs did not endlessly cycle with reconciliation loops.

:+1:

controllers/dataplane/openstackdataplanedeployment_controller.go
```diff
@@ -426,7 +428,11 @@ func (r *OpenStackDataPlaneDeploymentReconciler) setHashes(
 // SetupWithManager sets up the controller with the Manager.
 func (r *OpenStackDataPlaneDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&dataplanev1.OpenStackDataPlaneDeployment{}).
+		For(&dataplanev1.OpenStackDataPlaneDeployment{},
+			builder.WithPredicates(predicate.Or(
+				predicate.GenerationChangedPredicate{},
+				predicate.AnnotationChangedPredicate{},
+				predicate.LabelChangedPredicate{}))).
```
gibizer, line 435: If nobody else sets the status of the Deployment just the deployment controller itself then not reconciling on status change is OK. (In k8s in general it is allowed to update the status of another resources)

GeneratrionChangePredicate covers the Spec change, and the the other two covers for label and annotation change. Do we need to cover other metadata fields changes? E.g. finalizers, deletionTimestamp?

controllers/dataplane/openstackdataplanenodeset_controller.go
```diff
@@ -628,7 +628,11 @@ func (r *OpenStackDataPlaneNodeSetReconciler) SetupWithManager(
 		return err
 	}
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&dataplanev1.OpenStackDataPlaneNodeSet{}).
+		For(&dataplanev1.OpenStackDataPlaneNodeSet{},
+			builder.WithPredicates(predicate.Or(
+				predicate.GenerationChangedPredicate{},
+				predicate.AnnotationChangedPredicate{},
+				predicate.LabelChangedPredicate{}))).
```
gibizer, line 635: Same questions as for the deployment resource

controllers/dataplane/openstackdataplanedeployment_controller.go
rabi: > If nobody else sets the status of the Deployment just the deployment controller itself then not reconciling on status change is OK

No, we don't set status from another controller, which is a bad design though allowed in k8s. 

> Do we need to cover other metadata fields changes? E.g. finalizers, deletionTimestamp?

We don't have finalizers set on the nodeset/deployment resources.  As per the question on reconciling when deletionTimestamp changes, firstly there is no delete reconcile for these CRs and I don't know if there is a need to reconcile when deletionTimestamp changes, as it's for k8s to garbage collect the resource once the finalizers are removed if there is deletionTimestamp.

rabi: Also what's the scenario you're thinking of that would require reconciliation when finalizers are changed on a CR?

gibizer: To be able to reconcile on delete the controller needs to react on the deletionTimestamp change as that is the only changed field from k8s when the delete is requested. But you said that there is no need to reconcile on delete in these resources so this is not a problem.

Also reconciling on finalizer changes is a moot point if we never put finalizers on these resources.

gibizer: I vaguely remember that we need to look at finalizers during cleanup not to destroy backends while other resources finalizers are still present. In this scenario reacting on a finalizer removal is needed to do the last reconcile on deletion to trigger.

gibizer: I think this case was that I remembered about reconciling on finalizer change https://github.com/openstack-k8s-operators/mariadb-operator/blob/8b86dae4532b4bc9239c4da6425c009b60e24ed6/controllers/mariadbaccount_controller.go#L413

rabi: oh wait, I think deletionTimeStamp changes generation https://github.com/kubernetes/kubernetes/issues/24935

rabi: It seems that's also true for finalizers as well from some references, though I'm still to validate that in code.

gibizer: Thanks for the link. That eas new to me that k8s bumps generation for deletedTimestamp.

gibizer: I think all my question is answered here and none of those discussion lead to discovering a real error scenario. so for me this looks OK.

abays (approved):
/lgtm

PR #972, opened 2024-07-25
Use ansibleEE library instead of AnsibleEE v1 CR

bshephar:
This change swaps out the use of the AnsibleEE custom resource for a generic Go library implementation. This enables us to stop using the unnecessary OpenStackAnsibleEE abstraction that we currently have around Kubernetes Jobs for AnsibleEE executions.

Jira: https://issues.redhat.com/browse/OSPRH-8926

### DIFF
--- a/controllers/dataplane/openstackdataplanenodeset_controller.go
+++ b/controllers/dataplane/openstackdataplanenodeset_controller.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/go-playground/validator/v10"
 	"golang.org/x/exp/slices"
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
@@ -48,7 +49,6 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/serviceaccount"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
-	ansibleeev1 "github.com/openstack-k8s-operators/openstack-ansibleee-operator/api/v1beta1"
 	baremetalv1 "github.com/openstack-k8s-operators/openstack-baremetal-operator/api/v1beta1"
 	openstackv1 "github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1"
 	dataplanev1 "github.com/openstack-k8s-operators/openstack-operator/apis/dataplane/v1beta1"
@@ -622,7 +622,7 @@ func (r *OpenStackDataPlaneNodeSetReconciler) SetupWithManager(
 				predicate.GenerationChangedPredicate{},
 				predicate.AnnotationChangedPredicate{},
 				predicate.LabelChangedPredicate{}))).
-		Owns(&ansibleeev1.OpenStackAnsibleEE{}).
+		Owns(&batchv1.Job{}).
 		Owns(&baremetalv1.OpenStackBaremetalSet{}).
 		Owns(&infranetworkv1.IPSet{}).
 		Owns(&infranetworkv1.DNSData{}).
--- a/pkg/dataplane/deployment.go
+++ b/pkg/dataplane/deployment.go
@@ -26,6 +26,9 @@ import (
 	"sort"
 	"strconv"
 
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+
 	slices "golang.org/x/exp/slices"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
@@ -37,11 +40,9 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	"github.com/openstack-k8s-operators/lib-common/modules/storage"
-	ansibleeev1 "github.com/openstack-k8s-operators/openstack-ansibleee-operator/api/v1beta1"
 	openstackv1 "github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1"
 	dataplanev1 "github.com/openstack-k8s-operators/openstack-operator/apis/dataplane/v1beta1"
 	dataplaneutil "github.com/openstack-k8s-operators/openstack-operator/pkg/dataplane/util"
-	corev1 "k8s.io/api/core/v1"
 )
 
 // Deployer defines a data structure with all of the relevant objects required for a full deployment.
@@ -197,13 +198,13 @@ func (d *Deployer) ConditionalDeploy(
 	}
 
 	if nsConditions.IsFalse(readyCondition) {
-		var ansibleEE *ansibleeev1.OpenStackAnsibleEE
+		var ansibleEE *batchv1.Job
 		_, labelSelector := dataplaneutil.GetAnsibleExecutionNameAndLabels(&foundService, d.Deployment.Name, d.NodeSet.Name)
 		ansibleEE, err = dataplaneutil.GetAnsibleExecution(d.Ctx, d.Helper, d.Deployment, labelSelector)
 		if err != nil {
 			// Return nil if we don't have AnsibleEE available yet
 			if k8s_errors.IsNotFound(err) {
-				log.Info(fmt.Sprintf("%s OpenStackAnsibleEE not yet found", readyCondition))
+				log.Info(fmt.Sprintf("%s AnsibleEE job is not yet found", readyCondition))
 				return nil
 			}
 			log.Error(err, fmt.Sprintf("Error getting ansibleEE job for %s", deployName))
@@ -215,15 +216,15 @@ func (d *Deployer) ConditionalDeploy(
 				err.Error()))
 		}
 
-		if ansibleEE.Status.JobStatus == ansibleeev1.JobStatusSucceeded {
+		if ansibleEE.Status.Succeeded > 0 {
 			log.Info(fmt.Sprintf("Condition %s ready", readyCondition))
 			nsConditions.Set(condition.TrueCondition(
 				readyCondition,
 				readyMessage))
 		}
 
-		if ansibleEE.Status.JobStatus == ansibleeev1.JobStatusRunning || ansibleEE.Status.JobStatus == ansibleeev1.JobStatusPending {
-			log.Info(fmt.Sprintf("AnsibleEE job is not yet completed: Execution: %s, Status: %s", ansibleEE.Name, ansibleEE.Status.JobStatus))
+		if ansibleEE.Status.Active > 0 {
+			log.Info(fmt.Sprintf("AnsibleEE job is not yet completed: Execution: %s, Active pods: %d", ansibleEE.Name, ansibleEE.Status.Active))
 			nsConditions.Set(condition.FalseCondition(
 				readyCondition,
 				condition.RequestedReason,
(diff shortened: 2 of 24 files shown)


pkg/dataplane/util/ansible_execution.go
```diff
@@ -61,181 +58,165 @@ func AnsibleExecution(
...
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      executionName,
-				Namespace: deployment.GetNamespace(),
-				Labels:    labels,
-			},
-		}
+	// ansibleEE, err := GetAnsibleExecution(ctx, helper, deployment, labels)
+	//if err != nil && !k8serrors.IsNotFound(err) {
+	//	return err
+	//}
```
bshephar, line 64: Need to fix this before making this PR active.

bshephar:
Looks like an issue with the `GetAnsibleExecution()`:
```
2024-07-25T03:33:09.611Z        INFO    Controllers.OpenStackDataPlaneDeployment        ServiceRepoSetupDeploymentReady OpenStackAnsibleEE not yet found        {"controller": "openstackdataplanedeployment", "controllerGroup": "dataplane.openstack.org", "controllerKind": "OpenStackDataPlaneDeployment", "OpenStackDataPlaneDeployment": {"name":"edpm-deployment","namespace":"openstack"}, "namespace": "openstack", "name": "edpm-deployment", "reconcileID": "3c90b7e8-3a39-4224-9875-897d4ed3db08"}
```

go.mod
```diff
@@ -130,3 +130,5 @@ replace github.com/openshift/api => github.com/openshift/api v0.0.0-202304141430
 
 // custom RabbitmqClusterSpecCore for OpenStackControlplane (v2.6.0_patches_tag)
 replace github.com/rabbitmq/cluster-operator/v2 => github.com/openstack-k8s-operators/rabbitmq-cluster-operator/v2 v2.6.1-0.20240719064129-c0201810c8f5 //allow-merging
+
+replace github.com/openstack-k8s-operators/lib-common/modules/ansible => github.com/bshephar/lib-common/modules/ansible v0.0.0-20240725003125-09533a78d1b6 //allow-merging
```
bshephar, line 134: Remove before marking this PR active.

pkg/dataplane/util/ansible_execution.go
```diff
@@ -58,190 +56,189 @@ func AnsibleExecution(
...
 	var sshKeyMountSubPath string
 
+	const jobRestartPolicy string = "OnFailure"
+
 	ansibleEEMounts := storage.VolMounts{}
 
 	executionName, labels := GetAnsibleExecutionNameAndLabels(service, deployment.GetName(), nodeSet.GetName())
-	ansibleEE, err := GetAnsibleExecution(ctx, helper, deployment, labels)
+
+	existingAnsibleEE, err := GetAnsibleExecution(context.TODO(), helper, deployment, labels)
```
bshephar, line 65: Need to understand whether we actually want to re-use this Job, or whether we just check if it had succeeded. If it HAS succeeded, then we exit early on line 71. All other cases, we will be creating a new job. So I don't think we need to re-use this object in failure cases.

bshephar:
Kuttl tests left to go on this PR. Removed the draft from it so that all CI jobs would run and we can get a baseline for how it's looking.

pkg/dataplane/util/ansible_execution.go
```diff
@@ -58,190 +56,191 @@ func AnsibleExecution(
...
```
jpodivin, line 65: Why switch to `TODO` ?

pkg/dataplane/service.go
```diff
@@ -47,15 +47,13 @@ type ServiceYAML struct {
 // DeployService service deployment
 func (d *Deployer) DeployService(foundService dataplanev1.OpenStackDataPlaneService) error {
 	err := dataplaneutil.AnsibleExecution(
-		d.Ctx,
```
jpodivin, line 50: Why are you removing context across these changes?

jpodivin (commented):
`+kubebuilder:rbac:groups=ansibleee.openstack.org,resources=openstackansibleees,verbs=get;list;watch;create;update;patch;delete`
This isn't necessary anymore.

pkg/dataplane/service.go
bshephar: True, not sure why I did that. Added it back, and changed the AnsibleExecution() code back to using the provided context instead of TODO()

pkg/dataplane/util/ansible_execution.go
bshephar: Yeah, changed it back and added ctx back to the function signature. Thanks

bshephar:
> `+kubebuilder:rbac:groups=ansibleee.openstack.org,resources=openstackansibleees,verbs=get;list;watch;create;update;patch;delete` This isn't necessary anymore.

Very true, I removed this as well.

pkg/dataplane/util/ansible_execution.go
```diff
@@ -58,190 +57,191 @@ func AnsibleExecution(
...
-			return err
-		}
+	aeeJob, err := ansibleEE.JobForOpenStackAnsibleEE(helper)
+	if err != nil {
+		return err
+	}
 
-		return nil
+	// CreateOrPatch the AnsibleExecution Job
+	_, err = controllerutil.CreateOrPatch(context.TODO(), helper.GetClient(), aeeJob, func() error {
```
jpodivin, line 238: You can restore context here as well.

go.mod
jpodivin: Shouldnt this point to the https://github.com/openstack-k8s-operators/lib-common/pull/536/commits/ef2c08297fa11cf6c5b9cf71ebf1a86b48eb1204 ?

bshephar:
For those reviewing and follow this PR. Commit #1 has been updated now to use the recently merged ansibleee functions from:
https://github.com/openstack-k8s-operators/openstack-operator/pull/1004

The update there removes the `replaces` in `go.mod` that previously leveraged my personal `lib-common` fork.

I still need to fix all of the Kuttl tests. So far I have just worked on perfecting the jobs in: `tests/kuttl/tests/dataplane-deploy-global-service-test/01-assert.yaml`. Some idiosyncrasies between `batchv1.Job` and the now removed `OpenStackAnsibleEE` CR. Once I get them right in that job, I'll update the rest.

bshephar:
This was missing handling for NetworkAttachment so we don't end up with the job pods connected to the multus network:
https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/blob/main/controllers/openstack_ansibleee_controller.go#L172

Probably need to write some functional tests to catch that and validate the multus annotation is correct. Latest patchset adds this functionality, retesting now and will look at functional tests for the annotation.

bshephar:
Eh, job is being created, but I named it wrong in the kuttl test:
```
2024-08-20T05:37:09.278Z	INFO	Controllers.OpenStackDataPlaneDeployment	AnsibleEE job is not yet completed: Execution: custom-global-service-edpm-compute-global, Active pods: 1	{"controller": "openstackdataplanedeployment", "controllerGroup": "dataplane.openstack.org", "controllerKind": "OpenStackDataPlaneDeployment", "OpenStackDataPlaneDeployment": {"name":"edpm-compute-global","namespace":"openstack-kuttl-tests"}, "namespace": "openstack-kuttl-tests", "name": "edpm-compute-global", "reconcileID": "0eb3000b-89f6-40f7-bc55-bb1054a8ef7d"}

```

bshephar:
Ok, should just be the tls-data test now

jpodivin:
/recheck

node failure

jpodivin:
recheck

bshephar:
recheck

pkg/dataplane/deployment.go
```diff
@@ -215,34 +216,39 @@ func (d *Deployer) ConditionalDeploy(
 				err.Error()))
 		}
 
-		if ansibleEE.Status.JobStatus == ansibleeev1.JobStatusSucceeded {
+		if ansibleEE.Status.Succeeded != 0 {
```
rabi, line 219: Maybe use > 0 here

pkg/dataplane/util/ansible_execution.go
```diff
@@ -58,190 +57,196 @@ func AnsibleExecution(
...
-	ansibleEE, err := GetAnsibleExecution(ctx, helper, deployment, labels)
+
+	existingAnsibleEE, err := GetAnsibleExecution(ctx, helper, deployment, labels)
 	if err != nil && !k8serrors.IsNotFound(err) {
 		return err
 	}
 
 	// Don't patch and re-run jobs if the job status is already completed.
-	if ansibleEE != nil && ansibleEE.Status.JobStatus == ansibleeev1.JobStatusSucceeded {
+	if existingAnsibleEE != nil && existingAnsibleEE.Status.Succeeded > 1 {
```
rabi, line 72: existingAnsibleEE.Status.Succeeded > 0 ?

pkg/dataplane/deployment.go
```diff
@@ -215,34 +216,39 @@ func (d *Deployer) ConditionalDeploy(
...
 			log.Info(fmt.Sprintf("Condition %s ready", readyCondition))
 			nsConditions.Set(condition.TrueCondition(
 				readyCondition,
 				readyMessage))
 		}
 
-		if ansibleEE.Status.JobStatus == ansibleeev1.JobStatusRunning || ansibleEE.Status.JobStatus == ansibleeev1.JobStatusPending {
-			log.Info(fmt.Sprintf("AnsibleEE job is not yet completed: Execution: %s, Status: %s", ansibleEE.Name, ansibleEE.Status.JobStatus))
+		if ansibleEE.Status.Succeeded == 0 || ansibleEE.Status.Active != 0 && ansibleEE.Status.Failed != 0 {
+			log.Info(fmt.Sprintf("AnsibleEE job is not yet completed: Execution: %s, Active pods: %d", ansibleEE.Name, ansibleEE.Status.Active))
```
rabi, line 227: What about pending job status? I guess we've to check status.active == 0 && status.succeeded == 0 && status.failed == 0?

controllers/dataplane/openstackdataplanenodeset_controller.go
```diff
@@ -565,7 +565,8 @@ func checkDeployment(ctx context.Context, helper *helper.Helper,
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *OpenStackDataPlaneNodeSetReconciler) SetupWithManager(
-	ctx context.Context, mgr ctrl.Manager) error {
```
jpodivin, line 568: I don't know how I feel about these formatting changes. In principle they are fine, but it's kind of inflating the patch and polluting the blame. Not reason to object to the PR, obviously, but it would be better to avoid that.

jpodivin (commented):
I concur with some of Rabis suggestions. Other than that I like this very much.

go.mod
bshephar: Removed this and put all of the relevant functions directly in openstack-operator.

pkg/dataplane/deployment.go
bshephar: Yeah, true.  I don't think this condition makes sense actually. Because it checks for an Active and a Failed job.

We can probably drop the failed check from here since it's captured in the next section. So this would just become

```suggestion
if ansibleEE.Status.Succeeded > 0 || ansibleEE.Status.Active > 0 {
			log.Info(fmt.Sprintf("AnsibleEE job is not yet completed: Execution: %s, Active pods: %d", ansibleEE.Name, ansibleEE.Status.Active))
}
```

Then the next section will check for if it's failed. In this instance, we just want to log when a job is still active, if it had failed then we would want to log something different.

controllers/dataplane/openstackdataplanenodeset_controller.go
bshephar: Yeah, true. I'll turn off auto formatting and restore these

pkg/dataplane/deployment.go
rabi: I think there would be scenario when all the three are zero (status.active == 0, status.succeeded == 0 and status.failed == 0) which is when the job is pending and we need to update the status with `.readyWaitingMessage`. I can't see how that is handled here.

pkg/dataplane/deployment.go
```diff
@@ -215,34 +216,39 @@ func (d *Deployer) ConditionalDeploy(
...
 				condition.RequestedReason,
 				condition.SeverityInfo,
 				readyWaitingMessage))
 		}
 
-		if ansibleEE.Status.JobStatus == ansibleeev1.JobStatusFailed {
-			errorMsg := fmt.Sprintf("execution.name %s execution.namespace %s execution.status.jobstatus: %s", ansibleEE.Name, ansibleEE.Namespace, ansibleEE.Status.JobStatus)
-			ansibleCondition := ansibleEE.Status.Conditions.Get(condition.ReadyCondition)
+		var ansibleCondition *batchv1.JobCondition
+		if ansibleEE.Status.Failed != 0 {
```
rabi, line 236: Let's use Failed > 0 to be consistent.

pkg/dataplane/deployment.go
bshephar: Right, yes I see what you mean. Like, if the `pod` for the job was in pending. I'll see what the job reports in that case to make sure we have it covered.

bshephar: Looks like even when the pod is pending, the job reports that it's `Active`:

```
❯ oc apply -f test-job.yaml; oc get job pi -o yaml | yq .status.active ; oc get po | grep pi
job.batch/pi created
1
pi-gpwd9                               0/1     Pending      0          0s
```

bshephar: FWIW, I just added a `nodeSelector` that would result in the pod pending to test that:
```
❯ yq .spec.template.spec.nodeSelector test-job.yaml
kubernetes.io/arch: aarch64
```

Only `amd64` nodes in the cluster.

pkg/dataplane/deployment.go
bshephar: Updated

rabi (commented):
+2, could be refactored further to change some of the functions from `GetAnsibleExecution` to `GetEEJob` etc. Also, remove some of the comments like https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/dataplane/util/ansible_execution.go#L41.

bshephar:
recheck

pkg/dataplane/deployment.go
bshephar: Yeah, I don't think it will ever be 0 for pending jobs, the `Active` status field for jobs represents "pending and running pods that aren't terminating":
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/batch/types.go#L524-L528

So this should always be 1 if the Job hasn't culminated in success or failure.

bshephar:
> +2, could be refactored further to change some of the functions from `GetAnsibleExecution` to `GetEEJob` etc. Also, remove some of the comments like https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/dataplane/util/ansible_execution.go#L41.

Yeah, the file could use some work. I have another branch with some refactoring here:
https://github.com/openstack-k8s-operators/openstack-operator/commit/dae8a7090dd28bb417a57d05e4ba514ef78cf9f1

Once this one merges, I'll be able to tidy it up and send that branch as a PR to clean up the rest of it. This PR was already too big and too many changes, so I left any additional refactoring for another PR.

fao89 (approved):

fao89:
hold wasn't feast enough 😓 
I wanted to double-check first if we could merge it, but the bot was faster

PR #973, opened 2024-07-25
Don't patch ansibleee if the job has completed

rabi:
This would avoid running completed jobs for any changes to AEE spec including the runner image.


rabi:
error: build error: Failed to push image: trying to reuse blob sha256:247c2d03e9487628cb6754ff5385a670df160f7bba36af8fc1f2066e461dc36e at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout

rabi:
Duh it fails with below error all the time...

fao89 (approved):

PR #975, opened 2024-07-25
Hash dataplane volume names when bigger than DNS label

fao89:
closes [OSPRH-8801](https://issues.redhat.com//browse/OSPRH-8801)

### DIFF
--- a/pkg/dataplane/deployment.go
+++ b/pkg/dataplane/deployment.go
@@ -18,6 +18,8 @@ package deployment
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"path"
 	"reflect"
@@ -313,7 +315,8 @@ func (d *Deployer) addCertMounts(
 				}
 				volumeName := GetServiceCertsSecretName(d.NodeSet, service.Name, certKey, 0)
 				if len(volumeName) > apimachineryvalidation.DNS1123LabelMaxLength {
-					volumeName = volumeName[:apimachineryvalidation.DNS1123LabelMaxLength]
+					hash := sha256.Sum224([]byte(volumeName))
+					volumeName = "cert" + hex.EncodeToString(hash[:])
 				}
 				certVolume := corev1.Volume{
 					Name: volumeName,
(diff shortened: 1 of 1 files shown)


pkg/dataplane/deployment.go
```diff
@@ -313,7 +314,8 @@ func (d *Deployer) addCertMounts(
 				}
 				volumeName := GetServiceCertsSecretName(d.NodeSet, service.Name, certKey, 0)
 				if len(volumeName) > apimachineryvalidation.DNS1123LabelMaxLength {
-					volumeName = volumeName[:apimachineryvalidation.DNS1123LabelMaxLength]
+					hash := sha256.Sum224([]byte(volumeName))
+					volumeName = string(hash[:])
```
olliewalsh, line 318: doesn't produce a hex string, `hex.EncodeToString(hash[:])` from enconding/hex works

olliewalsh: since the hex string length is 56 maybe can add a short prefix e.g "cert", "cacert", "cm", "sec"

olliewalsh (dismissed):

pkg/dataplane/deployment.go
```diff
@@ -412,8 +416,8 @@ func (d *Deployer) addServiceExtraMounts(
 		for idx, key := range keys {
 			volumeName := fmt.Sprintf("%s-%s", cm.Name, strconv.Itoa(idx))
 			if len(volumeName) > apimachineryvalidation.DNS1123LabelMaxLength {
-				limit := apimachineryvalidation.DNS1123LabelMaxLength - len(strconv.Itoa(idx))
-				volumeName = volumeName[:limit] + strconv.Itoa(idx)
+				hash := sha256.Sum224([]byte(volumeName))
+				volumeName = "cm" + hex.EncodeToString(hash[:]) + strconv.Itoa(idx)
```
olliewalsh, line 420: any reason to keep the appended index?

olliewalsh: if it's intentional I would add a `-` or it will look like it's part of the hash

bshephar (approved):

olliewalsh (approved):

PR #979, opened 2024-07-26
Restore LastTransitionTime after setting Ready Condition

rabi:
instance.Status.Conditions.MarkTrue() always sets the current time as the LastTransitionTime and this results in infinite loop as the LastTransitionTime changes after every reconcile.

I think this issue is there in many other operator controllers, but we don't see the issue as the next reconcile is probably too quick to not change the LastTransitionTime(precision for which is in seconds). The issue showed up in dataplane controllers as we reconcile a number of services in a loop for every reconcile of nodeset.

This seems to have been changed in other controllers of this operator with
https://github.com/openstack-k8s-operators/openstack-operator/pull/768.

Jira: [OSPRH-8811](https://issues.redhat.com//browse/OSPRH-8811)

### DIFF
--- a/controllers/dataplane/openstackdataplanedeployment_controller.go
+++ b/controllers/dataplane/openstackdataplanedeployment_controller.go
@@ -124,8 +124,6 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context, 
 
 	// Always patch the instance status when exiting this function so we can persist any changes.
 	defer func() { // update the Ready condition based on the sub conditions
-		condition.RestoreLastTransitionTimes(
-			&instance.Status.Conditions, savedConditions)
 		if instance.Status.Conditions.AllSubConditionIsTrue() {
 			instance.Status.Conditions.MarkTrue(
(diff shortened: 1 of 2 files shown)


fao89 (approved):

controllers/dataplane/openstackdataplanedeployment_controller.go
```diff
@@ -135,6 +133,8 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context,
 				instance.Status.Conditions.Mirror(condition.ReadyCondition))
 		}
 
+		condition.RestoreLastTransitionTimes(
```
stuggi, line 136: Its probably not critical, but should this happen before the IsUnknown layered condition. If we restore the timestamps here, we'd not get the current timestamp of the mirror and instead the initial one. Or maybe thats what you want?

stuggi: its probably ok to have the initial timestamp when it went unknown

PR #990, opened 2024-08-01
Requeue result from EnsureTLSCerts even when err != nil

slagle:
The return[1] from EnsureTLSCerts can be a non nil err, and a ctrl.Result
with a requeue. That result should be returned to force a requeue,
instead of just a ctrl.Result{}.

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/fa62144840323419fdb74594870f82d77d8b5b78/pkg/dataplane/cert.go#L178-L180

### DIFF
--- a/controllers/dataplane/openstackdataplanedeployment_controller.go
+++ b/controllers/dataplane/openstackdataplanedeployment_controller.go
@@ -229,7 +229,7 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context, 
 								condition.SeverityError,
 								condition.TLSInputErrorMessage,
 								err.Error())
-							return ctrl.Result{}, err
+							return *result, err
 						} else if (*result != ctrl.Result{}) {
 							return *result, nil // requeue here
 						}


controllers/dataplane/openstackdataplanedeployment_controller.go
```diff
@@ -216,7 +216,7 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context,
 					for certKey := range service.Spec.TLSCerts {
 						result, err := deployment.EnsureTLSCerts(ctx, helper, &nodeSet,
 							nodeSet.Status.AllHostnames, nodeSet.Status.AllIPs, service, certKey)
-						if err != nil {
+						if (err != nil) || (result != ctrl.Result{}) {
```
abays, line 219: Should this use `*result` instead since we do in L233?

slagle: yes...i think i do. it needs to be dereferenced.

slagle:
/hold

I'm now actually not sure this is needed, and I got a nil pointer dereference when trying it. It's not obvious how that happened, so I'd hold on this for the time being.

controllers/dataplane/openstackdataplanedeployment_controller.go
```diff
@@ -216,7 +216,7 @@ func (r *OpenStackDataPlaneDeploymentReconciler) Reconcile(ctx context.Context,
 					for certKey := range service.Spec.TLSCerts {
 						result, err := deployment.EnsureTLSCerts(ctx, helper, &nodeSet,
 							nodeSet.Status.AllHostnames, nodeSet.Status.AllIPs, service, certKey)
-						if err != nil {
+						if (err != nil) || (*result != ctrl.Result{}) {
```
vakwetu, line 219: I don't think we need this addition here.  If err is non-nil, we end up returning *result - which can be ctrl.Result{} or otherwise.

fao89 (commented):
when we return `requeue` we make `err==nil`:
https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/certmanager/certificate.go#L246-L248

the idea was to ignore (and requeue) only the first error, when the certs get created

stuggi:
> when we return `requeue` we make `err==nil`: https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/certmanager/certificate.go#L246-L248
> 
> the idea was to ignore (and requeue) only the first error, when the certs get created

right, and when the cert is ready, the status condition of the cert cr gets updated, which should result in the owner (one of the openstack-operator controllers) to reconcile

controllers/dataplane/openstackdataplanedeployment_controller.go
jpodivin: There may be a situation when the nil is not a ... proper nil.[0] I don't think this is the case here. As it only happens if you abuse interfaces though and we shouldn't work around it. Rather stop it at the source.

[0] https://dev.to/arxeiss/in-go-nil-is-not-equal-to-nil-sometimes-jn8

stuggi:
> > when we return `requeue` we make `err==nil`: https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/certmanager/certificate.go#L246-L248
> > the idea was to ignore (and requeue) only the first error, when the certs get created
> 
> right, and when the cert is ready, the status condition of the cert cr gets updated, which should result in the owner (one of the openstack-operator controllers) to reconcile

Having said that, I think that what is missing https://github.com/openstack-k8s-operators/openstack-operator/pull/991 . the controller is not watching the certs it is creating and therefore not reconciling if there is an update/change

stuggi:
> > > when we return `requeue` we make `err==nil`: https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/certmanager/certificate.go#L246-L248
> > > the idea was to ignore (and requeue) only the first error, when the certs get created
> > 
> > 
> > right, and when the cert is ready, the status condition of the cert cr gets updated, which should result in the owner (one of the openstack-operator controllers) to reconcile
> 
> Having said that, I think that what is missing #991 . the controller is not watching the certs it is creating and therefore not reconciling if there is an update/change

@vakwetu I am not that familiar with the dataplane controllers. with the PR I submitted the owner of the certs, nodeset, will reconcile. But shouldn't the deployment also reconcile as it is doing more follow up steps when calling EnsureTLSCerts https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/dataplane/cert.go#L183-L216 ? or should the dataplanedeployment controller watch the nodesets? or always requeue when there is an error with GetTLSNodeCert at https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/dataplane/cert.go#L180

controllers/dataplane/openstackdataplanedeployment_controller.go
slagle: @vakwetu you're right, this change wasn't needed. removed it in the latest update

slagle:
> > when we return `requeue` we make `err==nil`: https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/certmanager/certificate.go#L246-L248
> > the idea was to ignore (and requeue) only the first error, when the certs get created
> 
> right, and when the cert is ready, the status condition of the cert cr gets updated, which should result in the owner (one of the openstack-operator controllers) to reconcile

Ok, that makes sense then. A ctrl.Result with RequeueAfter will only be returned from the lib-common module when err=nil. So this PR shouldn't be needed.

slagle:
closing per discussion in comments.

PR #991, opened 2024-08-02
    nodeset to watch certs it owns

stuggi:
The dataplanedeploynet creates the certificates and set the nodeset as the owner. The nodeset should reconcile when the cert changes

### DIFF
--- a/controllers/dataplane/openstackdataplanenodeset_controller.go
+++ b/controllers/dataplane/openstackdataplanenodeset_controller.go
@@ -40,6 +40,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	certmgrv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	"github.com/go-logr/logr"
 	infranetworkv1 "github.com/openstack-k8s-operators/infra-operator/apis/network/v1beta1"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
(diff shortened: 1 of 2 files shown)


controllers/dataplane/openstackdataplanedeployment_controller.go
```diff
@@ -443,5 +444,6 @@ func (r *OpenStackDataPlaneDeploymentReconciler) SetupWithManager(mgr ctrl.Manag
 				predicate.AnnotationChangedPredicate{},
 				predicate.LabelChangedPredicate{}))).
 		Owns(&ansibleeev1.OpenStackAnsibleEE{}).
+		Owns(&certmgrv1.Certificate{}).
```
fao89, line 447: I think the ownership should be on the nodeset (but I'm not sure how the deployment would behave)

stuggi: but its the deployment which creates the the certs, so it owns it https://github.com/openstack-k8s-operators/openstack-operator/blob/main/controllers/dataplane/openstackdataplanedeployment_controller.go#L217

stuggi: or the nodeset needs a watcher based on label

stuggi: ah wait you are right. it passes the nodeset and its the owner. I'll update

abays (approved):
/lgtm

controllers/dataplane/openstackdataplanedeployment_controller.go
fao89: I'm thinking about the owner deletion, when the deployment is deleted, we don't want the certs deleted, but when the nodeset gets deleted, it makes sense to delete the certs.

fao89 (approved):

stuggi:
the nodeset is not creating or consuming the certs, so should be not required. the deployment is managing the certs. it should watch them

PR #992, opened 2024-08-02
Add watcher for cert changes

vakwetu:
The deployment manages certs and needs to watch them for changes during the initial deployment.

Jira: [OSPRH-9089](https://issues.redhat.com//browse/OSPRH-9089)

### DIFF
--- a/pkg/dataplane/cert.go
+++ b/pkg/dataplane/cert.go
@@ -146,6 +146,7 @@ func EnsureTLSCerts(ctx context.Context, helper *helper.Helper,
 				}
 			}
 		}
+		sort.Strings(ips)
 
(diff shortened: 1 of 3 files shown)


pkg/dataplane/cert.go
```diff
@@ -146,6 +146,7 @@ func EnsureTLSCerts(ctx context.Context, helper *helper.Helper,
 				}
 			}
 		}
+		sort.Strings(ips)
```
slagle, line 149: note: this is needed so that we don't keep updating the Certificate with a different list and causing a reconcile loop.

slagle (approved):
LGTM

PR #997, opened 2024-08-06
Allow users to override baremetal provisioning images

rabi:
Jira: https://issues.redhat.com/browse/OSPRH-9093

### DIFF
--- a/controllers/dataplane/openstackdataplanenodeset_controller.go
+++ b/controllers/dataplane/openstackdataplanenodeset_controller.go
@@ -368,18 +368,8 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 		instance.Status.Conditions.MarkUnknown(dataplanev1.NodeSetBareMetalProvisionReadyCondition,
 			condition.InitReason, condition.InitReason)
 
-		// Set Images
-		if containerImages.OsContainerImage != nil {
-			instance.Spec.BaremetalSetTemplate.OSContainerImageURL = *containerImages.OsContainerImage
-		}
-		if containerImages.AgentImage != nil {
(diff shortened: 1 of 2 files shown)


fao89 (approved):

controllers/dataplane/openstackdataplanenodeset_controller.go
```diff
@@ -369,13 +369,13 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 			condition.InitReason, condition.InitReason)
 
 		// Set Images
-		if containerImages.OsContainerImage != nil {
+		if containerImages.OsContainerImage != nil && instance.Spec.BaremetalSetTemplate.OSContainerImageURL == "" {
```
slagle, line 372: i'm good with the change here to allow overrding the image. related though, does the spec change here actually get saved? i thought updating the spec in the reconcile was a bit of an anti-pattern.

slagle (approved):

controllers/dataplane/openstackdataplanenodeset_controller.go
rabi: Good point.. I think it does get saved, but let's change it to update when creating baremetalset spec.

rabi:
`error: build error: Failed to push image: trying to reuse blob sha256:247c2d03e9487628cb6754ff5385a670df160f7bba36af8fc1f2066e461dc36e at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout`

fao89 (approved):

slagle (approved):

PR #999, opened 2024-08-07
Replace Ceilometer's ProxyImage with a generic one to avoid duplicati…

jlarriba:
…on of related images

### DIFF
--- a/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
+++ b/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
@@ -17481,8 +17481,6 @@ spec:
                     type: string
                   ceilometerNotificationImage:
                     type: string
-                  ceilometerProxyImage:
-                    type: string
                   ceilometerSgcoreImage:
                     type: string
                   cinderAPIImage:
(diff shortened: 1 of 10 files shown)


jlarriba:
/hold

As discussed with Dan, I am not really comfortable of having Octavia and Telemetry sharing the same image, as that could mean that the image gets updated by one of the teams unkowingly to the other, and that can cause issues.

Sharing images is this going to be a pattern for all services?

dprince:
> /hold
> 
> As discussed with Dan, I am not really comfortable of having Octavia and Telemetry sharing the same image, as that could mean that the image gets updated by one of the teams unkowingly to the other, and that can cause issues.
> 
> Sharing images is this going to be a pattern for all services?

My viewpoint on this is it is an Apache image (pretty basic service) so probably okay to share it right? And if down the road we need to isolate and have multiple settings referring to *different* container images so be it... we can.

One more detail, the HTTP image would actually get used in 3 places: baremetal-operator(the original reason we have it), telemetry-operator, and then octavia as well.

dprince:
@jlarriba code here looks good. Aside from your question I am LGTM to merge this.

dprince (approved):

PR #1000, opened 2024-08-08
[Docs] Add recomendation to use 'install-cert' in the serviceOverride

slawqo:
When serviceOverride is used in the environment with TLS enabled, it is required to run additionally install-certs service together with other services in the serviceOverride because then certificates potentially needed for the new service will be installed on edpm nodes.

This patch adds note in the documentation about this.

Related: #[OSPRH-9294](https://issues.redhat.com//browse/OSPRH-9294)


bshephar (approved):
Sounds reasonable to me.

PR #1004, opened 2024-08-14
Add functions necessary to create AnsibleEE Jobs

bshephar:
This change imports the functions from the openstack-ansibleee-operator so that we can remove the CR and instead directly render and create batchv1.Jobs without the abstraction of the OpenStackAnsibleEE custom resource and controllers.


jpodivin (approved):
I like this. Keeping it all in one repo makes sense.

PR #1006, opened 2024-08-15
Fix OpenStackClient display name

abays:
In addition to fixing the `OpenStackClient` display name, apparently we also needed to run `make manifests` to get some other updates.  I can put those in a separate PR if it's a deal-breaker.


stuggi (approved):
/lgtm

jpodivin (approved):

PR #1007, opened 2024-08-15
Swap download-cache and bootstrap services

bshephar:
For `download-cache` to work, it needs to have access to the remote container registry and a RPM repository. This necessitates that we execute subscription-manager prior to trying to download anything.

This change swaps the `bootstrap` and `download-cache` services so that `bootstrap` is executed first before trying to download anything.

This allows users to register their nodes leveraging the `edpm_bootstrap_command` variable defined on the NodeSet prior to any attempts to download content.


bshephar:
We don't need to do this after all.

PR #1009, opened 2024-08-16
Share ApacheImage with baremetal, octavia, and telemetry

dprince:
The same http image is already used for these services but is causing bundle warnings at build time. This change updates the OpenStackVersion resource so that a single Apache parameter is shared by Baremetal, Octavia, and Telemetry.

### DIFF
--- a/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
+++ b/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
@@ -17527,6 +17527,8 @@ spec:
                     type: string
                   ceilometerNotificationImage:
                     type: string
+                  ceilometerProxyImage:
+                    type: string
                   ceilometerSgcoreImage:
                     type: string
                   cinderAPIImage:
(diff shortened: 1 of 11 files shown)


dprince:
@jlarriba @beagles how does this look?

config/manifests/bases/openstack-operator.clusterserviceversion.yaml
```diff
@@ -21,7 +21,7 @@ spec:
   customresourcedefinitions:
     owned:
     - description: OpenStackClient is the Schema for the openstackclients API
-      displayName: OpenStack Client
+      displayName: Open Stack Client
```
abays, line 24: Does this PR need to be rebased to pick up #1006 ?

dprince: sure, just rebased

dprince:
This PR was a blocker for https://github.com/openstack-k8s-operators/openstack-operator/pull/1012 . I've now rebased 1012 to include this. If CI passes I may just use the larger PR as folks don't seem to get the motivation for these changes.

jlarriba:
It was holiday, not motivation, which made this PR unreviewed. Im sorry about that, but I see that the changes were already merged in the other PR.

Thank you @dprince

PR #1012, opened 2024-08-19
Drop csv merger and deduplicate RELATED_IMAGES...

dprince:
Goal of this PR is to simplify how we build the bundle for openstack-operator by
dropping CSV merged. It has proven difficult to mirror the csv-merger workflow
downstream and thus makes bundle builds there difficult.

Additionally, RELATED_IMAGES with the same value's are deduplicated to fix bundle
validation warnings


config/manifests/bases/openstack-operator.clusterserviceversion.yaml
```diff
@@ -316,7 +317,7 @@ spec:
         displayName: Template
         path: octavia.template
       - description: OpenStackClient - Parameters related to the OpenStackClient
-        displayName: OpenStack Client
+        displayName: Open Stack Client
```
fao89, line 320:
```suggestion
        displayName: OpenStack Client
```

dprince: Just fixed in my latest PR. NOTE: while this showed up on the GA release it was because that release was missing the operators.operatorframework.io/internal-objects which should have also been set to hide this altogether. It was related to the downstream bundle generation as we already had that set upstream.

fao89 (approved):

PR #1018, opened 2024-08-21
Run global services from serviceOverrides once

slagle:
Previously, if a global service (deployOnAllNodeSets=true) was in
serviceOverrides, an ansibleEE would be started for each NodeSet in the
Deployment, instead of just a single ansibleEE.

This commit changes the openstacknodest label on the OpenStackAnsibleEE
to openstacknodesets, which is a joined string of all the nodesets in
the execution. This way, subsequent NodeSets will fild the same
OpenStackAnsibleEE already in progress and not change the existing one.

When CreateOrPatch'ing the OpenStackAnsibleEE for a global service, also
do not update the ServiceAccount/ServiceAccountName. Doing so causes the
OpenStackAnsibleEE to continue to change hash values, which causes job
re-executions.

Jira: [OSPRH-9415](https://issues.redhat.com//browse/OSPRH-9415)

### DIFF
--- a/pkg/dataplane/deployment.go
+++ b/pkg/dataplane/deployment.go
@@ -82,6 +82,7 @@ func (d *Deployer) Deploy(services []string) (*ctrl.Result, error) {
 		nsConditions := d.Status.NodeSetConditions[d.NodeSet.Name]
 		log.Info("Deploying service", "service", service)
 		foundService, err := GetService(d.Ctx, d.Helper, service)
+
 		if err != nil {
 			nsConditions.Set(condition.FalseCondition(
 				readyCondition,
@@ -198,7 +199,7 @@ func (d *Deployer) ConditionalDeploy(
 
 	if nsConditions.IsFalse(readyCondition) {
 		var ansibleEE *ansibleeev1.OpenStackAnsibleEE
-		_, labelSelector := dataplaneutil.GetAnsibleExecutionNameAndLabels(&foundService, d.Deployment.Name, d.NodeSet.Name)
+		_, labelSelector := dataplaneutil.GetAnsibleExecutionNameAndLabels(&foundService, d.Deployment.Name, d.NodeSet.Name, d.Deployment.Spec.NodeSets)
 		ansibleEE, err = dataplaneutil.GetAnsibleExecution(d.Ctx, d.Helper, d.Deployment, labelSelector)
 		if err != nil {
 			// Return nil if we don't have AnsibleEE available yet
--- a/pkg/dataplane/util/ansible_execution.go
+++ b/pkg/dataplane/util/ansible_execution.go
@@ -60,7 +60,7 @@ func AnsibleExecution(
 
 	ansibleEEMounts := storage.VolMounts{}
 
-	executionName, labels := GetAnsibleExecutionNameAndLabels(service, deployment.GetName(), nodeSet.GetName())
+	executionName, labels := GetAnsibleExecutionNameAndLabels(service, deployment.GetName(), nodeSet.GetName(), deployment.Spec.NodeSets)
 	ansibleEE, err := GetAnsibleExecution(ctx, helper, deployment, labels)
 	if err != nil && !k8serrors.IsNotFound(err) {
 		return err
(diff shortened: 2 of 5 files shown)


dprince (approved):

pkg/dataplane/deployment.go
```diff
@@ -82,6 +83,17 @@ func (d *Deployer) Deploy(services []string) (*ctrl.Result, error) {
...
 		log.Info("Deploying service", "service", service)
 		foundService, err := GetService(d.Ctx, d.Helper, service)
+
+		// For global services, check if it's already been deployed, and if so,
+		// skip deploying it.
+		if foundService.Spec.DeployOnAllNodeSets {
+			if !slices.Contains(d.GlobalServicesDeployed, service) {
+				d.GlobalServicesDeployed = append(d.GlobalServicesDeployed, service)
+			} else {
+				return &ctrl.Result{}, err
```
rabi, line 93: I  think this should go after we handle the error  from L85 as foundService could be nil in case of error.  Also I guess we should skip the service rather than returning from here in case the `servicesOverride` has more services after the global service. Good to test that scenario too in functional tests.

rabi: btw, I had a slightly different approach using the existing de-duplication logic https://github.com/openstack-k8s-operators/openstack-operator/compare/main...rabi:openstack-operator:override?expand=1, if thats of interest.

slagle: maybe I'm not reading it right, but it looks like DedupeServices would result in NodeSetServiceMap values being empty for both NodeSets when servicesOverride contains a single global service.

then that empty value gets passed to Deploy in deployment_controller.

slagle:
recheck

```
2024-08-22 15:53:00.204931 | controller | Thursday 22 August 2024  15:52:56 -0400 (0:00:00.156)       0:57:07.529 *******
2024-08-22 15:53:00.204953 | controller | fatal: [localhost]: FAILED! => changed=true
2024-08-22 15:53:00.204963 | controller |   cmd: |-
2024-08-22 15:53:00.204973 | controller |     set -euo pipefail
2024-08-22 15:53:00.204982 | controller |     oc exec -n openstack openstackclient -- openstack network create \
2024-08-22 15:53:00.204991 | controller |         --external \
2024-08-22 15:53:00.205000 | controller |         --default \
2024-08-22 15:53:00.205010 | controller |         --provider-network-type flat \
2024-08-22 15:53:00.205019 | controller |         --provider-physical-network datacentre \
2024-08-22 15:53:00.205028 | controller |         --no-share \
2024-08-22 15:53:00.205038 | controller |         public
2024-08-22 15:53:00.205047 | controller |   delta: '0:00:02.874907'
2024-08-22 15:53:00.205056 | controller |   end: '2024-08-22 15:53:00.171609'
2024-08-22 15:53:00.205066 | controller |   msg: non-zero return code
2024-08-22 15:53:00.205075 | controller |   rc: 1
2024-08-22 15:53:00.205084 | controller |   start: '2024-08-22 15:52:57.296702'
2024-08-22 15:53:00.205094 | controller |   stderr: |-
2024-08-22 15:53:00.205112 | controller |     Error while executing command: HttpException: 503, p.info:before {: content: "";: color: #fff;: The host doesn't exist.: padding-left: 51px;: p {: ul {: padding: 30px;: position: relative;: background: #0066CC;: margin: 0 0 10px;: position: absolute;: li {: Application is not available: font-weight: 400;: Possible reasons you are seeing this page:: @media (min-width: 768px) {: line-height: 1.66666667;: The host exists, but doesn't have a matching path.: p.info {: p.info:before, p.info:after {: font-size: 16px;: font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;: .alert.alert-info {: width: 24px;: h1 {: Route and path matches, but all pods are down.: font-size: 14px;: margin-top: 30px;: top: 0;: content: "i";: border-radius: 16px;: Make sure the hostname was typed correctly and that a route matching this hostname exists.: left: 0px;: background-color: #fff;: The application is currently not serving requests at this endpoint. It may not have been started or is still starting.: margin: 6em;: margin-bottom: 1em;: }: padding-left: 35px;: top: 4px;: body {: height: 24px;: color: #333;: Make sure that the resources exposed by this route (pods, services, deployment configs, etc) have at least one pod running.: margin: 2em 1em;: font-size: 20px;: .alert p {: background-color: #F0F0F0;: font: bold 16px/24px serif;: text-align: center;: font-size: 28px;: left: 0;: Check if the URL path was typed correctly and that the route was created using the desired path.
```

pkg/dataplane/deployment.go
rabi: Not really, it would pass empty for the second nodeset, unless you're not looking at code I linked in the diff. There is functional test for the scenario you mention in that and I proposed as a PR to get CI results.. https://github.com/openstack-k8s-operators/openstack-operator/pull/1029

slagle: ok, i can see how that fixes the issue. Do you want to go with that instead?

slagle:
/hold

this can likely be closed in favor of https://github.com/openstack-k8s-operators/openstack-operator/pull/1029

PR #1022, opened 2024-08-22
[DNM] Moving download-cache after install-os

jpodivin:


apis/dataplane/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -61,7 +61,7 @@ type OpenStackDataPlaneNodeSetSpec struct {
 	NetworkAttachments []string `json:"networkAttachments,omitempty"`
 
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default={download-cache,bootstrap,configure-network,validate-network,install-os,configure-os,ssh-known-hosts,run-os,reboot-os,install-certs,ovn,neutron-metadata,libvirt,nova,telemetry}
+	// +kubebuilder:default={bootstrap,configure-network,validate-network,install-os,download-cache,configure-os,ssh-known-hosts,run-os,reboot-os,install-certs,ovn,neutron-metadata,libvirt,nova,telemetry}
```
slagle, line 64: I don't think this will be needed in we solve the other problems.

jpodivin: Probably not.

PR #1027, opened 2024-08-22
Use empty Alm examples

dprince:
Previously with csv-merger we removed the Alm examples in the CSV. This is to save space as our bundles are quite large.

Now that we are using the raw data generated by operator-sdk's 'generate bundle' command the config/sample data is getting merged into the generated CSV for openstack-operator. As we had multiple controlplanes in the kustomization.yaml file this meant lots of extra YAML examples which aren't needed.

In order to maintain the previous behavior of effectively removing ALM examples, but still passing bundle generate validations this commit creates '_empty.yaml' versions of each CR and then uses those in the kustomization.yaml.


config/samples/kustomization.yaml
```diff
@@ -1,11 +1,12 @@
 ## Append samples you want in your CSV to this file as resources ##
+# NOTE(dprince): these get added to the bundle ALM examples
+# we do not have space for long examples here and want
+# to guide users to documentation instead
```
fao89, line 4: OpenShift UI takes the first CR to prepopulate the UI, do we really want to get everything empty?

dprince: We have always been doing this as csv-merger removed all the alm-examples. This is effectively just restoring that (old) behavior

fao89 (approved):

PR #1029, opened 2024-08-23
Handle global services with servicesOverride in DedupeServices

rabi:


bshephar (approved):
I prefer this approach, bit cleaner with a good set of functional tests to ensure it works as expected. I think we should merge this one.

bshephar:
The commit message would need to be updated with details of how it's solving the problem and also to mention the Jira that this is fixing:
Jira: OSPRH-9415

But other than that, this one looks nice and clean.

PR #1034, opened 2024-08-27
Export environment variables for Kepler deployment

yadneshk:
Required for openstack-k8s-operators/edpm-ansible#723
Jira: https://issues.redhat.com/browse/OSP-31399


jpodivin (commented):
Is there a jira for this?

jlarriba:
@jpodivin https://issues.redhat.com/browse/OSP-31399

apis/dataplane/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -229,6 +229,7 @@ var ContainerImageDefaults = openstackv1.ContainerImages{
 		CeilometerComputeImage:        getStrPtr("quay.io/podified-antelope-centos9/openstack-telemetry-ceilometer-compute:current-podified"),
 		CeilometerIpmiImage:           getStrPtr("quay.io/podified-antelope-centos9/openstack-telemetry-ceilometer-ipmi:current-podified"),
 		EdpmNodeExporterImage:         getStrPtr("quay.io/podified-antelope-centos9/openstack-telemetry-node-exporter:current-podified"),
+		EdpmKeplerImage:               getStrPtr("quay.io/podified-antelope-centos9/openstack-telemetry-kepler:current-podified"),
```
vyzigold, line 232: The image doesn't exist (or at least I can't find it). Same goes for the rest of telemetry images above (but that's not part of this PR). Shouldn't we have this fallback default point to an existing destination?

vyzigold: Maybe these should have the same values as in config/default/manager_default_images.yaml? This seems to be the case for other services

jpodivin (commented):
LGTM

Although if there is a dependency between this and https://github.com/openstack-k8s-operators/edpm-ansible/pull/723 it should be made explicit.

vyzigold (changes requested):

apis/dataplane/v1beta1/openstackdataplanenodeset_types.go
yadneshk: > Maybe these should have the same values as in config/default/manager_default_images.yaml? This seems to be the case for other services

Yeah, I meant to change that for Kepler and the node-exporter image as well. Thanks!

slagle (approved):

apis/dataplane/v1beta1/openstackdataplanenodeset_types.go
csibbitt: The first question I had upon review was why the node-exporter image was changing. I've read this discussion and agree that defaults shouldn't point to an image that doesn't exist, but I'm still left wondering why we look like the odd one out among the other containers in the list.

Why does
quay.io/podified-antelope-centos9/openstack-nova-compute:current-podified   exist, but 
quay.io/podified-antelope-centos9/openstack-telemetry-node-exporter:current-podified  does not?

docs/assemblies/common_configurations.adoc
```diff
@@ -69,6 +69,7 @@ ifeval::["{build}" != "downstream"]
         edpm_ovn_bgp_agent_local_ovn_northd_image: "quay.io/podified-antelope-centos9/openstack-ovn-northd
         edpm_ovn_bgp_agent_local_ovn_controller_image: "quay.io/podified-antelope-centos9/openstack-ovn-controller
         edpm_telemetry_node_exporter_image: quay.io/prometheus/node-exporter
+        edpm_telemetry_kepler_image: "quay.io/sustainable_computing_io/kepler"
```
csibbitt, line 72: nit: I know this is docs, and maybe most of the problem isn't on this patch, but the mismatched quotes caught my eye. The node exporter line is unquoted, but the kepler line has "double-quotes", so I looked up/down to compare to the others for consistency and see it's a mix of unquoted (our images only!) and the rest of the lines have a leading quote with no terminating one. Either the unterminated double-quote is a feature of adoc (or yaml?) that I'm unaware of (in which case your line is wrong), or most of these lines are wrong, and ours are additionally the odd ones out again.

A more general comment is that if the docs are reproducing a large chunk of strings that come from the code, it might be worthwhile to auto-generate them so they remain correct and pastable for our users.

I guess this is not a request for a change on this patch so much as a sanity-check; are all these unterminated quotes intentional? If they are not, I suggest changing them all to be unquoted instead.

jpodivin: Part of the docs is already generated. The one involving kustomize that is. We could, in principle, write a script to extract certain chunks of our code and turn it into adoc. That being said, I'd much rather use an existing solution if there is one.

As for the inconsistency, yep, that's a recurring issue. I suggest we merge this change, and later submit a PR fixing the docs. Including it here would be messy.

apis/dataplane/v1beta1/openstackdataplanenodeset_types.go
csibbitt: Thank you for the explanation!

csibbitt (approved):

yadneshk:
recheck

PR #1038, opened 2024-08-30
Picking DNSDataLabelSelectorValue from DNSMasq struct

jpodivin:
It may have been cleaner to get the dnsmasq struct right before the `DNSData` is created, but it would involve duplicating all the checks in `checkDNSService`, since we are already using that to fill field in `DNSDetails` it feels like reasonable thing to use it for selector as well.

That being said, moving to alternative is trivial.


bshephar (approved):
Sounds reasonable.

PR #1042, opened 2024-09-03
Cleanup unused service certs when using apiOverride

Deydra71:
Deletes service route certs issued by default/custom issuer when using custom route certs via apiOverride or service override (when using load balancer for public endpoints)


Deydra71:
Still need to/create test for cert deletion when using load balancer for public endpoints

pkg/openstack/common.go
```diff
@@ -646,6 +669,27 @@ func (ed *EndpointDetail) CreateRoute(
...
+		// Delete the issued certificate if it exists and custom cert secret or direct TLS data was provided
+		if ed.Route.TLS.SecretName != nil || (ed.Route.OverrideSpec.Spec != nil && ed.Route.OverrideSpec.Spec.TLS != nil &&
+			(ed.Route.OverrideSpec.Spec.TLS.CACertificate != "" &&
+				ed.Route.OverrideSpec.Spec.TLS.Certificate != "" &&
+				ed.Route.OverrideSpec.Spec.TLS.Key != "")) {
+			certName := fmt.Sprintf("%s-route", ed.Name)
+			cert := certmanager.NewCertificate(
+				&certmgrv1.Certificate{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      certName,
```
stuggi, line 681: if we do the above mentioned change it would be `ed.Route.TLS.CertName`

pkg/openstack/common.go
```diff
@@ -646,6 +669,27 @@ func (ed *EndpointDetail) CreateRoute(
 			return ctrlResult, nil
 		}
 
+		// Delete the issued certificate if it exists and custom cert secret or direct TLS data was provided
+		if ed.Route.TLS.SecretName != nil || (ed.Route.OverrideSpec.Spec != nil && ed.Route.OverrideSpec.Spec.TLS != nil &&
+			(ed.Route.OverrideSpec.Spec.TLS.CACertificate != "" &&
+				ed.Route.OverrideSpec.Spec.TLS.Certificate != "" &&
+				ed.Route.OverrideSpec.Spec.TLS.Key != "")) {
```
stuggi, line 676: lets create a small func for the 2nd part to check the cert info was provided via the overrideSpec. makes it probably easier to understand

pkg/openstack/common.go
```diff
@@ -570,6 +587,12 @@ func (ed *EndpointDetail) CreateRoute(
 					}
 				}
 			}
+		} else if ed.Route.OverrideSpec.Spec != nil && ed.Route.OverrideSpec.Spec.TLS != nil {
+			if ed.Route.OverrideSpec.Spec.TLS.CACertificate == "" ||
+				ed.Route.OverrideSpec.Spec.TLS.Certificate == "" ||
+				ed.Route.OverrideSpec.Spec.TLS.Key == "" {
+				return ctrl.Result{}, fmt.Errorf("incomplete tls override data")
+			}
```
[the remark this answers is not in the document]
stuggi: also should then print a warning if someone adds both for one endpoint and mention that the cert secret content takes precedence. At least I think thats how it works right now.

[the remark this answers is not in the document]
stuggi: maybe we remove this check from this PR and you could submit a separate PR for doing those checks via the webhook?

pkg/openstack/common.go
```diff
@@ -290,6 +291,22 @@ func EnsureEndpointConfig(
...
+							ObjectMeta: metav1.ObjectMeta{
+								Name:      certName,
+								Namespace: ed.Namespace,
+							},
+						},
+						5*time.Second,
+					)
+					err = cert.Delete(ctx, helper)
+					if err != nil {
+						helper.GetLogger().Error(err, fmt.Sprintf("Failed to delete unused route certificate %s", certName))
```
stuggi, line 307: its almost the same as the error message returned by the cert.Delete(), do we need this additional message? https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/certmanager/certificate.go#L160

pkg/openstack/common.go
```diff
@@ -290,6 +291,22 @@ func EnsureEndpointConfig(
...
+					cert := certmanager.NewCertificate(
+						&certmgrv1.Certificate{
+							ObjectMeta: metav1.ObjectMeta{
+								Name:      certName,
+								Namespace: ed.Namespace,
+							},
+						},
+						5*time.Second,
+					)
+					err = cert.Delete(ctx, helper)
```
stuggi, line 305: how about the cert secret? can not remember what we said to do with it in this situation? delete it? add the owner to the secret?

Deydra71: iirc we agreed that we keep it as it is, and if user wants to delete secret with the certs he can set the `--enable-certificate-owner-ref` in cert-manager, and we only document it.

pkg/openstack/common.go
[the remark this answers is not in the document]
Deydra71: Yes, I tried it and when using secret with cert data, it has a privilege. We need to change the logic a little bit here if we remove the `ed.Route.OverrideSpec.Spec.TLS`  and check the provided CR with the webhooks, otherwise route cert will be looped in creation and deletion.

pkg/openstack/common.go
```diff
@@ -599,7 +619,6 @@ func (ed *EndpointDetail) CreateRoute(
 				return ctrlResult, nil
 			}
 		}
-
 		// create default TLS route override
```
stuggi, line 622: when `hasCertInOverrideSpec(ed.Route.OverrideSpec)` is true, I think we should also run the condition in L632 to get the `DestinationCaCertificate` if the user has not provided it and set the Termination to Reencrypt, or do we expect the end user to provide all this information as part of the `route.OverrideSpec` . It probably won't hurt to do it, since if something else is provided via the OverrideSpec by the user, it will have precedence.

pkg/openstack/common.go
```diff
@@ -647,6 +666,23 @@ func (ed *EndpointDetail) CreateRoute(
...
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      ed.Route.TLS.CertName,
+						Namespace: ed.Namespace,
+					},
+				},
+				5*time.Second,
+			)
+			err := cert.Delete(ctx, helper)
+			if err != nil {
+				helper.GetLogger().Error(err, fmt.Sprintf("Failed to delete unused route certificate %s", ed.Route.TLS.CertName))
```
stuggi, line 682: its a duplicate error message as we already get the one from https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/certmanager/certificate.go#L160, do we need this?

pkg/openstack/common.go
```diff
@@ -291,11 +294,26 @@ func EnsureEndpointConfig(
...
+							ObjectMeta: metav1.ObjectMeta{
+								Name:      ed.Route.TLS.CertName,
+								Namespace: ed.Namespace,
+							},
+						},
+						5*time.Second,
+					)
+					err = cert.Delete(ctx, helper)
+					if err != nil {
+						helper.GetLogger().Error(err, fmt.Sprintf("Failed to delete unused route certificate %s", ed.Route.TLS.CertName))
```
stuggi, line 309: its a duplicate error message as we already get the one from https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/certmanager/certificate.go#L160, do we need this?

pkg/openstack/common.go
```diff
@@ -746,3 +782,15 @@ func GetIssuerCertSecret(
 	}
 	return issuer.Spec.CA.SecretName, nil
 }
+
+func hasCertInOverrideSpec(overrideSpec route.OverrideSpec) bool {
```
stuggi, line 786: should we have a kuttl test which provides certs via the overridespec?

tests/kuttl/tests/ctlplane-tls-custom-route/05-check-cert-deletion.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: TestStep
+commands:
+  - script: |
+      echo "Checking absence of barbican-public-route certificate..."
+      if oc get certificate barbican-public-route -n $NAMESPACE; then
+        echo "barbican-public-route certificate is found"
+        exit 1
+      else
+        echo "barbican-public-route certificate is absent"
+      fi
```
stuggi, line 11: we can do that via the errors, like we check in the cleanup that resources are gone. https://github.com/openstack-k8s-operators/openstack-operator/blob/main/tests/kuttl/tests/ctlplane-tls-custom-issuers/11-errors-cleanup.yaml

pkg/openstack/common.go
Deydra71: I did that in first commit, but the linter was failing for detecting raw key, so I just deleted it. If would just override the linter here it would fail in upcoming PRs, right?

pkg/openstack/common.go
Deydra71: I'm not sure I understood 100%. But basically we can just add the `hasCertInOverrideSpec(ed.Route.OverrideSpec)` to:

`// for internal TLS (TLSE) use routev1.TLSTerminationReencrypt
		if ed.Service.TLS.Enabled && (ed.Service.TLS.SecretName != nil || hasCertInOverrideSpec(ed.Route.OverrideSpec)) {
......`

To set the termination to reencrypt, if the user doesn't directly specify it.

stuggi: yes thats what I meant. I think that's needed so in case the user did not set the reencrypt and added the dst ca cert to the override, it would be added by the operator.

pkg/openstack/common.go
stuggi: oh, you still have it somewhere so that we can check what the exact thing is it complains?

Deydra71: I added new commit with updated kuttl test with tls data in overridespec, so let's see

tests/kuttl/tests/ctlplane-tls-custom-route/05-errors-cleanup.yaml
stuggi, line 1: should be a link to the https://github.com/openstack-k8s-operators/openstack-operator/blob/main/tests/kuttl/common/errors_cleanup_openstack.yaml . adding the certs to be expected to be gone there.

tests/kuttl/tests/ctlplane-tls-custom-route/05-cleanup.yaml
stuggi, line 1: In a follow up I think we can make this a common cleanup file with a sum of all the required commands by any of the ctlplane test cases.

stuggi (approved):
/lgtm

PR #1045, opened 2024-09-03
Fixing style issues

jpodivin:
- Two error messages didn't comply with recommended style. It's largely irrelevant but it was polluting my IDE.
- One comment had a typo
- Annotations were inconsistently formatted (I'm not completely sold on this one, it's technically fine either way and it may pollute blame, I'll drop it if there is a push back)


fao89 (approved):

jpodivin:
/restest

` error: build error: Failed to push image: trying to reuse blob`

fao89 (approved):

PR #1047, opened 2024-09-04
Fail baremetal deployment if IPAM isn't configured

jpodivin:


pkg/dataplane/baremetal.go
```diff
@@ -70,29 +70,26 @@ func DeployBaremetalSet(
...
-						}
-						ipPrefix, _ := ipNet.Mask.Size()
-						instanceSpec.CtlPlaneIP = fmt.Sprintf("%s/%d", res.Address, ipPrefix)
-						if res.Gateway == nil {
-							return fmt.Errorf("%s gateway is missing", dataplanev1.CtlPlaneNetwork)
-						}
-						baremetalSet.Spec.CtlplaneGateway = *res.Gateway
-						baremetalSet.Spec.BootstrapDNS = dnsAddresses
-						baremetalSet.Spec.DNSSearchDomains = []string{res.DNSDomain}
+				err := fmt.Errorf("IPAM is not configured for %s", hostName)
```
rabi, line 73: This is kind of dead code as we fail if there are no networks in the spec (i.e IPAM not configured) or it fails to create IPSets [1]. If still there can be scenario that IPSets are not there for a node for some reason after being created successfully, we can probably change the errror message to something like  `IPSet not found for host %s'.  

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/dataplane/ipam.go#L363-L371

jpodivin: Ok, that's a solid point.

hjensas (commented):
/lgtm

PR #1057, opened 2024-09-10
Fix DataPlaneNodeSet tls verification

stuggi:
The dataplane connects vi the metallb loadbalancer k8s service to the pods exposed to them. If TLS is configured for those is controlled via the tls.podLevel configuraturation of the ctlplane.

Right now the implementation checks for both tls.ingress and tls.podLevel configuration. With this it is not possible to deploy edpm nodes with tls.podLevel disabled. This change updates the verification to just check that spec.tlsEnabled of the DataPlaneNodeSet matches tls.podLevel of the ctlplane.

### DIFF
--- a/tests/functional/dataplane/openstackdataplanedeployment_controller_test.go
+++ b/tests/functional/dataplane/openstackdataplanedeployment_controller_test.go
@@ -1563,7 +1563,7 @@ var _ = Describe("Dataplane Deployment Test", func() {
 		})
 	})
 
-	When("A user sets TLSEnabled to true with control plane TLS disabled", func() {
+	When("A user sets TLSEnabled to true with control plane with PodLevel TLS disabled", func() {
 		BeforeEach(func() {
 			CreateSSHSecret(dataplaneSSHSecretName)
 			DeferCleanup(th.DeleteInstance, th.CreateSecret(neutronOvnMetadataSecretName, map[string][]byte{
@@ -1596,7 +1596,7 @@ var _ = Describe("Dataplane Deployment Test", func() {
 			SimulateIPSetComplete(dataplaneNodeName)
 			SimulateDNSDataComplete(dataplaneNodeSetName)
 
-			DeferCleanup(th.DeleteInstance, CreateOpenStackControlPlane(controlPlaneName, GetDefaultOpenStackControlPlaneSpec(false)))
+			DeferCleanup(th.DeleteInstance, CreateOpenStackControlPlane(controlPlaneName, GetDefaultOpenStackControlPlaneSpec(true, false)))
 		})
 
 		It("Should have Spec fields initialized", func() {
@@ -1661,7 +1661,7 @@ var _ = Describe("Dataplane Deployment Test", func() {
 
 	})
 
-	When("A user sets TLSEnabled to true with control plane TLS enabled", func() {
+	When("A user sets TLSEnabled to true with control plane PodLevel TLS enabled", func() {
 		BeforeEach(func() {
 			CreateSSHSecret(dataplaneSSHSecretName)
 			DeferCleanup(th.DeleteInstance, th.CreateSecret(neutronOvnMetadataSecretName, map[string][]byte{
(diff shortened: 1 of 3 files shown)


tests/functional/dataplane/openstackdataplanedeployment_controller_test.go
```diff
@@ -1563,7 +1563,7 @@ var _ = Describe("Dataplane Deployment Test", func() {
 		})
 	})
 
-	When("A user sets TLSEnabled to true with control plane TLS disabled", func() {
+	When("A user sets TLSEnabled to true with control plane with PodLeve TLS disabled", func() {
```
abays, line 1566:
```suggestion
	When("A user sets TLSEnabled to true with control plane with PodLevel TLS disabled", func() {
```

jpodivin:
> btw shouldn't this be called in a webhook so that the openstackdataplanenodeset can not be created/updated with a miss match to the ctlplane config? in general I think we could probably get rid of tlsEnabled in the nodeset and get it from the ctlplane, or default it to what the ctlplane is configured for. Or is there a use case to have it different?

There was considerable pushback against that. Instead it was decided that the loop is the right place for this.

jpodivin:
> The dataplane connects vi the metallb loadbalancer k8s service to the pods exposed to them. If TLS is configured for those is controlled via the tls.podLevel configuraturation of the ctlplane.

I'm having hard time parsing this. Do you mean that load balancer connection to pods is conditioned on TLS being configured for them on the controlplane? And if so why is that a reason to disable this check? 
Or is TLS being enabled on ingress a problem? And if so, why can't we just ensure that TLS on ingress is enabled, instead?

jpodivin (commented):
> The dataplane connects vi the metallb loadbalancer k8s service to the pods exposed to them. If TLS is configured for those is controlled via the tls.podLevel configuraturation of the ctlplane.

I'm having hard time parsing this. Do you mean that load balancer connection to pods is conditioned on TLS being configured for them on the controlplane? And if so why is that a reason to disable this check? 
Or is TLS being enabled on ingress a problem? And if so, why can't we just ensure that TLS on ingress is enabled, instead?

stuggi:
> > The dataplane connects vi the metallb loadbalancer k8s service to the pods exposed to them. If TLS is configured for those is controlled via the tls.podLevel configuraturation of the ctlplane.
> 
> I'm having hard time parsing this. Do you mean that load balancer connection to pods is conditioned on TLS being configured for them on the controlplane? And if so why is that a reason to disable this check? Or is TLS being enabled on ingress a problem? And if so, why can't we just ensure that TLS on ingress is enabled, instead?

The dataplane do not care about tls on the ingress. they do not connect to the ingress/route. They connect via the loadbalancer service to the pod and tls.PodLevel is what controls if tls is enabled there.

stuggi:
> > btw shouldn't this be called in a webhook so that the openstackdataplanenodeset can not be created/updated with a miss match to the ctlplane config? in general I think we could probably get rid of tlsEnabled in the nodeset and get it from the ctlplane, or default it to what the ctlplane is configured for. Or is there a use case to have it different?
> 
> There was considerable pushback against that. Instead it was decided that the loop is the right place for this.

ok, its a different discussion I think, so can be done afterwards/somewhere else

jpodivin:
I'm going to leave decision on this to @vakwetu, since the check was implemented based on his specification. If he says it's fine to ignore the ingress than that's what we'll do.

stuggi:
> I'm going to leave decision on this to @vakwetu, since the check was implemented based on his specification. If he says it's fine to ignore the ingress than that's what we'll do.

well its not if we can, right now the check introduced a regression where you can not deploy edpm nodes without tlse.

vakwetu:
@stuggi @jpodivin I agree with this patch.  The dataplane doesn't care about ingress and should match pod level.  I must have missed that when this code was originally merged.  Otherwise, you can't deploy edpm nodes when tls-e is not enabled but public tls is.

jpodivin (approved):

PR #1058, opened 2024-09-10
Mount cacerts on computes if tls-e not enabled

vakwetu:
We should mount cacerts even if tls-e is not enabled, possibly for third party certs.
Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/749

Jira: [OSPRH-9451](https://issues.redhat.com//browse/OSPRH-9451)


pkg/dataplane/deployment.go
```diff
@@ -118,11 +118,9 @@ func (d *Deployer) Deploy(services []string) (*ctrl.Result, error) {
 			return &ctrl.Result{}, err
 		}
 
-		// Add certMounts if TLS is enabled
-		if d.NodeSet.Spec.TLSEnabled {
-			if foundService.Spec.AddCertMounts {
-				d.AeeSpec, err = d.addCertMounts(services)
-			}
+		// Add certMounts
+		if foundService.Spec.AddCertMounts {
```
stuggi, line 122: should `AddCertMounts` in the openstackdataplaneservice then default to true [1]? like if we need a CA on the very first place on the edpm nodes to e.g. be able to connect to satellite to install rpms or its container registry it should also run on bootstrap and maybe the default should be to just get the CA into any deployment of any role?

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/main/apis/dataplane/v1beta1/openstackdataplaneservice_types.go#L94

slagle: I don't recall the history here. Why do we have AddCertMounts when we already have TLSEnabled? It seems like there was a case where TLSEnabled=True but certs should not be mounted.

slagle:
shouldn't https://github.com/openstack-k8s-operators/edpm-ansible/pull/749 depend on this one instead?

pkg/dataplane/deployment.go
vakwetu: We have AddCertMounts because we decided to have a service that was dedicated to copying all the certs over for all the services.  That service is called "install-certs".  Basically, "install-certs" runs and installs all the certs on the computes - and then the ansible modules for the other services use and mount those certs as needed.
That way we can centralize the code to copy over the certs in one ansible role.  Copying the certs over is non-trivial, because the certs are node-specific.

So AddCertMounts is the attribute that tells the framework that this is this special "install-certs" role.

vakwetu:
Maybe - on the depends-on question.  I need to finish testing all this together first - and then figure out how to fix the tests.

pkg/dataplane/deployment.go
stuggi: thanks for clarification. I was expecting that this role just always runs, checking for certs and if there are any copies them, otherwise just exits.

vakwetu:
recheck

slagle (approved):

stuggi:
I think kuttl tests need an update, e.g. https://github.com/openstack-k8s-operators/openstack-operator/blob/main/tests/kuttl/tests/dataplane-deploy-no-nodes-test/01-assert.yaml#L140 . most likely other steps also require an update.

slagle (approved):

PR #1059, opened 2024-09-10
Ansibleee refactor

bshephar:
Further refactoring was requested, and pretty necessary following:
https://github.com/openstack-k8s-operators/openstack-operator/pull/972

This PR is the follow up to that change and splits out logic into more specific functions, deduplicating code that was used to create Volumes and VolumeMounts.


bshephar:
recheck

fao89 (approved):

PR #1061, opened 2024-09-11
Drop aee-operator dependency

fao89:
should we archive openstack-ansibleee-operator?

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -53,10 +53,6 @@ dependencies:
       version: ">=0.0.0"
   - type: olm.package
     value:
-      packageName: openstack-ansibleee-operator
-      version: ">=0.0.0"
-  - type: olm.package
-    value:
       packageName: openstack-baremetal-operator
       version: ">=0.0.0"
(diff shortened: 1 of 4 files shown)


slagle (commented):
this LGTM, but we could also include the cleanups in dependencies.yaml and CI_TOOLS_REPO/renovate.sh, and custom-bundle.Dockerfile.pinned

fao89:
custom-bundle.Dockerfile was dropped: https://github.com/openstack-k8s-operators/openstack-operator/pull/1012
and here is the renovate change: https://github.com/openstack-k8s-operators/openstack-k8s-operators-ci/pull/110

slagle:
> custom-bundle.Dockerfile was dropped: #1012 and here is the renovate change: [openstack-k8s-operators/openstack-k8s-operators-ci#110](https://github.com/openstack-k8s-operators/openstack-k8s-operators-ci/pull/110)

ok, stale file in my checkout :+1: 

what about dependencies.yaml?

slagle (approved):

PR #1062, opened 2024-09-11
OpenStackVersion: Add releaseVersionScheme and csvVersionAppend support

dprince:
The releaseVersionScheme can be optionally used to enable 'csvEpochAppend' behavior This is used by some downstream builds to append the version to the CSV version suffix (epoch) which is obtained from the OPERATOR_CONDITION_NAME environment variable to the openstack release version. In some downstream build systems CSV version bumps can be easily automated where as the OPENSTACK_RELEASE_VERSION is more of a static constant.

Co-authored-by: Miguel Garcia(mgarciac@redhat.com)

Jira: [OSPRH-9568](https://issues.redhat.com//browse/OSPRH-9568)

### DIFF
--- a/apis/core/v1beta1/openstackversion_types.go
+++ b/apis/core/v1beta1/openstackversion_types.go
@@ -17,7 +17,10 @@ limitations under the License.
 package v1beta1
 
 import (
+	"regexp"
+
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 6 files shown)


apis/core/v1beta1/openstackversion_types.go
```diff
@@ -204,3 +207,42 @@ func init() {
 func (instance OpenStackVersion) IsReady() bool {
 	return instance.Status.Conditions.IsTrue(condition.ReadyCondition)
 }
+
+func getOpenStackReleaseVersion(openstackReleaseVersion string, releaseVersionScheme string, operatorConditionName string) string {
```
bshephar, line 211: Can we add a test for this function to bring some clarity to the intended behaviour?

I think I get it, but if we have a functional test to validate that regex along with a couple of valid and invalid examples, I think it would be helpful.

bshephar (commented):
Can we add a test for this function to bring some clarity to the intended behaviour?

I think I get it, but if we have a functional test to validate that regex along with a couple of valid and invalid examples, I think it would be helpful.

apis/core/v1beta1/openstackversion_types.go
dprince: good call, added a few test cases

dprince:
recheck

stuggi:
recheck

abays (approved):
/lgtm

PR #1065, opened 2024-09-12
Rename InstanceHA to InstanceHa and add image to configmap

lmiccini:
This commits renames InstanceHA to InstanceHa to adhere to the openstackversion naming (casing) scheme.
It also adds the Infra InstanceHa image to a dedicated configmap so that the infra-operator can watch it and react to changes.


stuggi (approved):
/lgtm

PR #1067, opened 2024-09-13
Add service and issuer certs to errors cleanup

Deydra71:
Service and issuer certificates are deleted together with the control plane, so they should be added to the errors cleanup.


stuggi (approved):
/lgtm

PR #1072, opened 2024-09-13
[OSPRH-10186] Fix OpenStackVersion defaults setup

abays:
The `OpenStackVersion` defaults struct declares `availableVersion` as a private field.  It needs to be public (capitalized) so that it can be accessed outside of the function in which the surrounding struct/object was created.

### DIFF
--- a/apis/core/v1beta1/openstackversion_webhook.go
+++ b/apis/core/v1beta1/openstackversion_webhook.go
@@ -37,7 +37,7 @@ var versionWebhookClient goClient.Client
 
 // OpenStackVersionDefaults -
 type OpenStackVersionDefaults struct {
-	availableVersion string
+	AvailableVersion string
 }
 
 var openstackVersionDefaults OpenStackVersionDefaults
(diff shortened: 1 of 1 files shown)


stuggi (approved):
/lgtm

dprince:
Going to hold this until I understand more why this is needed only recently. I don't think there is a problem with this, but I'm not sure the user should change the default as that should always get set to availableVersion via an environment variable

dprince:
Still not understanding, I uncommented the unit test and it breaks it with the old code

+++ b/apis/core/v1beta1/openstackversion_webhook.go
@@ -70,9 +70,9 @@ var _ webhook.Defaulter = &OpenStackVersion{}
 // Default implements webhook.Defaulter so a webhook will be registered for the type
 func (r *OpenStackVersion) Default() {
        openstackversionlog.Info("default", "name", r.Name)
-       if r.Spec.TargetVersion == "" {
-               r.Spec.TargetVersion = openstackVersionDefaults.availableVersion
-       }
+       //if r.Spec.TargetVersion == "" {
+       //r.Spec.TargetVersion = openstackVersionDefaults.availableVersion
+       //}

PR #1080, opened 2024-09-17
Ensure TTLSecondsAfterFinished is set on AEE jobs

bshephar:


pkg/dataplane/util/ansibleee.go
```diff
@@ -140,6 +142,9 @@ func (a *EEJob) JobForOpenStackAnsibleEE(h *helper.Helper) (*batchv1.Job, error)
 	// Populate hash
 	hashes := make(map[string]string)
 
+	if a.PreserveJobs {
+		job.Spec.TTLSecondsAfterFinished = &TTLSecondsAfterFinishedValue
```
rabi, line 146: Why set a value for TTLSecondsAfterFinished? It says "If the field is unset, this Job won't be cleaned up by the TTL controller after it finishes."

bshephar: Yeah, you're right. Maybe we just reverse this logic then and we only set it if `!PreserveJobs`.

pkg/dataplane/util/ansibleee.go
```diff
@@ -140,6 +142,9 @@ func (a *EEJob) JobForOpenStackAnsibleEE(h *helper.Helper) (*batchv1.Job, error)
 	// Populate hash
 	hashes := make(map[string]string)
 
+	if a.PreserveJobs {
```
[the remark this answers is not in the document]
bshephar: True, I added it.

fao89 (commented):
+2

pkg/dataplane/util/ansibleee.go
rabi: This is fine as a change, but how does this fix the issues we're seeing in CI? 

Also, please change/remove https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/dataplane/util/ansibleee.go#L49 as the default for bool is false.

bshephar:
recheck

fao89:
recheck

pkg/dataplane/util/ansibleee.go
slagle: agree, it looks like PreserveJobs just gets hardcoded to True. we might as well drop it and just leave TTLSecondsAfterFinished unset

bshephar: Right, that makes sense. So to make sure we're clear, did you guys want to take another look at the updated change then. I'm just removing the `PreserveJobs` option from the `AEEJob` structure and leaving everything else the way it was since the defaults should be fine.

slagle: in this version of the PR, a.PreserveJobs was hardcoded to True, so that's why the suggestion was to just drop it. But, to do so you also need to drop it from openstackdataplanedeployment_types.go and pkg/dataplane/util/ansible_execution.go.

pkg/dataplane/util/ansibleee.go
```diff
@@ -45,9 +45,6 @@ type EEJob struct {
 	Args []string `json:"args,omitempty"`
 	// NetworkAttachments is a list of NetworkAttachment resource names to expose the services to the given network
 	NetworkAttachments []string `json:"networkAttachments,omitempty"`
-	// PreserveJobs - do not delete jobs after they finished e.g. to check logs
-	// PreserveJobs default: true
-	PreserveJobs bool `json:"preserveJobs,omitempty"`
```
slagle, line 50: need to drop it from openstackdataplanedeployment_types.go and pkg/dataplane/util/ansible_execution.go also

bshephar: Right, I had to rebase on main. I wasn't seeing any of these in my branch

```
❯ rg -i preservejobs -l
docs/assemblies/ctlplane_resources.adoc
apis/bases/core.openstack.org_openstackcontrolplanes.yaml
config/crd/bases/core.openstack.org_openstackcontrolplanes.yaml
```

Rebased, removing and re-testing now. New revision of this incoming.

apis/bases/dataplane.openstack.org_openstackdataplanedeployments.yaml
```diff
@@ -65,12 +65,6 @@ spec:
                   type: string
                 minItems: 1
                 type: array
-              preserveJobs:
```
bshephar, line 68: I'm not sure if we should be changing the API at this stage of development though. We might want to consider actually making it functional and then dropping it in a future release.

slagle: we just dropped the entire OpenStackAnsibleEE CRD though. I don't think we need to hold to a strict contract with deprecation period yet. Especially given this wasn't used.

PR #1082, opened 2024-09-17
Use RestartPolicy: Never for the jobs

rabi:
With RestartPolicy: OnFailure new pods are not created[1], but the containers in esiting pod are restarted. We were using RestartPolicy: Never with openstack-ansibleee-operator. This was changed when we switched to drop it's usage.

[1] https://kubernetes.io/docs/concepts/workloads/controllers/job/#handling-pod-and-container-failures

### DIFF
--- a/pkg/dataplane/util/ansible_execution.go
+++ b/pkg/dataplane/util/ansible_execution.go
@@ -178,8 +178,6 @@ func (a *EEJob) BuildAeeJobSpec(
 	service *dataplanev1.OpenStackDataPlaneService,
 	nodeSet client.Object,
 ) {
-	const jobRestartPolicy string = "OnFailure"
-
 	if aeeSpec.DNSConfig != nil {
 		a.DNSConfig = aeeSpec.DNSConfig
 	}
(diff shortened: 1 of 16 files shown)


tests/kuttl/tests/dataplane-service-failure/00-assert.yaml
```diff
@@ -115,7 +115,7 @@ spec:
             path: inventory
           secretName: dataplanenodeset-edpm-compute-no-nodes
 status:
-  failed: 1
+  failed: 3
```
fao89, line 118: oh! so that's why the backoff limit was reached with just one failure

fao89 (approved):

pkg/dataplane/util/ansible_execution.go
```diff
@@ -178,7 +178,7 @@ func (a *EEJob) BuildAeeJobSpec(
 	service *dataplanev1.OpenStackDataPlaneService,
 	nodeSet client.Object,
 ) {
-	const jobRestartPolicy string = "OnFailure"
+	const jobRestartPolicy string = "Never"
```
stuggi, line 181: could use the corev1.RestartPolicyNever const instead of using a custom const?

rabi: sure, done! There is quite a bit of refactoring still to be done...

fao89 (approved):

fao89:
recheck

stuggi:
I think you'd have to rebase to get https://github.com/openstack-k8s-operators/openstack-operator/pull/1084 ?

rabi:
> I think you'd have to rebase to get #1084 ?

I think we pull the PR on top of latest main in jobs.

slagle (approved):

PR #1083, opened 2024-09-17
Rebuild ManilaShares list

fmount:
We're currently not able to trigger a `shareCleanup` reconciliation loop when the top-level `CR` is updated and a `ManilaShare` entry is removed. In the effort of adding the `shareCleanup` feature in [1], we need to fix the `openstack-operator` code to rebuild a fresh version of `ManilaShares` from the existing CR.
Similarly to Cinder [2], this also fixes a regression for users who wants multiple shares with different backends after the initial deployment.

Jira: https://issues.redhat.com/browse/OSPRH-6525

[1] https://github.com/openstack-k8s-operators/manila-operator/pull/79
[2] https://github.com/openstack-k8s-operators/openstack-operator/commit/ede5a0645f17b2b5e4a06ac8c660f9cdd7507424

### DIFF
--- a/pkg/openstack/manila.go
+++ b/pkg/openstack/manila.go
@@ -120,9 +120,8 @@ func ReconcileManila(ctx context.Context, instance *corev1beta1.OpenStackControl
 			return errors.New("default Manila Share images is unset")
 		}
 
-		if manila.Spec.ManilaShares == nil {
-			manila.Spec.ManilaShares = make(map[string]manilav1.ManilaShareTemplate)
-		}
+		// Discard old list of share services and rebuild it
+		manila.Spec.ManilaShares = make(map[string]manilav1.ManilaShareTemplate)
 
 		for name, share := range instance.Spec.Manila.Template.ManilaShares {
 			manilaCore := manilav1.ManilaShareTemplate{}


pkg/openstack/manila.go
```diff
@@ -120,9 +120,8 @@ func ReconcileManila(ctx context.Context, instance *corev1beta1.OpenStackControl
 			return errors.New("default Manila Share images is unset")
 		}
 
-		if manila.Spec.ManilaShares == nil {
-			manila.Spec.ManilaShares = make(map[string]manilav1.ManilaShareTemplate)
-		}
+		// Discard old list of share services and rebuild it
+		manila.Spec.ManilaShares = make(map[string]manilav1.ManilaShareTemplate)
```
abays, line 124: I don't think it matters here, but I wanted to ask if this could be problematic in terms of a generated hash and map ordering being nondeterministic.

fmount: Correct, we do not keep an hash of the deployed shares in the top-level controller, so the order doesn't really matter. The hash is computed only for input (the osp-secret and the generated config) for both the top-level controller and each sub-resource, but the deployed shares do not play a role in the hash.
I agree that in a scenario where we would process the shares and compute the hash with a given criteria, we should sort them (by key?), and make sure we have a consistent mechanism of iterating through them to avoid unexpected results.

fmount:
recheck

stuggi:
I think you'd have to rebase to get https://github.com/openstack-k8s-operators/openstack-operator/pull/1084 ?

fmount:
> I think you'd have to rebase to get #1084 ?

aah makes sense! thank you!

fmount:
recheck

fmount:
Tested locally and I can see the new reconciliation loop being triggered now when we delete a share.

jpodivin (approved):

PR #1084, opened 2024-09-18
Use tempest-all image for running tempest

raukadah:
Currently jobs are blocked on
https://issues.redhat.com/browse/OSPCIX-488.
Fix is already proposed in https://github.com/rdo-packages/tempest-distgit/commit/10a5d132a3605cf20dade3b40737911253d1dfe1 But it is not included in current-promotion.

Let's use to tempest-all image to fix the issue.


abays:
```
Removing debug pod ...
error: unable to upgrade connection: container container-00 not found in pod oko-12-vp5dd-master-0-debug-llx9q_openstack-kuttl-tests
```

karelyatin:
i think we can override kuttl here, as the patch just touching zuul jobs and those are green now

PR #1089, opened 2024-09-20
Restoring lost defaults for ansible env vars

jpodivin:
`EnvConfigMapName` used to default to `openstack-aee-default-env`, which is a config map we use[0] in install-yamls.
After the merger of ansibleeeoperator with the openstack operator, we have lost this (and other) defaults, since the new `EEJob` struct is only an internal representation and not a resource for customers to modify. 

Going forward we have essentially two options.

1. Restore the default for the config map name. Allowing us to create and modify it, in order to change ansibleee behavior. 
2. Remove the ` EnvConfigMapName` field entirely, as it is effectively unused.

Should we go with option 1. the obvios next step would be to allow modifications to the field value. In order to substitute alternative configuration in a simpler fashion. We could put that field in the `OpenStackDataPlaneDeployment` for example.

[0] https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/edpm/config/ansible-ee-env.yaml

### DIFF
--- a/pkg/dataplane/util/ansible_execution.go
+++ b/pkg/dataplane/util/ansible_execution.go
@@ -63,9 +63,10 @@ func AnsibleExecution(
 	}
 
 	ansibleEE := EEJob{
-		Name:      executionName,
-		Namespace: deployment.GetNamespace(),
-		Labels:    labels,
+		Name:             executionName,
+		Namespace:        deployment.GetNamespace(),
+		Labels:           labels,
+		EnvConfigMapName: "openstack-aee-default-env",
 	}
 
 	ansibleEE.NetworkAttachments = aeeSpec.NetworkAttachments
--- a/pkg/dataplane/util/ansibleee.go
+++ b/pkg/dataplane/util/ansibleee.go
@@ -230,10 +230,14 @@ func labelsForOpenStackAnsibleEE(name string, labels map[string]string) map[stri
 }
 
 func (a *EEJob) addEnvFrom(job *batchv1.Job) {
+	// Add optional config map
+
+	optional := true
 	job.Spec.Template.Spec.Containers[0].EnvFrom = []corev1.EnvFromSource{
 		{
 			ConfigMapRef: &corev1.ConfigMapEnvSource{
 				LocalObjectReference: corev1.LocalObjectReference{Name: a.EnvConfigMapName},
+				Optional:             &(optional),
 			},
 		},
 	}


fao89 (approved):

pkg/dataplane/util/ansible_execution.go
```diff
@@ -63,9 +63,10 @@ func AnsibleExecution(
 	}
 
 	ansibleEE := EEJob{
-		Name:      executionName,
-		Namespace: deployment.GetNamespace(),
-		Labels:    labels,
+		Name:             executionName,
+		Namespace:        deployment.GetNamespace(),
+		Labels:           labels,
+		EnvConfigMapName: "openstack-aee-default-env",
```
slagle, line 69: This would make the ConfigMap required wouldn't it?
Should we add optional:True here, https://github.com/openstack-k8s-operators/openstack-operator/blob/98fbd73bd2606479f25151a4d9a56282c092f1bb/pkg/dataplane/util/ansibleee.go#L236

jpodivin: Yes that does make sense.

slagle (approved):

abays:
> @jpodivin: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> Test name 	Commit 	Details 	Required 	Rerun command
> ci/prow/openstack-operator-build-deploy-kuttl 	[9ca9b3d](https://github.com/openstack-k8s-operators/openstack-operator/commit/9ca9b3d42bcd35ceb78aeb6ead76d524d9a6c19e) 	[link](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openstack-k8s-operators_openstack-operator/1089/pull-ci-openstack-k8s-operators-openstack-operator-main-openstack-operator-build-deploy-kuttl/1837122749114355713) 	true 	`/test openstack-operator-build-deploy-kuttl`
> 
> [Full PR test history](https://prow.ci.openshift.org/pr-history?org=openstack-k8s-operators&repo=openstack-operator&pr=1089). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:jpodivin).

```
error: unable to upgrade connection: container container-00 not found in pod oko-07-pv78l-master-2-debug-7t985_openstack-kuttl-tests
```

pkg/dataplane/util/ansibleee.go
```diff
@@ -230,10 +230,14 @@ func labelsForOpenStackAnsibleEE(name string, labels map[string]string) map[stri
 }
 
 func (a *EEJob) addEnvFrom(job *batchv1.Job) {
```
rabi, line 232: This method is not even used atm. I had a PR commit [1] already in the queue that was making  the change in a different context (i.e removing the interface). Anyway, I'll rebase that. 

[1] https://github.com/openstack-k8s-operators/openstack-operator/commit/64e9614a84192420cb8b18fa59ebf6a7130fd275

PR #1094, opened 2024-09-20
[18.0.0-proposed] Mount cacerts on computes if tls-e not enabled

vakwetu:
We should mount cacerts even if tls-e is not enabled, possibly for third party certs.

Jira: [OSPRH-9451](https://issues.redhat.com//browse/OSPRH-9451)
(cherry picked from 7b8474294618527fd35abedd82c5d26771cd7d12)


slagle (approved):
lgtm, https://github.com/openstack-k8s-operators/edpm-ansible/pull/754 is the related PR for edpm-ansible

vakwetu:
recheck

stuggi:
cold migration on openstack-operator-tempest-multinode job is failing because of host key verification error due to ssh_known_hosts file is not generated. this is being tracked in https://issues.redhat.com/browse/OSPRH-10305 . The PR which introduced this is only merged in main and not yet backported to 18.0.0-proposed. The reason why this PR hit it is that 18.0.0-proposed jobs currently use the openstack-ansibleee-runner latest tagged image and not the 18.0.0-proposed-latest. this is tracked in https://issues.redhat.com/browse/OSPRH-10306.

I'll override the job

PR #1098, opened 2024-09-22
Use OCP registries.conf for disconnected deployments

bshephar:
This change adds functionality to check for ImageContentSourcePolicy in the OCP cluster. If we find it, we infer from its existance that we are running in a disconnected OCP deployment. If this is indeed the case, we will pull the registries.conf file from the MachineConfig and use this for our dataplane nodes configured via edpm_ansible.

Jira: https://issues.redhat.com/browse/OSPRH-10303

### DIFF
--- a/controllers/dataplane/openstackdataplanenodeset_controller.go
+++ b/controllers/dataplane/openstackdataplanenodeset_controller.go
@@ -120,6 +120,10 @@ func (r *OpenStackDataPlaneNodeSetReconciler) GetLogger(ctx context.Context) log
 // +kubebuilder:rbac:groups="image.openshift.io",resources=imagetags,verbs=get;list;watch
 // +kubebuilder:rbac:groups="image.openshift.io",resources=imagestreamtags,verbs=get;list;watch
 
+// RBAC for ImageContentSourcePolicy and MachineConfig
+// +kubebuilder:rbac:groups="operator.openshift.io",resources=imagecontentsourcepolicy,verbs=get;list
+// +kubebuilder:rbac:groups="machineconfiguration.openshift.io",resources=machineconfig,verbs=get;list
+
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
 // TODO(user): Modify the Reconcile function to compare the state specified by
@@ -565,7 +569,8 @@ func checkDeployment(ctx context.Context, helper *helper.Helper,
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *OpenStackDataPlaneNodeSetReconciler) SetupWithManager(
-	ctx context.Context, mgr ctrl.Manager) error {
+	ctx context.Context, mgr ctrl.Manager,
+) error {
 	// index for ConfigMaps listed on ansibleVarsFrom
 	if err := mgr.GetFieldIndexer().IndexField(ctx,
 		&dataplanev1.OpenStackDataPlaneNodeSet{}, "spec.ansibleVarsFrom.ansible.configMaps",
--- a/controllers/dataplane/openstackdataplanenodeset_controller.go
+++ b/controllers/dataplane/openstackdataplanenodeset_controller.go
@@ -54,6 +54,8 @@ import (
 	dataplanev1 "github.com/openstack-k8s-operators/openstack-operator/apis/dataplane/v1beta1"
 	deployment "github.com/openstack-k8s-operators/openstack-operator/pkg/dataplane"
 	dataplaneutil "github.com/openstack-k8s-operators/openstack-operator/pkg/dataplane/util"
+
+	machineconfig "github.com/openshift/api/machineconfiguration/v1"
 )
 
 const (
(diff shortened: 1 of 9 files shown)


bshephar:
Go 1.21 and release-4.16 allowed me to use the `MachineConfig` CRD from openshift/api which we already have a dependency on:
https://github.com/openshift/api/blob/release-4.16/machineconfiguration/v1/types.go#L301

So, we will need to wait until the Go upgrade PR merges before we look at this one. It should reduce the size of the PR once that other one is merged.

bshephar:
Failure is:
```
❯ curl -s "https://logserver.rdoproject.org/98/1098/0ef0806ac21876cddc079be6c4a6d8f4d70dca6c/github-check/podified-multinode-edpm-deployment-crc/ecd5596/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/openstackdataplanenodesets.dataplane.openstack.org/openstack-edpm-ipam.yaml" | yq ".status.conditions[0]"
lastTransitionTime: "2024-09-26T04:21:07Z"
message: NodeSetDNSDataReady error occurred DNSData.network.openstack.org "openstack-edpm-ipam" not found
reason: Error
severity: Error
status: "False"
type: Ready
```

It does indeed appear to be there:
https://logserver.rdoproject.org/98/1098/0ef0806ac21876cddc079be6c4a6d8f4d70dca6c/github-check/podified-multinode-edpm-deployment-crc/ecd5596/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/dnsdata.network.openstack.org/

Need to dig deeper to see if this is somehow related. But initial look appears to not be related.

pkg/dataplane/util/image_registry.go
```diff
@@ -0,0 +1,127 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// machineConfigIgnition - holds the relevant parts of the ignition file that we need to create
+// the registries.conf on the dataplane nodes.
+type machineConfigIgnition struct {
+	Ignition struct {
```
gibizer, line 21: I guess something similar is defined in MachineConfig api module. Are we intentionally not trying to import the original struct?

pkg/dataplane/util/image_registry.go
```diff
@@ -0,0 +1,127 @@
...
+	return registriesConf, nil
+}
+
+func (mci *machineConfigIgnition) removePrefixFromB64String() string {
+	const b64Prefix string = "data:text/plain;charset=utf-8;base64,"
+	if strings.HasPrefix(mci.Storage.Files[0].Contents.Source, b64Prefix) {
+		return mci.Storage.Files[0].Contents.Source[len(b64Prefix):]
+	}
+
+	return ""
```
gibizer, line 80: hm, this feels wrong, if it does not have a prefix then it is either an error or we can use the unprefixed value but returning empty here feels wrong. Or do we rely on the caller to error out if it gets an empty string?

controllers/dataplane/openstackdataplanenodeset_controller.go
```diff
@@ -120,6 +120,10 @@ func (r *OpenStackDataPlaneNodeSetReconciler) GetLogger(ctx context.Context) log
 // +kubebuilder:rbac:groups="image.openshift.io",resources=imagetags,verbs=get;list;watch
 // +kubebuilder:rbac:groups="image.openshift.io",resources=imagestreamtags,verbs=get;list;watch
 
+// RBAC for ImageContentSourcePolicy and MachineConfig
+// +kubebuilder:rbac:groups="operator.openshift.io",resources=imagecontentsourcepolicy,verbs=get;list
+// +kubebuilder:rbac:groups="machineconfiguration.openshift.io",resources=machineconfig,verbs=get;list
+
```
gibizer, line 126: do we need to watch them for changes? I.e. can the ICSP or the MC change after the initial deployment?

bshephar: True, we probably do need a Watcher for the MachineConfig. Probably not so much for the ImageContentSourcePolicy though. We're only using that to determine whether we need to configure mirrors at all.

pkg/dataplane/util/image_registry.go
bshephar: I thought the same, but it's just defined like this:
https://github.com/openshift/api/blob/master/machineconfiguration/v1/types.go#L320-L322

So I thought just implementing a minimal struct with our requirements would be a cleaner option.

pkg/dataplane/util/image_registry.go
bshephar: Yeah, I don't really like this in general tbh. I thought there would be a better way to parse that text based on the prefix. I might need to dig deeper into how MCO uses this value and see if there's some ideas for a cleaner implementation here. I can't really see it being defined as anything else, unless the Spec for this base64 string designation changes, but I guess we should return an error here. Or at least some feedback to the user that we couldn't retrieve the OCP `registries.conf` and we will fall back to the original implementation in edpm-ansible in those situations.

bshephar:
recheck

pkg/dataplane/util/image_registry.go
bshephar: Changed this to return an error if it can't find the prefix string in the output.

controllers/dataplane/openstackdataplanenodeset_controller.go
bshephar: Last commit on the PR now aims to implement this watcher

controllers/dataplane/openstackdataplanenodeset_controller.go
```diff
@@ -639,9 +646,46 @@ func (r *OpenStackDataPlaneNodeSetReconciler) SetupWithManager(
...
+		Name:      registryMachineConfigName,
+		Namespace: "",
+	}
+
+	if err := r.Get(ctx, mcNamespacedName, obj); err != nil {
+		Log.Error(err, fmt.Sprintf("Unable to retrieve MachingConfig %s", registryMachineConfigName))
+		return nil
+	}
+
+	requests := make([]reconcile.Request, 0, len(nodeSets.Items))
```
fao89, line 675: I think we need to query the nodesets first
```suggestion
	listOpts := []client.ListOption{
		client.InNamespace(obj.GetNamespace()),
	}
	if err := r.Client.List(ctx, nodeSets, listOpts...); err != nil {
		Log.Error(err, "Unable to retrieve OpenStackDataPlaneNodeSetList")
		return nil
	}

	requests := make([]reconcile.Request, 0, len(nodeSets.Items))
```

bshephar:
recheck

fao89 (approved):

bshephar:
recheck

fao89:
recheck

fao89 (approved):

PR #1101, opened 2024-09-23
[18.0.0-proposed] Update AEE tag

fao89:
closes [OSPRH-10306](https://issues.redhat.com//browse/OSPRH-10306)

### DIFF
--- a/config/default/manager_default_images.yaml
+++ b/config/default/manager_default_images.yaml
@@ -14,7 +14,7 @@ spec:
         - name: RELATED_IMAGE_AGENT_IMAGE_URL_DEFAULT
           value: quay.io/openstack-k8s-operators/openstack-baremetal-operator-agent:latest
         - name: RELATED_IMAGE_ANSIBLEEE_IMAGE_URL_DEFAULT
-          value: quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest
+          value: quay.io/openstack-k8s-operators/openstack-ansibleee-runner:18.0.0-proposed-latest
         - name: RELATED_IMAGE_AODH_API_IMAGE_URL_DEFAULT
           value: quay.io/podified-antelope-centos9/openstack-aodh-api:current-podified
         - name: RELATED_IMAGE_AODH_EVALUATOR_IMAGE_URL_DEFAULT


apis/dataplane/v1beta1/openstackdataplanenodeset_types.go
```diff
@@ -207,7 +207,7 @@ func (instance OpenStackDataPlaneNodeSet) GetAnsibleEESpec() AnsibleEESpec {
 var ContainerImageDefaults = openstackv1.ContainerImages{
 	ContainerTemplate: openstackv1.ContainerTemplate{
 		AgentImage:                    getStrPtr("quay.io/openstack-k8s-operators/openstack-baremetal-operator-agent:current-podified"),
-		AnsibleeeImage:                getStrPtr("quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest"),
+		AnsibleeeImage:                getStrPtr("quay.io/openstack-k8s-operators/openstack-ansibleee-runner:18.0.0-proposed-latest"),
```
rabi, line 210: I don't think we can change it here and that too in the main branch.  It has to be updated in `OpenStackVersion` CR when deployed with 18.0.0-proposed branch I guess.

fao89: but it updates the `OpenStackVersion`: https://logserver.rdoproject.org/01/1101/34cca5e4af767ffdd79db54ab79a415fa466e7f7/github-check/podified-multinode-edpm-deployment-crc/f80d6de/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/openstackversions.core.openstack.org/controlplane.yaml

dprince:
The image should only get updated in the 18.0.1 bundle. A new OpenStackVersion CR will then pick it up and allow the user to "update" (choose) to use the new targetVersion with this image. There should be no need to patch an image like this for a maintenance branch

dprince:
This isn't a pattern we want to repeat for the service containers but there is a need and it is blocking us. So lets go ahead and make the exception.

dprince (approved):

PR #1102, opened 2024-09-23
Adding general map field for configuring AEE

jpodivin:
New config map holding data for additional AEE configuration. By it's nature it's open ended and all the fields will require checking. Normally we would could use verification in schema for this. But since we don't want to change signature of resources too much conditionals are a way to go.


pkg/dataplane/deployment.go
```diff
@@ -496,3 +504,25 @@ func (d *Deployer) addServiceExtraMounts(
...
+	var test bool
+
+	err := dataplaneutil.UnmarshalRawStrings(
+		d.Deployment.Spec.AnsibleEEConfig, additionalAEEOptions)
+	if err != nil {
+		return err
+	}
+
+	// Check presence of the key, type and contents
+	configMapName, test := additionalAEEOptions["envConfMapName"]
```
fao89, line 519:
```suggestion
	configMapName, test := additionalAEEOptions["envConfigMapName"]
```

pkg/dataplane/deployment.go
```diff
@@ -496,3 +504,24 @@ func (d *Deployer) addServiceExtraMounts(
...
+	var test bool
+	cm, _, err := dataplaneutil.GetDataSourceCmSecret(d.Ctx, d.Helper, d.NodeSet.Namespace, d.Deployment.Spec.AnsibleEEConfig)
+
+	if err != nil || cm == nil {
+		d.AeeSpec.EnvConfigMapName = "openstack-aee-default-env"
+		return fmt.Errorf("aditional AEE options not found")
+	}
+
+	// Check presence of the key, type and contents
+	configMapName, test := cm.Data["envConfigMapName"]
```
slagle, line 518: is it that the CM named d.Deployment.Spec.AnsibleEEConfig stores the name of another CM that actually stores the config? if so, why is that the case?

jpodivin: Idea is that the `AnsibleEEConfig` will be catch all place for all the configuration options that don't necessarily need their own field. Whether they will be defined as config maps, such as in this case, or plain values.

In this case we are storing name of a config map, which is used for environment variables. So yes, we could just remove that one level and get the same value. But it would prevent us from using this field for anything else. At some point in the future, if customer came along and asked, we would have to add another field, and another. This way we just put it in a general config map, and let this function do whatever is necessary with it.

slagle:
I like the overall idea here, but I would like to see this proposed as a feature in Jira, aligned against the next FR (FR2), so we can properly scope and plan with QE and docs.

PR #1103, opened 2024-09-26
Revert "[ovn-controller] Don't create ovn-controller if nicMappings empty"

averdagu:
This reverts commit a02a1596b972a599d77158d9dc746849f8075b23.

This change created a bug on the adoption where all the gates reaches TIME_OUT, reverting this patch to unblock prod chain while fixing it.
The problem that this patch was introducing was that edpm deployment expects a config map created by ovn-operator (ovncontroller) called ovncontroller-config. With this patch the ovncontroller CR was not created and the ovn-operator was not reconciling this object, which was needed to create the said config map.

Since, as said before, this is blocking product chain gates, the faster solution to unblock it is to revert the patch and in parallel work on the best way to implement the code.

Related: [OSPRH-7463](https://issues.redhat.com//browse/OSPRH-7463)
Related: [OSPCIX-504](https://issues.redhat.com//browse/OSPCIX-504)


karelyatin:
/lgtm

Can rerevert once adoption issue is sorted. Also would be good to have some adoption jobs to gate patches in openstack-operator to avoid similar issue with future merges.

booxter:
/approve

Let's hope CI is happy. :)

cescgina:
> /lgtm
> 
> Can rerevert once adoption issue is sorted. Also would be good to have some adoption jobs to gate patches in openstack-operator to avoid similar issue with future merges.

we do have adoption jobs in this repo https://github.com/openstack-k8s-operators/openstack-operator/blob/eee64b7ee793cebe2e5a8e0fe6009c581db2b837/zuul.d/projects.yaml#L6 it looks like we need to improve when they are triggered.
For reference this is the project template used here https://github.com/openstack-k8s-operators/ci-framework/blob/d14df7163b1ce5b8ba92941f6d0059497fa0db32/zuul.d/project-templates.yaml#L69. It would run this job https://github.com/openstack-k8s-operators/ci-framework/blob/d14df7163b1ce5b8ba92941f6d0059497fa0db32/zuul.d/adoption.yaml#L104 we should expand the `files` section there to run when relevant files are modified. We kept it simple to avoid running the long adoption jobs in many PRs where they were not needed. CC @marios @frenzyfriday @jistr

PR #1106, opened 2024-09-27
Bump capability level to Seamless Upgrades

dprince:
We have a minor updates workflow which I think
means we satisfy this capability level.


odyssey4me (approved):

jpodivin (approved):

fao89:
recheck

PR #1113, opened 2024-10-02
Add a webhook validation for tls apiOverride

Deydra71:
Add webhook to validate TLS apiOverride completeness (CA certificate, certificate, key) before CR is applied.

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_webhook.go
+++ b/apis/core/v1beta1/openstackcontrolplane_webhook.go
@@ -46,8 +46,8 @@ import (
 	heatv1 "github.com/openstack-k8s-operators/heat-operator/api/v1beta1"
 	horizonv1 "github.com/openstack-k8s-operators/horizon-operator/api/v1beta1"
 	memcachedv1 "github.com/openstack-k8s-operators/infra-operator/apis/memcached/v1beta1"
-	redisv1 "github.com/openstack-k8s-operators/infra-operator/apis/redis/v1beta1"
 	networkv1 "github.com/openstack-k8s-operators/infra-operator/apis/network/v1beta1"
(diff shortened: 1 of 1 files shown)


abays:
Looks good to me.  Maybe @olliewalsh should give the final go-ahead though since it relates to TLS.

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -902,3 +937,27 @@ func (r *OpenStackControlPlane) DefaultLabel() {
...
+func validateTLSOverrideSpec(override **route.OverrideSpec, basePath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+
+	if *override == nil {
+		return allErrs
+	}
+
+	tlsSpec := (*override).Spec
+	if tlsSpec != nil && tlsSpec.TLS != nil {
+		if tlsSpec.TLS.CACertificate == "" {
```
olliewalsh, line 951: CA cert is optional IIRC

olliewalsh (changes requested):

apis/core/v1beta1/openstackcontrolplane_webhook.go
Deydra71: You mean when use the hostname validation using service name? I didn't try this specific case, but when we don't include the CA cert, `destinationCACertificate` is missing error shows up, and route is not created.

Deydra71: Yes, it's optional, when the CA cert is not used, the bundle one is

Deydra71:
recheck

stuggi:
lgtm, @olliewalsh are your change requests addressed?

olliewalsh (approved):
/lgtm

PR #1123, opened 2024-10-09
Add openstack must-gather to required images

bshephar:
In disconnected environments, we should have a reference to the openstack-must-gather image. This allows for users to clone in the required image to their disconnected registries and enable them to gather logs for support cases and troubleshooting.

Jira: https://issues.redhat.com/browse/OSPRH-10635

### DIFF
--- a/config/default/manager_default_images.yaml
+++ b/config/default/manager_default_images.yaml
@@ -177,3 +177,5 @@ spec:
           value: quay.io/podified-antelope-centos9/openstack-ansible-tests:current-podified
         - name: RELATED_IMAGE_TEST_HORIZONTEST_IMAGE_URL_DEFAULT
           value: quay.io/podified-antelope-centos9/openstack-horizontest:current-podified
+        - name: RELEATED_IMAGE_OPENSTACK_MUST_GATHER_DEFAULT
+          value: quay.io/openstack-k8s-operators/openstack-must-gather:latest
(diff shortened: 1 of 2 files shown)


hack/export_related_images.sh
```diff
@@ -85,3 +85,4 @@ export RELATED_IMAGE_TEST_TEMPEST_IMAGE_URL_DEFAULT=quay.io/podified-antelope-ce
 export RELATED_IMAGE_TEST_TOBIKO_IMAGE_URL_DEFAULT=quay.io/podified-antelope-centos9/openstack-tobiko:current-podified
 export RELATED_IMAGE_TEST_ANSIBLETEST_IMAGE_URL_DEFAULT=quay.io/podified-antelope-centos9/openstack-ansible-tests:current-podified
 export RELATED_IMAGE_TEST_HORIZONTEST_IMAGE_URL_DEFAULT=quay.io/podified-antelope-centos9/openstack-horizontest:current-podified
+export RELATED_IMAGE_OPENSTACK_MUST_GATHER_DEFAULT=quay.io/openstack-k8s-operators/openstack-must-gather:latest
```
leifmadsen, line 88: Nothing else has a tag on the end. Probably should remove.

leifmadsen: Disregard, I misread!

config/default/manager_default_images.yaml
```diff
@@ -177,3 +177,5 @@ spec:
           value: quay.io/podified-antelope-centos9/openstack-ansible-tests:current-podified
         - name: RELATED_IMAGE_TEST_HORIZONTEST_IMAGE_URL_DEFAULT
           value: quay.io/podified-antelope-centos9/openstack-horizontest:current-podified
+        - name: RELEATED_IMAGE_OPENSTACK_MUST_GATHER_DEFAULT
```
fmount, line 180: As per the interface defined in the commit [1], it seems the right place to define the `must-gather` image. We can solve the problem of disconnected environments and provide documentation about this particular use case.
I'm ok with this change if it represents the best practice in the `operator-framework` context.


[1] https://github.com/operator-framework/api/commit/4f51a5d0bab048fa7d3751396ebef662cf8b69a5#diff-3490b8e064835cb2059f50d7a4adda830a7a048708976c257f04ca6de42b9742R317

hack/export_related_images.sh
fmount, line 88: @bshephar for each FR that we branch, do we need to maintain a different tag? e.g. I see there is a `18.0.0-proposed` build in quay, and I'm wondering if that should proactively be added to the `18.0.0-proposed` branch of the openstack operator.

fao89 (approved):

PR #1127, opened 2024-10-10
Make sure to log error in findObjectsForSrc()

stuggi:
It is hidden right now if there is an error in configured named fields and index. Lets log the error and return the current state of requests.

With this the findObjectsForSrc() will exit with an hidden error like:

```
"error": "Index with name field:.spec.ksmTls.caBundleSecretName does not exist"
```

### DIFF
--- a/controllers/client/openstackclient_controller.go
+++ b/controllers/client/openstackclient_controller.go
@@ -525,11 +525,6 @@ func (r *OpenStackClientReconciler) SetupWithManager(
 			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
 		).
 		Watches(
-			&corev1.Secret{},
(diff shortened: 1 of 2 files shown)


controllers/client/openstackclient_controller.go
```diff
@@ -542,6 +537,8 @@ func (r *OpenStackClientReconciler) SetupWithManager(
 func (r *OpenStackClientReconciler) findObjectsForSrc(ctx context.Context, src client.Object) []reconcile.Request {
 	requests := []reconcile.Request{}
 
+	l := log.FromContext(context.Background()).WithName("Controllers").WithName("PlacementAPI")
```
abays, line 540: Should `PlacementAPI` be something else here?

stuggi: sure

abays (approved):
/lgtm

PR #1128, opened 2024-10-11
minor update: unblock if dataplane never deployed

dprince:
Add a unit test and adjust logic to allow minor updates to complete when only the ControlPlane is completely deployed. Previously if the dataplaneDeployments were in a state of failure (upon initial deployment) a minor update would not succeed. The new logic allows targetVersiont to be switched without following the minor update workflow sequence and immediately updates the images on the Controlplane.

This is a rare case but a user actually hit it when updating the operators with failed/incomplete dataplane resources.

Jira: OSPRH-10645


stuggi (approved):
/lgtm

dprince:
test failure here looks valid although they weren't happening before. Will have to take a look

PR #1130, opened 2024-10-11
Update Samples and configure Swift backend for Glance

fmount:
The unsupported `File` backend is currently used in most of the samples. They are not consistent even with the documentation, where we suggest to add `replicas: 0` to the glance service until a backend is chosen. Considering we have `Swift` deployed by default, this patch fixes the `Glance` samples and adds `Swift` as a backend. This change will be reflected in the openstack-operator guide as well.

### DIFF
--- a/config/samples/base/openstackcontrolplane/core_v1beta1_openstackcontrolplane.yaml
+++ b/config/samples/base/openstackcontrolplane/core_v1beta1_openstackcontrolplane.yaml
@@ -56,13 +56,24 @@ spec:
     template:
       secret: osp-secret
       databaseInstance: openstack
+      customServiceConfig: |
+        [DEFAULT]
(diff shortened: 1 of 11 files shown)


fmount:
This change might be required to not mess w/ CI if we land https://github.com/openstack-k8s-operators/glance-operator/pull/440

stuggi:
@fmount this looks good to me. how do you want to proceed? should we get this in? or wait until after the next branching?

fmount:
> @fmount this looks good to me. how do you want to proceed? should we get this in? or wait until after the next branching?

Because we're just changing samples, I'm not concerned about landing them at a particular time frame.
+1 to merge this patch and move forward.
Thank you @stuggi for looking at that!

fao89 (commented):
+2

PR #1147, opened 2024-10-21
Switch EDPMServiceType to Enum

bshephar:
This change switches the EDPMServiceType parameter to be an Enum that accepts a fixed list of service names. This is required, since allowing arbitrary names can interfere with the TLS mounts desribed in the referenced Jira.

Jira: https://issues.redhat.com/browse/OSPRH-10804

### DIFF
--- a/pkg/dataplane/service.go
+++ b/pkg/dataplane/service.go
@@ -196,7 +196,7 @@ func dedupe(ctx context.Context, helper *helper.Helper,
 			return dedupedServices, updatedglobalServices, err
 
 		}
-		if !slices.Contains(nodeSetServiceTypes, service.Spec.EDPMServiceType) && !slices.Contains(dedupedServices, svc) {
+		if !slices.Contains(nodeSetServiceTypes, service.Name) && !slices.Contains(dedupedServices, svc) {
 			if service.Spec.DeployOnAllNodeSets {
 				if !slices.Contains(globalServices, svc) {
 					updatedglobalServices = append(globalServices, svc)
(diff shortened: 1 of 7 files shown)


pkg/dataplane/service.go
```diff
@@ -196,7 +196,7 @@ func dedupe(ctx context.Context, helper *helper.Helper,
 			return dedupedServices, updatedglobalServices, err
 
 		}
-		if !slices.Contains(nodeSetServiceTypes, service.Spec.EDPMServiceType) && !slices.Contains(dedupedServices, svc) {
+		if !slices.Contains(nodeSetServiceTypes, service.Name) && !slices.Contains(dedupedServices, svc) {
```
rabi, line 199: This does not look correct and I'm not sure what're we trying to achieve here. 

a. We already dedupe based on service names in the services list (see the second part of the if statement i.e `!slices.Contains(dedupedServices, svc)` ). 

b. If `EDPMServiceType` is not provided it defaults to the service name and we want to ensure that there are no duplicates for the same service type. If we're saying that `generic` service types can be repeated, then it won't be good user experience as we can't deduplicate them and users may end up with same services with different names and of `generic` service type.

If we still want to go with this approach for whatver reason then we've to ensure that we only allow duplicates for `generic` service type.

bshephar: The issue we're trying to solve for is defined here:
https://issues.redhat.com/browse/OSPRH-10733

We don't necessarily need to do exactly this. This was just one idea floated, and I didn't really know how we were using EDPMServiceType, so this PR just gives us something to talk about based on my exploration of using a fixed enumerable list of possible values for the service.

Ultimately, arbitrary names for the EDPMServiceType will be incompatible with how we're defining the Volumes for TLS. So we either need to do something with EDPMServiceType, or we need to change how we're mounting CA certs to be independent of EDPMServiceType to ensure it is functional with custom services.

PR #1149, opened 2024-10-21
[18.0-fr1][make]Update force-bump target for 18.0-fr1

gibizer:


dprince:
@gibizer once the commits on main move forwards the branch should diverge. But this will take time.

PR #1150, opened 2024-10-22
Reuse OpenstackClientImage instead of dedicated InstanceHaImage

lmiccini:
This resolves the warning at "make bundle" time:

```
WARN[0001] warning: multiple related images with the same image ref, "infra-instance-ha-image-url-default", and different names found.The image will only be listed once with an empty name.It is recmmended to either remove the duplicate or use the exact same name. 
```

while preserving the needed configmap:

```
[cifmw@haa-09 install_yamls]$ oc get cm infra-instanceha-config -o yaml
apiVersion: v1
data:
  instanceha-image: quay.io/podified-antelope-centos9/openstack-openstackclient@sha256:403c4f660f1b704d9ebbb9838c711ec6d20ca7183c63b2f224ad450f5ec37cf2
kind: ConfigMap
metadata:
  creationTimestamp: "2024-10-22T04:39:21Z"
  name: infra-instanceha-config
  namespace: openstack
  ownerReferences:
    blockOwnerDeletion: true
    controller: true
    kind: OpenStackControlPlane
    name: openstack-galera-network-isolation
    uid: cee79e70-639c-4974-acc5-5ec8f75ca299
  resourceVersion: "51877"
  uid: 8dd46abb-0cf1-414b-a3d7-6a3f34a9bb56
```

### DIFF
--- a/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
+++ b/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
@@ -18373,8 +18373,6 @@ spec:
                     type: string
                   infraDnsmasqImage:
                     type: string
-                  infraInstanceHaImage:
-                    type: string
                   infraMemcachedImage:
                     type: string
                   infraRedisImage:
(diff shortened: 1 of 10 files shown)


stuggi (approved):
/lgtm

PR #1154, opened 2024-10-25
Add nodeselector for ansible jobs

rabi:
This way we can run the jobs only on nodes where required NAD networks are available.

jira: [OSPRH-10939](https://issues.redhat.com//browse/OSPRH-10939)


fao89 (approved):

fao89:
recheck

PR #1163, opened 2024-10-29
Set Heat timeout

fao89:
Depends-On: https://github.com/openstack-k8s-operators/heat-operator/pull/463

ref OSPRH-10965


bshephar:
If we enable Heat in functional tests: https://github.com/openstack-k8s-operators/openstack-operator/pull/1164

Then we can validate that this route annotation is applied like we do with others here:
https://github.com/openstack-k8s-operators/openstack-operator/blob/main/tests/functional/ctlplane/openstackoperator_controller_test.go#L578

fao89:
recheck

fao89:
recheck

bshephar (approved):

fao89:
need to wait https://github.com/openstack-k8s-operators/heat-operator/pull/487 to cherrypick

PR #1167, opened 2024-10-30
Fix ansible job error reason

fao89:
closes [OSPRH-11068](https://issues.redhat.com//browse/OSPRH-11068)

### DIFF
--- a/pkg/dataplane/deployment.go
+++ b/pkg/dataplane/deployment.go
@@ -195,18 +195,18 @@ func (d *Deployer) ConditionalDeploy(
 
 	}
 
-	var ansibleCondition *batchv1.JobCondition
+	var ansibleCondition batchv1.JobCondition
 	if nsConditions.IsFalse(readyCondition) {
-		var ansibleEE *batchv1.Job
+		var ansibleJob *batchv1.Job
 		_, labelSelector := dataplaneutil.GetAnsibleExecutionNameAndLabels(&foundService, d.Deployment.Name, d.NodeSet.Name)
-		ansibleEE, err = dataplaneutil.GetAnsibleExecution(d.Ctx, d.Helper, d.Deployment, labelSelector)
+		ansibleJob, err = dataplaneutil.GetAnsibleExecution(d.Ctx, d.Helper, d.Deployment, labelSelector)
 		if err != nil {
 			// Return nil if we don't have AnsibleEE available yet
 			if k8s_errors.IsNotFound(err) {
 				log.Info(fmt.Sprintf("%s AnsibleEE job is not yet found", readyCondition))
 				return nil
 			}
-			log.Error(err, fmt.Sprintf("Error getting ansibleEE job for %s", deployName))
+			log.Error(err, fmt.Sprintf("Error getting ansibleJob job for %s", deployName))
 			nsConditions.Set(condition.FalseCondition(
 				readyCondition,
 				condition.ErrorReason,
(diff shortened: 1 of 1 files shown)


pkg/dataplane/deployment.go
```diff
@@ -215,33 +214,34 @@ func (d *Deployer) ConditionalDeploy(
...
-		if ansibleEE.Status.Succeeded > 0 {
+		if ansibleJob.Status.Succeeded > 0 {
 			log.Info(fmt.Sprintf("Condition %s ready", readyCondition))
 			nsConditions.Set(condition.TrueCondition(
 				readyCondition,
 				readyMessage))
-		} else if ansibleEE.Status.Active > 0 {
-			log.Info(fmt.Sprintf("AnsibleEE job is not yet completed: Execution: %s, Active pods: %d", ansibleEE.Name, ansibleEE.Status.Active))
+		} else if ansibleJob.Status.Active > 0 {
+			log.Info(fmt.Sprintf("AnsibleEE job is not yet completed: Execution: %s, Active pods: %d", ansibleJob.Name, ansibleJob.Status.Active))
```
olliewalsh, line 223: should this block also apply when `ansibleEE.Status.Failed <= *ansibleEE.Spec.BackoffLimit`? since "AnsibleEE job is not yet complete" is true in that case

pkg/dataplane/deployment.go
```diff
@@ -215,33 +214,34 @@ func (d *Deployer) ConditionalDeploy(
...
 			nsConditions.Set(condition.FalseCondition(
 				readyCondition,
 				condition.RequestedReason,
 				condition.SeverityInfo,
 				readyWaitingMessage))
-		} else if ansibleEE.Status.Failed > 0 {
-			errorMsg := fmt.Sprintf("execution.name %s execution.namespace %s failed pods: %d", ansibleEE.Name, ansibleEE.Namespace, ansibleEE.Status.Failed)
-			for _, condition := range ansibleEE.Status.Conditions {
+		} else if ansibleJob.Status.Failed > 0 {
```
olliewalsh, line 229: and this become `else if ansibleJob.Status.Failed > *ansibleEE.Spec.BackoffLimit`

pkg/dataplane/deployment.go
fao89: I think we should keep updating the status when the job fails, so the human operator can check if there is something wrong or just wait for the retry

pkg/dataplane/deployment.go
bshephar: I think `else if ansibleJob.Status.Failed = *ansibleEE.Spec.BackoffLimit`. It wouldn't be greater than the BackoffLimit would it?

olliewalsh: It would, BackoffLimit is max number of retries

pkg/dataplane/deployment.go
```diff
@@ -215,33 +214,34 @@ func (d *Deployer) ConditionalDeploy(
...
+			log.Info(fmt.Sprintf("AnsibleEE job is not yet completed: Execution: %s, Active pods: %d", ansibleJob.Name, ansibleJob.Status.Active))
 			nsConditions.Set(condition.FalseCondition(
 				readyCondition,
 				condition.RequestedReason,
 				condition.SeverityInfo,
 				readyWaitingMessage))
-		} else if ansibleEE.Status.Failed > 0 {
-			errorMsg := fmt.Sprintf("execution.name %s execution.namespace %s failed pods: %d", ansibleEE.Name, ansibleEE.Namespace, ansibleEE.Status.Failed)
-			for _, condition := range ansibleEE.Status.Conditions {
+		} else if ansibleJob.Status.Failed == *ansibleJob.Spec.BackoffLimit {
```
bshephar, line 229: @olliewalsh wdyt about this approach. I don't think Failed should be > than BackoffLimit. But this should ensure that we don't enter the code block where we reference `Reason` before it has been added to the Job status.

rabi:
I think what @owalsh suggested makes sense. failed >0 (failed pods) does not mean failed job and we don't need to tell the user that the job has failed. I worked on a change like that yesterday but forgot to push it https://github.com/openstack-k8s-operators/openstack-operator/compare/main...rabi:openstack-operator:fix_nil?expand=1

slagle (approved):

olliewalsh (approved):

PR #1168, opened 2024-10-31
[kuttl] Add tls scenarios for non API services

Deydra71:


jpodivin:
Hi. The script itself is working. It reports mismatches and other issues as expected. However, do we want the test to pass even if there are issues? Because as it is, the step is just going to deposit info about it in logs.

jpodivin (commented):
Hi. The script itself is working. It reports mismatches and other issues as expected. However, do we want the test to pass even if there are issues? Because as it is, the step is just going to deposit info about it in logs.

Deydra71:
@jpodivin 
Hello, yes I know. I keep it like this until I get a review from @stuggi  or @olliewalsh what's their preferred approach

olliewalsh:
> Hi. The script itself is working. It reports mismatches and other issues as expected. However, do we want the test to pass even if there are issues? Because as it is, the step is just going to deposit info about it in logs.

? returns 1 IIUC - https://github.com/openstack-k8s-operators/openstack-operator/pull/1168/files#diff-291551847705ed67b32c66c3659c73adee6d1b07693a5b08ab9f288f274cd0e7R57

Deydra71:
@olliewalsh  It was a misunderstanding that the script returns 1, but the kuttl doesn't fail immediately, but tries again until the timeout. All clear now.

stuggi (commented):
looks good to me, but maybe we could extend it for the db ?

stuggi:
> looks good to me, but maybe we could extend it for the db ?

strange my inline comment was lost. for the db can use `openssl s_client -starttls mysql -connect ...`

stuggi (approved):
/lgtm

PR #1176, opened 2024-11-08
Require EDPMServiceType

bshephar:
This change switches EDPMServiceType from optional to required. Since the EDPMServiceType of custom services needs to match that of an existing service, we also remove the defaulting mechanism in the webhook.

jira: https://issues.redhat.com/browse/OSPRH-11333

### DIFF
--- a/docs/assemblies/proc_creating-a-custom-service.adoc
+++ b/docs/assemblies/proc_creating-a-custom-service.adoc
@@ -112,7 +112,7 @@ spec:
   deployOnAllNodeSets: true
 ----
 
-. Optional: Specify the `edpmServiceType` field for the service. Different custom services may use the same ansible content to manage the same EDPM service (such as `ovn` or `nova`). The `DataSources`, TLS certificates, and CA certificates need to be mounted at the same locations so they can be found by the ansible content even when using a custom service. `edpmServiceType` is used to create this association. The value is the name of the default service that uses the same ansible content as the custom service. If there are multiple services with the same `edpmServiceType` listed in a nodeset or deployment spec, latter ones would be ignored.
+. Required: Specify the `edpmServiceType` field for the service. Different custom services may use the same ansible content to manage the same EDPM service (such as `ovn` or `nova`). The `DataSources`, TLS certificates, and CA certificates need to be mounted at the same locations so they can be found by the ansible content even when using a custom service. `edpmServiceType` is used to create this association. The value is the name of the default service that uses the same ansible content as the custom service. If there are multiple services with the same `edpmServiceType` listed in a nodeset or deployment spec, latter ones would be ignored.
 +
 For example, a custom service that uses the `edpm_ovn` ansible content from `edpm-ansible` would set `edpmServiceType` to `ovn`, which matches the default `ovn` service name provided by `openstack-operator`.
 +
--- a/apis/bases/dataplane.openstack.org_openstackdataplaneservices.yaml
+++ b/apis/bases/dataplane.openstack.org_openstackdataplaneservices.yaml
@@ -129,6 +129,8 @@ spec:
                   - contents
                   type: object
                 type: object
+            required:
+            - edpmServiceType
             type: object
           status:
             properties:
--- a/apis/dataplane/v1beta1/openstackdataplaneservice_types.go
+++ b/apis/dataplane/v1beta1/openstackdataplaneservice_types.go
@@ -114,7 +114,8 @@ type OpenStackDataPlaneServiceSpec struct {
 	// corresponds to the ansible role name (without the "edpm_" prefix) used
 	// to manage the service. If not set, will default to the
 	// OpenStackDataPlaneService name.
-	EDPMServiceType string `json:"edpmServiceType,omitempty" yaml:"edpmServiceType,omitempty"`
+	// +kubebuilder:validation:Required
+	EDPMServiceType string `json:"edpmServiceType" yaml:"edpmServiceType"`
 }
 
 // OpenStackDataPlaneServiceStatus defines the observed state of OpenStackDataPlaneService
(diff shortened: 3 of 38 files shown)


docs/assemblies/proc_creating-a-custom-service.adoc
```diff
@@ -112,7 +112,7 @@ spec:
...
+. Required: Specify the `edpmServiceType` field for the service. Different custom services may use the same ansible content to manage the same EDPM service (such as `ovn` or `nova`). The `DataSources`, TLS certificates, and CA certificates need to be mounted at the same locations so they can be found by the ansible content even when using a custom service. `edpmServiceType` is used to create this association. The value is the name of the default service that uses the same ansible content as the custom service. If there are multiple services with the same `edpmServiceType` listed in a nodeset or deployment spec, latter ones would be ignored.
```
slagle, line 115: It's not actually required that edpmServiceType matches that of an existing service. It's really just if you are re-using existing ansible content from epdm-ansible, and need to re-use the same paths for cert mounts. So, I think that needs to be clarified. The field can still be required though, so that we can force the user to make that distinction when they create the service.

bshephar: I'm still kind of on the fence as to whether we should do this or not. Maybe it can just be improved in documentation. But I'll note my thoughts on the topic below since we have a thread here related to it.

The only thing that concerns me with the current implementation is that it doesn't work if users try to create a new Nova service for example, but fail to specify `edpmServiceType: nova`. I think it probably improves the user experience by requiring the field. But then at the same time, it also puts more of a mental burden on them to now go and understand what this `edpmServiceType` means.

I think it's probably easier for the user if we block the CR creation by requiring this field, then they have to go and read about the parameters. Verse the alternative of having to debug why the deployment failed, which is technically more challenging / a worse user experience.


I agree with this comment, I'm just dropping my thoughts on the thread. I'll do some rewriting of this section to include that clarification.

bshephar:
Functional tests pass locally, rechecking it:
```
 SUCCESS! 1m11.00949s
coverage: 50.5% of statements
[1734330348] Webhook Suite - 0/0 specs - 4 procs  SUCCESS! 2.333ms
coverage: 1.0% of statements
[1734330348] Controller v1beta1 Suite - 3/3 specs - 4 procs ••• SUCCESS! 2.88ms
coverage: 0.6% of statements
composite coverage: 62.6% of statements

Ginkgo ran 4 suites in 1m41.194835792s
Test Suite Passed
```

bshephar:
So, the remaining problem we have here is that in making this field mandatory, we risk the update process. Since CR's defined without the field will be un-upgradable until the field is defined.

I don't think this should be an issue, since we were adding the field with a validation webhook, so all existing CR's in the cluster _should_ have the field. But we should consider it as a potential upgrade issue and test for it accordingly.

bshephar:
All Kuttl tests seem to be having problems with Quay:
```
4d9a29bd7a80877d8e
Copying blob sha256:8afba38a614e150d5cec9d32c41bedfe2fe3cc1a11fbd33c712e10abb9e65de9
time="2024-12-18T03:43:50Z" level=warning msg="Failed, retrying in 4s ... (3/3). Error: trying to reuse blob sha256:8afba38a614e150d5cec9d32c41bedfe2fe3cc1a11fbd33c712e10abb9e65de9 at destination: pinging container registry quay.rdoproject.org: Get \"https://quay.rdoproject.org/v2/\": dial tcp 38.129.56.158:443: i/o timeout"
Getting image source signatures
Copying blob sha256:3d7039b254a490f48fac42b072ca3e2249ff9b4b6bb65cd1930b4ff94600a86e
Copying blob sha256:8afba38a614e150d5cec9d32c41bedfe2fe3cc1a11fbd33c712e10abb9e65de9
Copying blob sha256:34db82eb2cd0f20a0d8e43721c3dc6e7c2900ee73351f94d9a29bd7a80877d8e
Warning: Push failed, retrying in 5s ...
Registry server Address: 
Registry server User Name: openstack-k8s-operators+cirobot
Registry server Email: 
Registry server Password: <<non-empty>>
error: build error: Failed to push image: trying to reuse blob sha256:8afba38a614e150d5cec9d32c41bedfe2fe3cc1a11fbd33c712e10abb9e65de9 at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout
```

slagle:
> So, the remaining problem we have here is that in making this field mandatory, we risk the update process. Since CR's defined without the field will be un-upgradable until the field is defined.
> 
> I don't think this should be an issue, since we were adding the field with a validation webhook, so all existing CR's in the cluster _should_ have the field. But we should consider it as a potential upgrade issue and test for it accordingly.

Given the webhook was in GA, I think we're good in this regard. Any other concerns before we merge this one?

bshephar:
The kuttl test looks legit after rebase. Let me check that:
```
 Container test exited with code 2, reason Error
---
name: nova-combined-ca-bundle
        -        - mountPath: /var/lib/openstack/cacerts/custom-global-service
        +        - mountPath: /var/lib/openstack/cacerts
                   name: custom-global-service-combined-ca-bundle 
```

bshephar:
Latest kuttl failure is a quay issue:
```
b56a76e38e2e028020458a3016d3b at destination: pinging container registry quay.rdoproject.org: Get \"https://quay.rdoproject.org/v2/\": dial tcp 38.129.56.158:443: i/o timeout"
Getting image source signatures
```

stuggi:
@bshephar iirc this was moved to be documented, correct? could close this PR?

bshephar:
> @bshephar iirc this was moved to be documented, correct? could close this PR?

Yeah, just documented instead. Closing

PR #1184, opened 2024-11-13
Use new nad.EnsureNetworksAnnotation() func to get NAD annotations

stuggi:
For BGP setup there is the need to set the default gateway to the additional interface defined via the multus annotations. To allow this a user can configure `ipam.gateway` in the NAD. EnsureNetworksAnnotation() will override the pod network default route by reading the NAD. If `ipam.gateway` is defined and not "", it gets set on the networks annotation as the `default-route`.

Jira: https://issues.redhat.com/browse/OSPRH-8680

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/579


apis/go.mod
```diff
@@ -115,3 +115,35 @@ replace github.com/openshift/api => github.com/openshift/api v0.0.0-202408300231
...
+
+replace github.com/openstack-k8s-operators/octavia-operator/api => github.com/stuggi/octavia-operator/api v0.0.0-20241113150020-d74ecfeabff2
+
+replace github.com/openstack-k8s-operators/ovn-operator/api => github.com/stuggi/ovn-operator/api v0.0.0-20241113150626-0823d992f39c
+
+replace github.com/openstack-k8s-operators/placement-operator/api => github.com/stuggi/placement-operator/api v0.0.0-20241113150744-31cc26a7465e
+
+replace github.com/openstack-k8s-operators/swift-operator/api => github.com/stuggi/swift-operator/api v0.0.0-20241113150817-8e1dcd67ff28
+
+replace github.com/openstack-k8s-operators/test-operator/api => github.com/stuggi/test-operator/api v0.0.0-20241113150948-d549cb3236e3
```
stuggi, line 149: needs an update when all those deps are merged

fernandoroyosanchez:
I think this was forgotten here https://github.com/openstack-k8s-operators/openstack-operator/blame/main/config/manager/kustomization.yaml#L15

stuggi:
> I think this was forgotten here https://github.com/openstack-k8s-operators/openstack-operator/blame/main/config/manager/kustomization.yaml#L15

yes, thanks for the pointer. will be fixed with https://github.com/openstack-k8s-operators/openstack-operator/pull/1186

PR #1187, opened 2024-11-15
Inherit top-level nodeSelector in nova template

olliewalsh:
Depends-On: https://github.com/openstack-k8s-operators/nova-operator/pull/894
Jira: [OSPRH-10734](https://issues.redhat.com//browse/OSPRH-10734)


gibizer (approved):
Thanks for fixing it.

PR #1188, opened 2024-11-15
Inherit top-level nodeSelector in placement, horizon, keystone template

olliewalsh:
Depends-On: https://github.com/openstack-k8s-operators/placement-operator/pull/266
Depends-On: https://github.com/openstack-k8s-operators/horizon-operator/pull/388
Depends-On: https://github.com/openstack-k8s-operators/keystone-operator/pull/502
Jira: [OSPRH-10734](https://issues.redhat.com//browse/OSPRH-10734)

### DIFF
--- a/pkg/openstack/placement.go
+++ b/pkg/openstack/placement.go
@@ -41,6 +41,10 @@ func ReconcilePlacementAPI(ctx context.Context, instance *corev1beta1.OpenStackC
 		instance.Spec.Placement.Template = &placementv1.PlacementAPISpecCore{}
 	}
 
+	if instance.Spec.Placement.Template.NodeSelector == nil {
+		instance.Spec.Placement.Template.NodeSelector = &instance.Spec.NodeSelector
+	}
+
 	// add selector to service overrides
 	for _, endpointType := range []service.Endpoint{service.EndpointPublic, service.EndpointInternal} {
 		if instance.Spec.Placement.Template.Override.Service == nil {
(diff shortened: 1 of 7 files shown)


pkg/openstack/placement.go
```diff
@@ -108,9 +112,6 @@ func ReconcilePlacementAPI(ctx context.Context, instance *corev1beta1.OpenStackC
 		if placementAPI.Spec.Secret == "" {
 			placementAPI.Spec.Secret = instance.Spec.Secret
 		}
-		if placementAPI.Spec.NodeSelector == nil && instance.Spec.NodeSelector != nil {
-			placementAPI.Spec.NodeSelector = instance.Spec.NodeSelector
-		}
```
gibizer, line 113: I thought we set it inside the createOrPatch to actually ensure it is updated later not just during initial created.

stuggi: we apply it to the `instance.Spec.Placement.Template` which added via the DeepCopyInto within the `CreateOrPatch` a couple of lines above this. so it should be the same as having this

olliewalsh: `placementAPI.Spec.NodeSelector == nil` can only be true on the first invocation

olliewalsh: `DeepCopyInto` will reset it to nil though, so probably did work in this case. IIRC it wasn't correct for some services

gibizer:
looks good, just the replace line needs to be removed

abays (approved):
/lgtm

PR #1202, opened 2024-11-22
test crd schema checker

gibizer:
- **[make][pre-commit]Check CRD schema to avoid update issues**
- **DNM:test make crd-schema-check**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,6 +25,12 @@ repos:
       entry: make
       args: ['operator-lint']
       pass_filenames: false
+    - id: make-crd-schema-check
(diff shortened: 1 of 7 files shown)


gibizer:
The pre-commit check should have caught the CRD change but it did not. The make target catches is for me locally so something is with the CI job.

gibizer:
I see the issue when the pre-commit runs in the CI the proposed commit already placed on the main branch of he repo:
```
    Merge commit '5975d9fde9b40d1a1a93fc53d469d3321fa25f91'
```

So the schema diff between the proposed commit and main is always empty in CI. I can try to be more specific using origin/main instead of main...

PR #1207, opened 2024-11-25
[Designate] Add Mdns replica count

omersch381:
DesignateMdns deployment type is being changed to stateful set.

This commit adds replica count to Mdns in all the Designate CRs.

### DIFF
--- a/config/samples/base/openstackcontrolplane/core_v1beta1_openstackcontrolplane.yaml
+++ b/config/samples/base/openstackcontrolplane/core_v1beta1_openstackcontrolplane.yaml
@@ -205,7 +205,7 @@ spec:
       designateProducer:
         replicas: 1 # backend needs to be configured
       designateMdns:
-        replicas: 0 # backend needs to be configured
+        replicas: 1 # backend needs to be configured
       designateBackendbind9:
         replicas: 1 # backend needs to be configured
         storageClass: local-storage
--- a/config/samples/core_v1beta1_openstackcontrolplane_galera.yaml
+++ b/config/samples/core_v1beta1_openstackcontrolplane_galera.yaml
@@ -232,7 +232,7 @@ spec:
       designateProducer:
         replicas: 1
       designateMdns:
-        replicas: 0
+        replicas: 1
         networkAttachments:
           - designate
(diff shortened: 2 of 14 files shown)


omersch381:
https://github.com/openstack-k8s-operators/designate-operator/pull/255 will have to get merged first.

jpodivin (changes requested):
Hi. Could you add an explicit depends on to this PR? Just so it's obvious that it's missing a dependency. 

Also, it would be great to have a link to JIRA item, or some other description of what are you implementing and why. Thanks a lot.

omersch381:
> me other description of what are you implementing and why. Thanks a lot.

Sure, thanks for raising it.

omersch381:
The Designate-operator patch has been merged yesterday, and it added the Replicas field to the API directory files, I wonder when CI will stop failing because of lacking it.

bshephar (approved):

config/samples/base/openstackcontrolplane/core_v1beta1_openstackcontrolplane.yaml
```diff
@@ -205,7 +205,7 @@ spec:
       designateProducer:
         replicas: 1 # backend needs to be configured
       designateMdns:
-        replicas: 0 # backend needs to be configured
+        replicas: 1 # backend needs to be configured
```
dprince, line 208: comment here says backend needs to be configured still. If you are setting replicas to 1 does this (and if functionally works) does this comment still make sense?

omersch381: It looks like it is a generic comment that appear on many files. We are still configuring MiniDNS backend on the designate-operator repository, so I think it still applies.
This openstack-operator patch mainly saves us time, so we won't have to modify the csv manually with the same values. We modify its value to make sure changes in the current replica count will behave correctly with our changes.

omersch381:
https://github.com/openstack-k8s-operators/openstack-operator/pull/1204 has been merged

omersch381:
recheck podified-multinode-edpm-deployment-crc

PR #1211, opened 2024-11-28
[openstackclient] Store hash to determine when pod spec changes

olliewalsh:
Comparing the desired pod spec to the current pod spec is not reliable as
the spec can be mutated by k8s e.g to add volume mount for a service account
or to add the default nodeselector from the namespace/cluster on OpenShift

Instead store a hash of the original pod spec and compare this to the hash
of the current desired pod spec to determine if the spec has changed.

Resolves: [OSPRH-11701](https://issues.redhat.com//browse/OSPRH-11701)


controllers/client/openstackclient_controller.go
```diff
@@ -307,17 +308,23 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		return ctrl.Result{}, err
+	}
+
+	podSpecAnnotation := labels.GetGroupLabel("openstackclient") + "/pod-spec-hash"
+
 	op, err := controllerutil.CreateOrPatch(ctx, r.Client, osclient, func() error {
 		isPodUpdate := !osclient.ObjectMeta.CreationTimestamp.IsZero()
-		if !isPodUpdate {
+		currentPodSpecHash, hasPodSpecHash := osclient.Annotations[podSpecAnnotation]
+		if !isPodUpdate || (hasPodSpecHash && currentPodSpecHash != podSpecHash) {
```
olliewalsh, line 321: first time this reconciles it will not update the spec, but will add the hash annotation. From then on it will update the spec when the hash changes (which will almost definitely result in a forbidden error)

my only concern is that the spec could have changed for that initial reconcile, so maybe worth asserting the spec in that case too...

controllers/client/openstackclient_controller.go
```diff
@@ -307,17 +308,23 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	spec := openstackclient.ClientPodSpec(ctx, instance, helper, configVarsHash)
 
+	podSpecHash, err := util.ObjectHash(spec)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	podSpecAnnotation := labels.GetGroupLabel("openstackclient") + "/pod-spec-hash"
```
olliewalsh, line 316: Note: replicaset uses a label vs an annotation but a label is too short to store our object hash string. Replicaset uses a shorter hash type but then needs logic to handle hash collisions

olliewalsh:
recheck

stuggi (commented):
lgtm

PR #1212, opened 2024-11-29
When OVNController is disabed, set current version images

stuggi:
When there is no nicMappings configured, the OVNController CR gets cleanup, because there is no need to run one on the ctlplane. In this case, ReconcileOVNController deletes the CR and returns ctrl.Result{}, nil .
`status.deployedOVNVersion:` on the ctlplane gets set correct to the new version, but there was no change for the images on
* instance.Status.ContainerImages.OvnControllerImage
* instance.Status.ContainerImages.OvnControllerOvsImage

As a result the minor update process gets stuck and never proceeds.

Jira: [OSPRH-12063](https://issues.redhat.com//browse/OSPRH-12063)


karelyatin (commented):
/lgtm

PR #1215, opened 2024-12-02
Use dns cluster info from lib common get function

averdagu:
Openshift coreDNS creates the domain name using an string located in dnses.operator.openshift.io. This string can change in the future, calling lib-common/GetDNSClusterDomain the responsability of gathering this information correctly only falls under lib-common intead of all operators.

Depends-on: openstack-k8s-operators/lib-common#580
Related-Issue: [OSPRH-3627](https://issues.redhat.com//browse/OSPRH-3627)


stuggi (approved):
/lgtm

PR #1217, opened 2024-12-02
Add commit hash as suffix to the OpenStackVersion

stuggi:
Right now when a PR merges, the build openstack-operator has the OpenStackVersion 0.0.1. Since they are all the same for each PR it is not possible to test updates. Lets add the commit hash as a suffix so that we get a different OpenStackVersion with each operator build.


controllers/core/openstackversion_controller.go
```diff
@@ -301,6 +301,15 @@ func (r *OpenStackVersionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		Log.Info("Setting DeployedVersion")
 		instance.Status.DeployedVersion = &instance.Spec.TargetVersion
 	}
+	if instance.Status.AvailableVersion != nil &&
+		instance.Status.DeployedVersion != nil &&
+		*instance.Status.AvailableVersion != *instance.Status.DeployedVersion {
+		instance.Status.Conditions.Set(condition.TrueCondition(
+			corev1beta1.OpenStackVersionMinorUpdateAvailable,
+			corev1beta1.OpenStackVersionMinorUpdateAvailableMessage))
```
dprince, line 309: Don't we need to Init this new condition up top? Also could we add a functional test to verify this?

controllers/core/openstackversion_controller.go
```diff
@@ -301,6 +301,15 @@ func (r *OpenStackVersionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		Log.Info("Setting DeployedVersion")
 		instance.Status.DeployedVersion = &instance.Spec.TargetVersion
 	}
+	if instance.Status.AvailableVersion != nil &&
```
dprince, line 304: I don't think it is possible for AvailableVersion to be nil here.

dprince (changes requested):

dprince:
I'm not sure we strictly need this new condition to implement upgrade testing. If we add it we'll have it on main, but not on the FR1 maintenance branch so we won't be able to use the job there.

It is enough to just check that the new catalog/index is deployed and that OpenStackVersion is in a ReadyCondition True state. Then if targetVersion != availableVersion you'd know there is an update (and this is shown clearly on the CLI too)

Makefile
```diff
@@ -5,6 +5,8 @@
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= 0.0.1
 
+OPENSTACK_RELEASE_VERSION ?= $(VERSION)-$(shell git rev-parse --short HEAD)
```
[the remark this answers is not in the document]
stuggi: should we just use a timestamp?

controllers/core/openstackversion_controller.go
stuggi: the condition is only set when there is an update available. if not, we remove it in the else. so don't think we have to init it.

stuggi: will add test

stuggi:
> I'm not sure we strictly need this new condition to implement upgrade testing. If we add it we'll have it on main, but not on the FR1 maintenance branch so we won't be able to use the job there.
> 
> It is enough to just check that the new catalog/index is deployed and that OpenStackVersion is in a ReadyCondition True state. Then if targetVersion != availableVersion you'd know there is an update (and this is shown clearly on the CLI too)

having the condition makes it a lot easier, as you could see from my install_yamls pr. it would be a single simple `oc wait` command, instead of checking the ReadyCondition, which is not really useful as it is True already from the previous update. so we'd have to check the deployedVersion or targetVersion against the availableVersion with a  loop. we could really simplify this with this condition. Right now I think the initial focus is to test against main. but yes for running this job against latest in FR1 we have to backport this, which I think is not an issue.

Makefile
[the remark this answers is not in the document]
stuggi: also for the Makefile OPENSTACK_RELEASE_VERSION its just the default, you can set to whatever is required downstream if the Makefile is used

controllers/core/openstackversion_controller.go
stuggi: done

Makefile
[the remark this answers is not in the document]
stuggi: changed it to be the epoch

dprince (approved):

PR #1224, opened 2024-12-09
Determine Netmask from IP CIDR

bshephar:
jira: https://issues.redhat.com/browse/OSPRH-12303

### DIFF
--- a/pkg/dataplane/baremetal.go
+++ b/pkg/dataplane/baremetal.go
@@ -89,6 +89,7 @@ func DeployBaremetalSet(
 						return fmt.Errorf("%s gateway is missing", dataplanev1.CtlPlaneNetwork)
 					}
 					baremetalSet.Spec.CtlplaneGateway = *res.Gateway
+					baremetalSet.Spec.CtlplaneNetmask = string(ipNet.Mask)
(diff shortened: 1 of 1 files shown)


pkg/dataplane/baremetal.go
```diff
@@ -89,6 +89,7 @@ func DeployBaremetalSet(
 						return fmt.Errorf("%s gateway is missing", dataplanev1.CtlPlaneNetwork)
 					}
 					baremetalSet.Spec.CtlplaneGateway = *res.Gateway
+					baremetalSet.Spec.CtlplaneNetmask = string(ipNet.Mask)
```
rabi, line 92: I think `ctlplaneNetmask` is not used at all atm. ctlplaneIP is in CIDR format[1] and that's used instead in openstack-baremetal-operator. 

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/dataplane/baremetal.go#L87

rabi (changes requested):

pkg/dataplane/baremetal.go
bshephar: Ok, let's remove it from baremetal-operator then, and that should solve this problem that I was trying to address. So we can just close this PR.

PR #1226, opened 2024-12-09
[18.0-fr1][openstackclient] Store hash to determine when pod spec changes

olliewalsh:
Comparing the desired pod spec to the current pod spec is not reliable as
the spec can be mutated by k8s e.g to add volume mount for a service account
or to add the default nodeselector from the namespace/cluster on OpenShift

Instead store a hash of the original pod spec and compare this to the hash
of the current desired pod spec to determine if the spec has changed.

Cherry-pick of https://github.com/openstack-k8s-operators/openstack-operator/pull/1211
Resolves: [OSPRH-11701](https://issues.redhat.com//browse/OSPRH-11701)


stuggi (approved):
/lgtm

PR #1227, opened 2024-12-10
Support Roles on OpenStackDataPlaneService

fao89:
Additionally to playbooks, it enables running roles directly with ansible-runner

closes [OSPRH-12358](https://issues.redhat.com//browse/OSPRH-12358)

### DIFF
--- a/pkg/dataplane/util/ansibleee.go
+++ b/pkg/dataplane/util/ansibleee.go
@@ -22,6 +22,8 @@ type EEJob struct {
 	PlaybookContents string `json:"playbookContents,omitempty"`
 	// Playbook is the playbook that ansible will run on this execution, accepts path or FQN from collection
 	Playbook string `json:"playbook,omitempty"`
+	// Role is the role that ansible will run on this execution, accepts path or FQN from collection
+	Role string `json:"role,omitempty"`
 	// Image is the container image that will execute the ansible command
 	Image string `json:"image,omitempty"`
 	// Name is the name of the execution job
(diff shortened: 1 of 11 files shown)


bshephar (approved):

fao89:
recheck

fao89:
recheck

fao89:
recheck

pkg/dataplane/util/ansibleee.go
```diff
@@ -78,12 +80,17 @@ func (a *EEJob) JobForOpenStackAnsibleEE(h *helper.Helper) (*batchv1.Job, error)
 
 	args := a.Args
 
-	playbook := a.Playbook
+	artifact := a.Playbook
+	param := "-p"
 	if len(args) == 0 {
-		if len(playbook) == 0 {
-			playbook = CustomPlaybook
+		if len(a.PlaybookContents) > 0 {
```
rabi, line 86: I think this changes the old behavior. Earlier `Playbook` provided used to supersede `PlaybookContents`. Now it's the reverse and `role` provided supersedes both of them. Is that what we want now?

fao89: `playbook`, `playbookContents` and `role` are not mandatory fields, so we need to check which one isn't empty. Ideally, we shouldn't supersede, but it would happen when more than one field is filled. I think we might need a validation to ensure only one of the three fields is filed

rabi: Yeah either have a validation or ensure that there is proper prededence. playbook->playbookContent->role

fao89: I've updated the precedence

pkg/dataplane/util/ansibleee.go
```diff
@@ -78,12 +80,17 @@ func (a *EEJob) JobForOpenStackAnsibleEE(h *helper.Helper) (*batchv1.Job, error)
 
 	args := a.Args
 
-	playbook := a.Playbook
 	if len(args) == 0 {
-		if len(playbook) == 0 {
-			playbook = CustomPlaybook
+		artifact := a.Role
+		param := "-r"
+		if len(artifact) == 0 {
```
rabi, line 86: Are we still saying role if provided has the highest precedence?

fao89: oh! I misplaced the order

fao89:
recheck

rabi (commented):
/lgtm

PR #1229, opened 2024-12-16
Drop redhat service from default nodeset samples

karelyatin:
The service requires rhel nodes to run which we
don't have upstream and so is not being tested.
Also enabling this service requires additional
vars like activationkey,token, username etc.

Also it broke some downstream jobs which running
older version of operators and using these upstream samples.

Related-Issue: [OSPCIX-596](https://issues.redhat.com//browse/OSPCIX-596)

### DIFF
--- a/apis/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
+++ b/apis/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
@@ -900,7 +900,6 @@ spec:
                 type: integer
               services:
                 default:
-                - redhat
                 - download-cache
                 - bootstrap
                 - configure-network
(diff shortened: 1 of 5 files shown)


fao89:
it shouldn't break older versions, as the redhat service isn't on older samples: https://github.com/openstack-k8s-operators/openstack-operator/blob/18.0.0-proposed/config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml#L10-L25
if it is breaking, it means older versions are wrongly using the sample from the main branch

karelyatin:
> it shouldn't break older versions, as the redhat service isn't on older samples: https://github.com/openstack-k8s-operators/openstack-operator/blob/18.0.0-proposed/config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml#L10-L25 if it is breaking, it means older versions are wrongly using the sample from the main branch

Yes that is one of the reason to remove it from default samples. it can be handled downstream by using different branches. But removing it from default is due to other reasons in commit message (just adding the service is not sufficient as require other vars to be configured and it's not being tested upstream) , additionally it will also fix the downstream jobs.

karelyatin:
infra-operator pr merged

karelyatin:
storage cleanup

karelyatin:
quay.rdo

slagle (approved):
approving to unblock CI.
However, the other issue of downstream using the upstream samples from main should still be fixed separately. I just don't see that continuing to work long term.

PR #1056, opened 2024-09-10
Add support for KSM

paramite:
This patch add default image and TLS support for kube-state-metrics, a new component of Telemetry/Ceilometer resource.

Related: [OBSDA-574](https://issues.redhat.com/browse/OBSDA-574)
Related: [OSPRH-1052](https://issues.redhat.com/browse/OSPRH-1052)
Closes: [OSPRH-11115](https://issues.redhat.com/browse/OSPRH-11115)

PR #1236, opened 2024-12-18
[OSPRH-12079] Add mysqld exporter

vyzigold:
This adds support for deploying the mysqld exporter as part of ceilometer CR. This is very similar to https://github.com/openstack-k8s-operators/openstack-operator/pull/1056

Depends-On: https://github.com/openstack-k8s-operators/telemetry-operator/pull/571

### DIFF
--- a/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
+++ b/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
@@ -16739,6 +16739,8 @@ spec:
                     type: string
                   ceilometerIpmiImage:
                     type: string
+                  ceilometerMysqldExporterImage:
+                    type: string
                   ceilometerNotificationImage:
                     type: string
                   ceilometerProxyImage:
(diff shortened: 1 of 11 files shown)


vyzigold:
I saw that there were some concerns by @dprince on the KSM PR https://github.com/openstack-k8s-operators/openstack-operator/pull/1056 about adding new images to the openstack version. To be honest I'm not fully sure what the final resolution there is and what's required to be done to add new images.

I saw there were concerns about the image being set to nil in an old target version in the openstackversion. So I added a check for that. If it's nil, then openstack-operator will set the image to empty string and telemetry-operator will default it with its webhook to the image set here: https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/api/v1beta1/ceilometer_types.go#L44 Once the target version is changed to a newer version, which includes the mysqld_exporter image, then the image gets set by the openstack-operator similarly to all other images.

Did I miss something? Are there some other concerns I should address?

dprince (approved):

PR #1237, opened 2024-12-20
[OSPRH-12359] Add dnsnames to telemetry certs

vyzigold:
Telemetry-operator tries to use IP addresses to access its dataplane resources, but as a backup it might use hostnames. In order to access Node Exporter and Kepler when they're with hostnames when they're behind TLS, we need to include the hostnames in the certificates.


stuggi (approved):
/lgtm

PR #1238, opened 2024-12-20
Set rhoso 1.0.2 as OPENSTACK_IMG_BASE_RELEASE

stuggi:
OPENSTACK_IMG_BASE_RELEASE specifies the base image to be used in the prow kuttl job to beform a minor update within this job, which was enabled in [1]. For this right now we want to use the downstream 1.0.2 release.

[1] https://github.com/openshift/release/pull/59699


stuggi:
fails to deploy the downstream release

PR #1239, opened 2024-12-20
Add sample for new OpenStack initialization resource

dprince:
This is to work around some issues in CI where
not having this exist is causing some issues


olliewalsh (approved):
/lgtm

PR #1258, opened 2025-01-13
Don't set BaremetalHosts in the nodeset spec

rabi:
Jira: https://issues.redhat.com/browse/OSPRH-12709

### DIFF
--- a/apis/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
+++ b/apis/bases/dataplane.openstack.org_openstackdataplanenodesets.yaml
@@ -51,33 +51,6 @@ spec:
                     - metadata
                     - disabled
                     type: string
-                  baremetalHosts:
-                    additionalProperties:
-                      properties:
-                        bmhLabelSelector:
-                          additionalProperties:
-                            type: string
-                          type: object
-                        ctlPlaneIP:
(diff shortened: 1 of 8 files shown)


rabi:
@abays  Can you chime on what issues you see with this during update? I've disabled the check for field removal here for the crd-schema-checker to pass.

rabi:
recheck

rabi:
> To be fair, I have never actually tried this.

I tested this by updating the operators (uninstalling and installing) and there were no issues. I also realized that we test the same in the kuttl job[1] as well.

https://github.com/openshift/release/blob/master/ci-operator/step-registry/openstack-k8s-operators/kuttl/openstack-k8s-operators-kuttl-commands.sh#L144-L153

abays:
> > To be fair, I have never actually tried this.
> 
> I tested this by updating the operators (uninstalling and installing) and there were no issues. I also realized that we test the same in the kuttl job[1] as well.
> 
> https://github.com/openshift/release/blob/master/ci-operator/step-registry/openstack-k8s-operators/kuttl/openstack-k8s-operators-kuttl-commands.sh#L144-L153

Good point and good to know. :+1:

abays (approved):
/lgtm

PR #1259, opened 2025-01-13
[18.0-fr1] Set Heat timeout

fao89:
ref [OSPRH-10965](https://issues.redhat.com//browse/OSPRH-10965)

(cherry picked from commit b2bd44dc27cbf72c429ef9ea939e6a5d722de674)

PR #1261, opened 2025-01-14
[DNM] test

fao89:


fao89:
checking if we'll get the same error: https://github.com/openstack-k8s-operators/openstack-operator/pull/1259#issuecomment-2589691645

PR #1264, opened 2025-01-15
Don't reset deployment status when generation has not changed

rabi:
This would be the case when we remove certain redundant fields during update.

 jira: https://issues.redhat.com/browse/OSPRH-12709

### DIFF
--- a/controllers/dataplane/openstackdataplanenodeset_controller.go
+++ b/controllers/dataplane/openstackdataplanenodeset_controller.go
@@ -178,7 +178,11 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 	// this reconcile loop.
 	instance.InitConditions()
 	// Set ObservedGeneration since we've reset conditions
-	instance.Status.ObservedGeneration = instance.Generation
+	var generationChanged bool
+	if instance.Generation != instance.Status.ObservedGeneration {
+		instance.Status.ObservedGeneration = instance.Generation
+		generationChanged = true
+	}
 
 	// Always patch the instance status when exiting this function so we can persist any changes.
 	defer func() { // update the Ready condition based on the sub conditions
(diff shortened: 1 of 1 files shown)


rabi:
recheck

controllers/dataplane/openstackdataplanenodeset_controller.go
```diff
@@ -460,6 +464,7 @@ func (r *OpenStackDataPlaneNodeSetReconciler) Reconcile(ctx context.Context, req
 
 func checkDeployment(ctx context.Context, helper *helper.Helper,
 	instance *dataplanev1.OpenStackDataPlaneNodeSet,
+	generationChanged bool,
```
bshephar, line 467: Since we already have the pointer to the instance in this function, it might be better to do the `generationChanged` computation within this function. Just avoids having to copy the variable between memory stacks when we call this function.

controllers/dataplane/openstackdataplanenodeset_controller.go
```diff
@@ -514,7 +519,12 @@ func checkDeployment(ctx context.Context, helper *helper.Helper,
 			} else if deploymentConditions.IsFalse(dataplanev1.NodeSetDeploymentReadyCondition) {
 				isDeploymentRunning = true
 			} else if deploymentConditions.IsTrue(dataplanev1.NodeSetDeploymentReadyCondition) {
-				if deployment.Status.NodeSetHashes[instance.Name] != instance.Status.ConfigHash {
+				// If the nodeset configHash does not match with what's in the deployment and the
```
bshephar, line 522: Like here for example?

```
	if instance.Generation != instance.Status.ObservedGeneration {
 		generationChanged = true
 	}
```

bshephar (commented):
This would be ideal. If we can give ourselves some flexibility to minimize the footprint of our CRD fields without causing undue concern to users and unnecessary deployments. I think that would be valuable.

Just a small comment on where we perform the check for changed generations. Would avoid having to pass the variable around more than necessary.

controllers/dataplane/openstackdataplanenodeset_controller.go
rabi: we set `instance.Status.ObservedGeneration` to `instance.Generation` at the beginning of the reconcile for a reason, so `ObservedGeneration` would be always the same as `instance.Generation` here and it won't work.

controllers/dataplane/openstackdataplanenodeset_controller.go
rabi: Check reply to your other comment.

fao89 (approved):

PR #1270, opened 2025-01-20
Security team testing if absolutely anyone can merge PR

Yadnyawalkya:
For context: ProdSec received reports that absolutely anyone could merge their own PR in the OpenStack-Operator repo, which is clearly not a secure practice. We decided to investigate and this PR is related to that

### DIFF
--- a/main.go
+++ b/main.go
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Adding one more comment line just for testing purpose
 package main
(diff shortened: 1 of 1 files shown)


Yadnyawalkya:
Discovered that not just anyone can set labels, but individuals from the group might be able to. Still testing this, and I’ll close the PR once confirmed.

stuggi:
@Yadnyawalkya is this still required, or could be closed?

PR #1282, opened 2025-01-30
Cleanup legacy OLM Subscription & CSVs

dprince:
This cleans up legacy resources installed by OLM dependencies in old releases by searching for specifically named resources.

This PR uses the unstructured client so that we don't have to import structs (and dependencies).


dprince:
Got this link from Joe today: https://github.com/operator-framework/kubectl-operator/blob/ddb2132976fa3e0ca5da63915c29c77b9240da70/internal/pkg/action/operator_uninstall.go#L186-L242 which may help us improve this further

controllers/operator/openstack_controller.go
```diff
@@ -367,6 +382,65 @@ func (r *OpenStackReconciler) renderAndApply(
 	return nil
 }
 
+var serviceOperatorNames = []string{"barbican", "cinder", "designate", "glance", "heat", "horizon", "infra",
```
bshephar, line 385: This probably doesn't need to be heap allocated. We're only using it in `isServiceOperatorResource`. I know that function looks nice and tidy as it is, but from a memory management perspective, it would be more efficient and clean if this were stack allocated where it's being referenced.

bshephar (commented):
Very nice feature considering how we're changing the installation process. Makes users life much easier.

controllers/operator/openstack_controller.go
leifmadsen: I think it would be worth adding a comment here that this list static and will not require updates in the future due to there not being backported operators. The only other thing I can think of though is that it _might_ update to add watcher perhaps once that is rolled in?

leifmadsen:
Looks ok to me. What are the leftover "operator" resources?

controllers/operator/openstack_controller.go
dprince: @bshephar will move inside the function
@leifmadsen will add a comment according to your suggestion
thanks for the feedback

dprince:
> Looks ok to me. What are the leftover "operator" resources?

oc get operator

controllers/operator/openstack_controller.go
```diff
@@ -372,6 +380,146 @@ func (r *OpenStackReconciler) renderAndApply(
...
+	csvList.SetGroupVersionKind(csvGVR.GroupVersion().WithKind("ClusterServiceVersion"))
+	err := r.Client.List(ctx, csvList, &client.ListOptions{Namespace: instance.Namespace})
+	if err != nil {
+		return err
+	}
+	for _, csv := range csvList.Items {
+		Log.Info("Found CSV", "name", csv.GetName())
+		if isServiceOperatorResource(csv.GetName()) {
+			err = r.Client.Delete(ctx, &csv)
+			if err != nil {
```
abays, line 431: Maybe check for "not found" error here in case the CSV was somehow deleted (in which case it can be ignored and we can just continue)?  Probably won't happen, but just a thought.

controllers/operator/openstack_controller.go
```diff
@@ -372,6 +381,182 @@ func (r *OpenStackReconciler) renderAndApply(
...
+		Log.Info("Found Operator", "name", operator.GetName())
+		if isServiceOperatorResource(operator.GetName()) {
+
+			refs, found, err := uns.NestedSlice(operator.Object, "status", "components", "refs")
+			if err != nil {
+				return err
+			}
+			if found {
+
+				// The horizon-operator.openstack-operators has references to old roles/bindings
```
abays, line 519: Curious as to why Horizon is specifically referenced here.  Is this truly something unique to that operator?

dprince: I took a look and didn't see anything jump out. It is just a Role and RoleBinding that cause the need for this code. All other objects seem to be cleaned up upon removal of the CSV/Subscriptions above.

controllers/operator/openstack_controller.go
dprince: sure, I can add a check for not found

dprince:
recheck

abays:
recheck

stuggi (approved):
/lgtm

abays (approved):
/lgtm

PR #1285, opened 2025-02-03
Add BINDATA_GIT_ADD for 'make bindata'

dprince:
This guards the 'git add' behavior so that we can enabled it only when needed (Renovate jobs, etc)


stuggi (approved):
/lgtm

stuggi (approved):
/lgtm

PR #1288, opened 2025-02-04
Add checkWebhooksActive function

dprince:
to check if webhooks are fully active as it seems the k8s API can take some time. this sends a dry run API request to the infra-operator's Redis API and the OpenStackVersion API


dprince:
I think the ci-framework fix may have resolved what we were trying to solve here:

https://github.com/openstack-k8s-operators/ci-framework/pull/2718/files

PR #1290, opened 2025-02-07
Add checkServiceEndpoints, own deployments

dprince:
Adds a checkServiceEndpoints function which uses a static list of service names to ensure an endpoint exists and has associated addresses. This should help make sure webhooks are fully configured before setting the
final ready condition upon initial deployment.

Also removes old code to "watch" deployments as
we can just own them since the initialization resource is namespaced.

### DIFF
--- a/controllers/operator/openstack_controller.go
+++ b/controllers/operator/openstack_controller.go
@@ -29,8 +29,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/handler"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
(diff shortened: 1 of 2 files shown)


dprince:
recheck

controllers/operator/openstack_controller.go
```diff
@@ -340,6 +353,37 @@ func (r *OpenStackReconciler) countDeployments(ctx context.Context, instance *op
...
 
+// checkServiceEndpoints -
+func (r *OpenStackReconciler) checkServiceEndpoints(ctx context.Context, instance *operatorv1beta1.OpenStack) (ctrl.Result, error) {
+
+	// NOTE: this is a static list for all operators with webhooks enabled
+	endpointNames := []string{"openstack-operator-webhook-service", "infra-operator-webhook-service", "openstack-baremetal-operator-webhook-service"}
+
+	for _, endpointName := range endpointNames {
+		endpoints := &corev1.Endpoints{}
+		err := r.Client.Get(ctx, client.ObjectKey{Name: endpointName, Namespace: instance.Namespace}, endpoints)
```
abays, line 364: Do we maybe need to add kubebuilder annotations to grant us cluster-scoped permission to `get` or `list` `Endpoint`s?

dprince: ack, yes we are missing permissions

dprince:
This should be updated to use EndpointSlice I think...

PR #1295, opened 2025-02-12
only apply CRDs and RBAC once per containerImage

dprince:
Adds a new ContainerImage status field to display the deployed ContainerImage. If this matches the currently configured operator image then applying CRDs and RBAC can be skipped as it was already applied.

Operator deployments are still reconciled each time to ensure they remain in place.


abays (approved):
/lgtm

PR #1298, opened 2025-02-12
Deprecate OpenStackDataPlaneService CertsFrom and EDPMRoleServiceName fields

slagle:
The CertsFrom and EDPMRoleServiceName were both added to solve slightly
different use cases related to ansible content and expected mount paths
in the AEE pod. Later, EDPMServiceType was also added to more broadly
solve all the related issues.

This patch deprecates CertsFrom and EDPMRoleServiceName in favor of
using EDPMServiceType exclusively. The deprecated fields can still be
used with the same functionality until they are officially removed. The
existing usage of EDPMServiceType has also been expanded to cover all
use cases of the previous fields, of which there were 2 cases:

The first was in the deployment controller where certs are issued.
EDPMServiceType is now considered and used to issue certs when it's set
differently than the service name.

The second instance addresses the usage of CertsFrom for creating cert
mounts for the install-certs service. EDPMServiceType is now considered
there as well if it's set differently from the service name.


apis/dataplane/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -53,6 +53,7 @@ type OpenstackDataPlaneServiceCert struct {
 	// not set, OpenStackDataPlaneService.Spec.EDPMServiceType is used. If
 	// OpenStackDataPlaneService.Spec.EDPMServiceType is not set, then
 	// OpenStackDataPlaneService.Name is used.
+	// DEPRECATED: Will be removed in a future release. Use EDPMServiceType instead.
```
slagle, line 56: @dprince @stuggi did we ever decide on a consistent way to deprecate fields and migrate any existing CRs that exist?

dprince: There is a way, but it would probably involve ConversionWebhooks. If you mark them as deprecated now that is fine. Then in the future we can create a new *version* of this CRD and then create a ConversionWebhook to automatically convert CRs to that upon update.

slagle:
recheck

slagle:
recheck

apis/dataplane/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -88,6 +89,7 @@ type OpenStackDataPlaneServiceSpec struct {
 	// CertsFrom - Service name used to obtain TLSCert and CACerts data. If both
 	// CertsFrom and either TLSCert or CACerts is set, then those fields take
 	// precedence.
+	// DEPRECATED: Will be removed in a future release. Use EDPMServiceType instead.
 	// +kubebuilder:validation:Optional
 	CertsFrom string `json:"certsFrom,omitempty" yaml:"certsFrom,omitempty"`
```
rabi, line 94: We could have dropped this optional field (if there is no update impact) as we've done in multiple other cases till now as there would be no issues installing the new CRD. The crd schema checker which does not allow this can be bypassed by disabling `NoFieldsRemoval`.

rabi (approved):
/lgtm

rabi:
> ❌ adoption-standalone-to-crc-ceph-provider FAILURE in 1h 45m 53s

This broken job would be fixed with https://github.com/openstack-k8s-operators/data-plane-adoption/pull/821

PR #1302, opened 2025-02-17
[OSPRH-13961] Remove the creation of the external Prometheus route

jlarriba:
Prometheus will only be exposing an internal route from now on so it makes sense to use the Internal certificate for that.

### DIFF
--- a/pkg/openstack/telemetry.go
+++ b/pkg/openstack/telemetry.go
@@ -205,7 +205,7 @@ func ReconcileTelemetry(ctx context.Context, instance *corev1beta1.OpenStackCont
 			return ctrlResult, nil
 		}
 		// update TLS settings with cert secret
-		instance.Spec.Telemetry.Template.MetricStorage.PrometheusTLS.SecretName = endpointDetails.GetEndptCertSecret(service.EndpointPublic)
+		instance.Spec.Telemetry.Template.MetricStorage.PrometheusTLS.SecretName = endpointDetails.GetEndptCertSecret(service.EndpointInternal)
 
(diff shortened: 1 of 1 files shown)


amoralej:
Wasn't one of the use cases for the prometheus to be available for cloud admins to be integrated in external monitoring or observability tools for dashboarding etc..?

jlarriba:
> Wasn't one of the use cases for the prometheus to be available for cloud admins to be integrated in external monitoring or observability tools for dashboarding etc..?

It is indeed, but as Prometheus is not authenticated in any way, it exposes a lot of information about the cloud to whoever wants to connect to the route, which is deemed not acceptable.

Authenticating Prometheus will be implemented, but it is not a short task so in the meantime, it seems to be better to just close the route to only allow internal access by default. Remember that if a user wants to expose Prometheus via an external route, it can be done manually.

jlarriba:
recheck

amoralej:
> > Wasn't one of the use cases for the prometheus to be available for cloud admins to be integrated in external monitoring or observability tools for dashboarding etc..?
> 
> It is indeed, but as Prometheus is not authenticated in any way, it exposes a lot of information about the cloud to whoever wants to connect to the route, which is deemed not acceptable.
> 
> Authenticating Prometheus will be implemented, but it is not a short task so in the meantime, it seems to be better to just close the route to only allow internal access by default. Remember that if a user wants to expose Prometheus via an external route, it can be done manually.

Thanks for the explanation.

jlarriba:
recheck

jlarriba:
recheck

stuggi (approved):
/lgtm

PR #1307, opened 2025-02-18
Switch DNS to use *Core* spec

stuggi:
These fields should have been dropped when the DNS *Core* spec was implemented.

Images get overwritten during reconcile to what exists in the openstackversion resource so there should be no user facing issues in dropping these fields.

### DIFF
--- a/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
+++ b/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
@@ -2390,8 +2390,6 @@ spec:
                     type: boolean
                   template:
                     properties:
-                      containerImage:
-                        type: string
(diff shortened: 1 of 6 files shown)


dprince:
recheck

stuggi:
recheck

stuggi:
recheck

stuggi:
recheck

bshephar:
recheck

stuggi:
recheck

stuggi:
recheck

stuggi:
overriding zuul since all tests were successful combining the last two runs. also override pre-commit as it is expected to fail for the change

dprince (approved):

PR #1314, opened 2025-02-26
Add omitempty for ansibleUser

rabi:
ansibleUser can be provided either in the NodeSection or NodeTemplate. This shows as empty string in the CR if not provided which is confusing.


stuggi (approved):
/lgtm

PR #1316, opened 2025-02-26
Consume a top-level TopologyRef

fmount:
In the general use case, where there are no hard constraints to the services deployed by the openstack-operator, we might want to reference `TopologyRef` only once, and not override it ~20 times with the same value in the CR.
Because we have default `labelSelectors`, creating a Topology like the following:

```
apiVersion: v1
items:
- apiVersion: topology.openstack.org/v1beta1
  kind: Topology
...
  spec:
    topologySpreadConstraints:
    - maxSkew: 1
      topologyKey: topology.kubernetes.io/hostname
      whenUnsatisfiable: ScheduleAnyway

```

might be a good example where each service applies its own label selector.
This patch adds the `TopoRef` parameter at the openstack-operator top-level, so that updating the spec like the following:

```
apiVersion: core.openstack.org/v1beta1
kind: OpenStackControlPlane
...
spec:
  topologyRef:
     name: foo
  barbican:
  cinder:
  ....
 ````
 applies the `foo` spec to the underlying services unless an override is present.
 By doing this we can define the basic scenario without asking to repeat the same parameter multiple times.

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_webhook.go
+++ b/apis/core/v1beta1/openstackcontrolplane_webhook.go
@@ -55,6 +55,7 @@ import (
 	octaviav1 "github.com/openstack-k8s-operators/octavia-operator/api/v1beta1"
 	swiftv1 "github.com/openstack-k8s-operators/swift-operator/api/v1beta1"
 	telemetryv1 "github.com/openstack-k8s-operators/telemetry-operator/api/v1beta1"
+	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
 )
 
 var ctlplaneWebhookClient client.Client
@@ -117,6 +118,11 @@ func (r *OpenStackControlPlane) ValidateCreate() (admission.Warnings, error) {
 	}
 
 	allWarn, allErrs = r.ValidateCreateServices(basePath)
+
+	if err := r.ValidateTopology(basePath); err != nil {
+		allErrs = append(allErrs, err)
+	}
+
 	if len(allErrs) != 0 {
 		return allWarn, apierrors.NewInvalid(
 			schema.GroupKind{Group: "core.openstack.org", Kind: "OpenStackControlPlane"},
@@ -139,6 +145,10 @@ func (r *OpenStackControlPlane) ValidateUpdate(old runtime.Object) (admission.Wa
 	basePath := field.NewPath("spec")
 	if err := r.ValidateUpdateServices(oldControlPlane.Spec, basePath); err != nil {
 		allErrs = append(allErrs, err...)
+	}
+
+	if err := r.ValidateTopology(basePath); err != nil {
+		allErrs = append(allErrs, err)
 	}
 
 	if len(allErrs) != 0 {
@@ -970,4 +980,15 @@ func validateTLSOverrideSpec(override **route.OverrideSpec, basePath *field.Path
 	}
 
 	return allErrs
+}
+
+func (r *OpenStackControlPlane) ValidateTopology(basePath *field.Path) *field.Error {
+	// When a TopologyRef CR is referenced, fail if a different Namespace is
+	// referenced because is not supported
+	if r.Spec.TopologyRef != nil {
+		if err := topologyv1.ValidateTopologyNamespace(r.Spec.TopologyRef.Namespace, *basePath, r.Namespace); err != nil {
+			return err
+		}
+	}
+	return nil
 }
--- a/controllers/core/openstackcontrolplane_controller.go
+++ b/controllers/core/openstackcontrolplane_controller.go
@@ -32,10 +32,12 @@ import (
 	ironicv1 "github.com/openstack-k8s-operators/ironic-operator/api/v1beta1"
 	keystonev1 "github.com/openstack-k8s-operators/keystone-operator/api/v1beta1"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	common_helper "github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	corev1 "k8s.io/api/core/v1"
 
 	designatev1 "github.com/openstack-k8s-operators/designate-operator/api/v1beta1"
+	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
 	manilav1 "github.com/openstack-k8s-operators/manila-operator/api/v1beta1"
 	mariadbv1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
 	neutronv1 "github.com/openstack-k8s-operators/neutron-operator/api/v1beta1"
@@ -115,6 +117,7 @@ func (r *OpenStackControlPlaneReconciler) GetLogger(ctx context.Context) logr.Lo
 // +kubebuilder:rbac:groups=cert-manager.io,resources=issuers,verbs=get;list;watch;create;update;patch;delete;
 // +kubebuilder:rbac:groups=cert-manager.io,resources=certificates,verbs=get;list;watch;create;update;patch;delete;
 // +kubebuilder:rbac:groups=config.openshift.io,resources=networks,verbs=get;list;watch;
+// +kubebuilder:rbac:groups=topology.openstack.org,resources=topologies,verbs=get;list;watch;update
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
@@ -285,6 +288,21 @@ func (r *OpenStackControlPlaneReconciler) reconcileOVNControllers(ctx context.Co
 }
 
 func (r *OpenStackControlPlaneReconciler) reconcileNormal(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, version *corev1beta1.OpenStackVersion, helper *common_helper.Helper) (ctrl.Result, error) {
+	if instance.Spec.TopologyRef != nil {
+		if err := r.checkTopologyRef(ctx, helper,
+			instance.Spec.TopologyRef, instance.Namespace); err != nil {
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				condition.TopologyReadyCondition,
+				condition.ErrorReason,
+				condition.SeverityWarning,
+				condition.TopologyReadyErrorMessage,
+				err.Error()))
+			return ctrl.Result{}, fmt.Errorf("waiting for Topology requirements: %w", err)
+		}
+		// TopologyRef != nil and exists and we're able to get it
+		instance.Status.Conditions.MarkTrue(condition.TopologyReadyCondition, condition.TopologyReadyMessage)
+	}
+
 	ctrlResult, err := openstack.ReconcileCAs(ctx, instance, helper)
 	if err != nil {
 		return ctrl.Result{}, err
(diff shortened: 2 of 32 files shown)

### CODE REVIEW

File: controllers/core/openstackcontrolplane_controller.go [hunk 1]
```diff
@@ -559,6 +579,17 @@ func (r *OpenStackControlPlaneReconciler) SetupWithManager(
...
+	if err := mgr.GetFieldIndexer().IndexField(ctx, &corev1beta1.OpenStackControlPlane{}, topologyField, func(rawObj client.Object) []string {
+		// Extract the topology name from the spec, if one is provided
+		cr := rawObj.(*corev1beta1.OpenStackControlPlane)
+		if cr.Spec.TopologyRef == nil {
+			return nil
+		}
+		return []string{cr.Spec.TopologyRef.Name}
+	}); err != nil {
+		return err
+	}
```
stuggi, line 592: Do we have to index it in the ctlplane controller? If I have not missed something, we are not using the content of the topology here, so the only situation we care is the change of the name and thats a change we get also without indexing the field.

fmount: Right, we do not need this because it's just about passing parameters to the underlying resources. I removed it, and I removed the Topology watch as well: for any content change we do not need to trigger a reconciliation here, but we trigger the reconciliation on the services that are already watching this topology.

fmount: (Done, resolving the thread unless there's something else to address)

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -971,3 +980,14 @@ func validateTLSOverrideSpec(override **route.OverrideSpec, basePath *field.Path
 
 	return allErrs
 }
+
+func (r *OpenStackControlPlane) ValidateTopology(basePath *field.Path) *field.Error {
+	// When a TopologyRef CR is referenced, fail if a different Namespace is
+	// referenced because is not supported
+	if r.Spec.TopologyRef != nil {
+		if err := topologyv1.ValidateTopologyNamespace(r.Spec.TopologyRef.Namespace, *basePath, r.Namespace); err != nil {
```
gibizer, line 988: I just noticed that we pass basePath here which is the "spec" and then infra-operator appends "topology.namespace" when reports an error. However the actual path in the CR is "spec.topologyRef.namespace".

This is probably a bug affecting all the service operator validation logic as well. So I'm OK to fix this as a bug in a follow up after FR2.

In general the the path handling should be in the same code as the struct field referencing. I.e. if r.Spec.TopologyRef is done here then basePath.Child(...)... should also be here. (or vice verse move both together), to keep the reported path in sync with the used field.

https://github.com/openstack-k8s-operators/infra-operator/blob/ee102640a8459c9d9aad9b0bc442ba0e576ed051/apis/topology/v1beta1/topology_types.go#L114

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -285,6 +288,21 @@ func (r *OpenStackControlPlaneReconciler) reconcileOVNControllers(ctx context.Co
 }
 
 func (r *OpenStackControlPlaneReconciler) reconcileNormal(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, version *corev1beta1.OpenStackVersion, helper *common_helper.Helper) (ctrl.Result, error) {
+	if instance.Spec.TopologyRef != nil {
+		if err := r.checkTopologyRef(ctx, helper,
```
gibizer, line 292: This is a different pattern compared what is now merged in nova-operator. I.e. the nova controller does not check the existence of the provided ref, just pass it down to the service CRs (like NovaAPI) and the service CRs are checking the existence. The opentstackcontrolplance controller also just passing the ref forwards so I don't see why we are checking for existence here.

Also Nova CR has no topology condition just the service CRs has, while here OpenStackControlPlane now got topology condition

pkg/openstack/manila.go
```diff
@@ -146,6 +154,7 @@ func ReconcileManila(ctx context.Context, instance *corev1beta1.OpenStackControl
 			//manila.Spec.DatabaseInstance = instance.Name // name of MariaDB we create here
 			manila.Spec.DatabaseInstance = "openstack" //FIXME: see above
 		}
+
```
gibizer, line 157: seem like whitespace damage.

tests/functional/ctlplane/suite_test.go
gibizer, line 1: I think we need some test coverage here

controllers/core/openstackcontrolplane_controller.go
fmount: This represents an early checking that improves user experience: if the CR doesn't exist in the namespace you have early feedback at the openstack-operator level and you can immediately have the right understanding of  the problem by  looking at the condition status / message (for the same reason we add webhook at this level while we could defer it to the underlying services).
I think if you're trying to inject a topology from the top-level, before moving forward with the reconciliation you should "check your input" and block the process (of course it's a different story if you add the override under the service template, where the process is deferred to the service controller)

apis/core/v1beta1/openstackcontrolplane_webhook.go
fmount: oh! you're right! this is a leftover of the initial iteration (when we moved the name from topology to topologyRef), and it's that kind of detail that I would have never seen before some time. Thank you, I'll fix it in `infra-operator`.
I think I will fix this, like you mentioned, by having something like:

``` 
topologyv1.ValidateTopologyNamespace(r.Spec.TopologyRef.Namespace, *basePath.Child("topolofyRef"), r.Namespace); err != nil {
```

so we can make sure the path is where the struct is defined. Having it in a central location might generate a problem in case the `topologyRef` variable changes (it will never happen but what you said makes a lot of sense).
I will link this comment to the same Jira that I use to follow up in the infra-operator.

gibizer: > ```topologyv1.ValidateTopologyNamespace(r.Spec.TopologyRef.Namespace, *basePath.Child("topolofyRef"), r.Namespace); err != nil {
```
this works for me if the helper does the .namespace field read and does the path.Child("namespace"') as well.

> so we can make sure the path is where the struct is defined
exactly.

---

Also as an alternative

> Having it in a central location might generate a problem in case the topologyRef variable changes 

Yeah if we move the field reading and path append to a central place that means we centralize the way we read into the spec struct which today is symmetric (i.e. it is always called topologyRef) but maybe not in the future. If want to move then we would need an interface passed to the helper that provides the call that reads the topologyRef and also appends the path accordingly. Then this interface can be implemented to all our Spec structs so they can be passed to the helper.

File: controllers/core/openstackcontrolplane_controller.go (hunk 3 above)
Comment (gibizer) [reply controllers/core/openstackcontrolplane_controller.go:292]: > This represents an early checking that improves user experience: if the CR doesn't exist in the namespace you have early feedback at the openstack-operator level and you can immediately have the right understanding of the problem by looking at the condition status / message (for the same reason we add webhook at this level while we could defer it to the underlying services).

How early is that early feedback? Like propagating a topologyRef change from OpenStackControlPlance CR to top level service CRs are as fast as a single reconcile run doing the POST calls on those CRs (on a deployed env, on greenfield it is longer due to dependency ordering across services). I don't feel that this early check is too much gain. 
Still if there is gain then the question is there do we want a similar gain in the service operator top level CR step as well? I.e. why do we wait until NovaAPI reconciles and detects a nonexistent topologyRef if during the Nova CR reconciliation we could catch that?

> I think if you're trying to inject a topology from the top-level, before moving forward with the reconciliation you should "check your input" and block the process (of course it's a different story if you add the override under the service template, where the process is deferred to the service controller)

I agree but then in my head the same argument applies to top level service CRs like Nova. The nova controller should then check the validity of the input before passing it forward to NovaAPI (and the rest)

Comment (fmount) [reply controllers/core/openstackcontrolplane_controller.go:292]: > > This represents an early checking that improves user experience: if the CR doesn't exist in the namespace you have early feedback at the openstack-operator level and you can immediately have the right understanding of the problem by looking at the condition status / message (for the same reason we add webhook at this level while we could defer it to the underlying services).
> 
> How early is that early feedback? Like propagating a topologyRef change from OpenStackControlPlance CR to top level service CRs are as fast as a single reconcile run doing the POST calls on those CRs (on a deployed env, on greenfield it is longer due to dependency ordering across services). I don't feel that this early check is too much gain. Still if there is gain then the question is there do we want a similar gain in the service operator top level CR step as well? I.e. why do we wait until NovaAPI reconciles and detects a nonexistent topologyRef if during the Nova CR reconciliation we could catch that?
> 
I think the advantage here is to not repeat the "checking pattern" for the top level service resources a large amount of times, while we can check it here only once in advance.
We could do the same check for each top-level controller, I agree, but here I think we can save N(=20) reconciliation loops to report the same issue, and not changing the spec with an incorrect parameter.

> > I think if you're trying to inject a topology from the top-level, before moving forward with the reconciliation you should "check your input" and block the process (of course it's a different story if you add the override under the service template, where the process is deferred to the service controller)
> 
> I agree but then in my head the same argument applies to top level service CRs like Nova. The nova controller should then check the validity of the input before passing it forward to NovaAPI (and the rest)

Yes, it might be the right thing to do in case of a standalone Nova deployment, because the same pattern applies. So my question is whether you see any value in saving potentially ~x20 reconciliation loops that try to get the resource and fail. Also, do you think what I'm doing represents an anti-pattern?

File: tests/functional/ctlplane/suite_test.go
Comment (fmount) [reply tests/functional/ctlplane/suite_test.go:1]: FYI I don't know how much remaining time I have but I'll try to write more tests other than the bad case that I already pushed. If I don't have time, I'll do with a follow up PR and backport to FR2.

File: controllers/core/openstackcontrolplane_controller.go (hunk 3 above)
Comment (fmount) [reply controllers/core/openstackcontrolplane_controller.go:292]: FYI this is the main thread I'd like to solve that potentially requires an additional change to "remove" the existing check logic from the openstack-operator. Because we don't have too much time I'd like to converge into an agreement, but I understand that making checks to a level or another might be opinionated. 
I like the idea, for a user who sees the ctlplane CR as the interface, to get a feedback like:

```
$ oc get oscp
NAME               STATUS   MESSAGE
openstack-galera   False    Topology config create error occurred Topology.topology.openstack.org "foo" not found
``` 
and at the same time to not triggering anything behind the scenes (for a good reason I guess), but I understand this apporach might result opinionated, and related to the fact that the meta operator exists to also validate some common  input that flows down to the rest of the services.

fmount:
recheck

fmount:
ok IMO this is ready and considering that CI is green I'm going to push tests update in a follow up.
@gibizer we can discuss more about the logic on a different patch, or we can start a different thread about it, but I don't see any outstanding issue here other than the fact we might not converge on the same idea.
I'll let someone from the core team to decide (@stuggi) whether we want the current check logic before triggering all the reconciliations or if it doesn't make too much sense. I find this an important thing to keep things more efficient.

controllers/core/openstackcontrolplane_controller.go
stuggi: I think you both have valid points:
* I think this check is good to have, since it is user input for a global object we can validate at an really early state where the openstack-operator does not have to know details on a service component internal. maybe we could do similar for the service passwords if the implementation happens in each service operator.
* I think we can also add a check in the service top level controller, like most of the services do for memcache available, or verify the service secret holding the pwd is there. we should probably enhance this in follow ups.

stuggi (approved):
/lgtm

controllers/core/openstackcontrolplane_controller.go
gibizer: First of all. I do not block this PR with this comment. The check is correctly implements its intention. Also it seems that the group is leaning towards this being a useful check. And I'm accepting that.

Second, however I want to show that we might assume that the check gives more value than it actually does:
1. It only validates 1 out of 20+ potential topologyRef human input values in the OpenStackControlPlane CRD. So it might gives a false sense of correctness. I.e. 
  * passing this check does not guarantee that other operators don't fail on another topologyRef input being invalid, We could validate all those input refs here with a bunch of code if we want to, but we don't today. So it seems very arbitrary that we validate this single one. No more no less.
  * failing this check also can be false negative in a very edge case. If the global topologyRef is invalid, but all the service level templates has a valid override then the invalid value is actually unused. And this check actually prevents an otherwise successful deployment. This hints to me that we might trying to do the validation at the wrong place in the dataflow.
2. It only speeds up providing validation feedback slightly. The service operators eventually do the validation for all the 20+ refs. The validations are in the reconcile code in all cases, so all of them anyhow happens async from the user's input. (We are not supposed to do this validation in the webhook. I'm not suggesting it. Even though that would be the real speed up of the feedback.)

Bottom line this check feels very opinionated to me when I'm looking at the big picture.

PR #1317, opened 2025-02-26
Add podman-exporter image default value

paramite:


dprince (approved):

bshephar (approved):

tests/functional/ctlplane/openstackversion_controller_test.go
```diff
@@ -182,6 +183,7 @@ var _ = Describe("OpenStackOperator controller", func() {
 				g.Expect(version.Status.ContainerImages.SwiftProxyImage).ShouldNot(BeNil())
 				g.Expect(version.Status.ContainerImages.EdpmNodeExporterImage).ShouldNot(BeNil())
 				g.Expect(version.Status.ContainerImages.EdpmKeplerImage).ShouldNot(BeNil())
+				g.Expect(version.Status.ContainerImages.EdpmPodmanExporterImage).ShouldNot(BeNil())
```
bshephar, line 186: I don't think you meant to add this twice?

bshephar (changes requested):
Just need to remove that duplicate in the openstackversion_controller_test.go file

tests/functional/ctlplane/openstackversion_controller_test.go
paramite: Yeah, I actually did. All the rest of the EDPM images are there twice too, so I was just following the pattern.

bshephar: Ack, I didn't notice the others were also duplicated. Let's go ahead then and I'll come back and clean them up in a follow up PR. Just in case there is some actual reason for it

paramite:
ci/prow/openstack-operator-build-deploy-kuttl

PR #1319, opened 2025-02-26
container_images: add support for the openstack-network-exporter container

atyronesmith:
The openstack-network-exporter is utilized by the telemetry-operator and collects metrics from ovn, ovs, ovs-dpdk, and DPDK.

Depends-On: https://github.com/openstack-k8s-operators/telemetry-operator/pull/627

PR #1321, opened 2025-02-27
Add bootc image reference

bshephar:
To support image mode RHEL deployments, this change adds the reference to the bootc container image.

Jira: https://issues.redhat.com/browse/OSPRH-14326


bshephar:
recheck FAILURE

bshephar:
Unrelated failures:
```
❯ curl -s "https://logserver.rdoproject.org/21/1321/e320a18f35f52482ca40de94b9c0b121ec5619e2/github-check/cifmw-crc-podified-edpm-baremetal/54b00d4/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/repo-setup-edpm-deployment-ipam-openstack-edpm-ipam-prjfm/logs/repo-setup-edpm-deployment-ipam-openstack-edpm-ipam.log" | tail -n10
fatal: [edpm-compute-0]: FAILED! => {"changed": true, "cmd": "set -euxo pipefail\npushd /var/tmp\ncurl -sL https://github.com/openstack-k8s-operators/repo-setup/archive/refs/heads/main.tar.gz | tar -xz\npushd repo-setup-main\npython3 -m venv ./venv\nPBR_VERSION=0.0.0 ./venv/bin/pip install ./\n# This is required for FIPS enabled until trunk.rdoproject.org\n# is not being served from a centos7 host, tracked by\n# https://issues.redhat.com/browse/RHOSZUUL-1517\nupdate-crypto-policies --set FIPS:NO-ENFORCE-EMS\n./venv/bin/repo-setup current-podified -b antelope\npopd\nsed -i '/AppStream/a exclude=\"libvirt*10.10.0-4*,qemu*9.1.0-9*\"' /etc/yum.repos.d/repo-setup-centos-appstream.repo\nsed -i '/AppStream/a exclude=\"libvirt*10.10.0-4*,qemu*9.1.0-9*\"' /etc/yum.repos.d/centos.repo || true\nrm -rf repo-setup-main\n", "delta": "0:00:20.046624", "end": "2025-02-27 17:57:59.442203", "msg": "non-zero return code", "rc": 2, "start": "2025-02-27 17:57:39.395579", "stderr": "+ pushd /var/tmp\n+ tar -xz\n+ curl -sL https://github.com/openstack-k8s-operators/repo-setup/archive/refs/heads/main.tar.gz\n\ngzip: stdin: unexpected end of file\ntar: Child returned status 1\ntar: Error is not recoverable: exiting now", "stderr_lines": ["+ pushd /var/tmp", "+ tar -xz", "+ curl -sL https://github.com/openstack-k8s-operators/repo-setup/archive/refs/heads/main.tar.gz", "", "gzip: stdin: unexpected end of file", "tar: Child returned status 1", "tar: Error is not recoverable: exiting now"], "stdout": "/var/tmp /home/cloud-admin", "stdout_lines": ["/var/tmp /home/cloud-admin"]}

PLAY RECAP *********************************************************************
edpm-compute-0             : ok=1    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

abays:
Does this by any chance need a change to `pkg/dataplane/util/version.go` as well?  Just checking because I was reviewing something similar this weekend and, in that case, a change was needed in the file [1].

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/1319/files#diff-80f6225dc117813e2cfef98d6356561b01b979f012f2bdc7ff98113f232a2da6

stuggi:
looks good to me, but I'd say lets wait for it to land after the FR2 branching

slagle (commented):
I think we might want to hold on this until we have a targeted release (FR3 perhaps), or at least a plan for how this would be set downstream.

bshephar:
> Does this by any chance need a change to `pkg/dataplane/util/version.go` as well? Just checking because I was reviewing something similar this weekend and, in that case, a change was needed in the file [1].
> 
> [1] https://github.com/openstack-k8s-operators/openstack-operator/pull/1319/files#diff-80f6225dc117813e2cfef98d6356561b01b979f012f2bdc7ff98113f232a2da6

Indeed, thanks. I've updated that reference now.

rabi:
> From a user perspective. They would consume the QCOW2 image to build their nodes during the provisioning process. During the Ansible deployment of the NodeSet, it would update or switch to whatever the user has defined here with this new variable.

OK, I'm not sure if we should keep two container images for bootc in openstackversion. That can also cause confusion and inconsistency. From long term perspective when we move to BMHs/metal3 directly consuming container images, just a new  OsContainerImage during update would suffice. Till that time IMO we should just provide an ansible var for a new image if the user wants os update during minor update (using bootc switch and update).

rabi:
> But that's doable without the ref to this image here. Just needs some thought about how the images are shipped, and what that minor and major upgrade path is going to look like.

Yeah, interim shipping could be adhoc if we support both. Once we move to RHOS19 with RHEL10(bootc) and use bootc deploy interface from ironic as the only supported way, we don't have to maintain both I guess.

apevec:
> > But that's doable without the ref to this image here. Just needs some thought about how the images are shipped, and what that minor and major upgrade path is going to look like.
> 
> Yeah, interim shipping could be adhoc if we support both. Once we move to RHOS19 with RHEL10(bootc) and use bootc deploy interface from ironic as the only supported way, we don't have to maintain both I guess.

yes, please assume this is for RHOSO 19 and RHEL 10 only

PR #1330, opened 2025-03-03
Revert "Add mysqld exporter"

jlarriba:
This reverts commit bb1bbc49fe6fd4bb8162694d43cc1931db3b74d8.

### DIFF
--- a/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
+++ b/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
@@ -17470,8 +17470,6 @@ spec:
                     type: string
                   ceilometerIpmiImage:
                     type: string
-                  ceilometerMysqldExporterImage:
-                    type: string
                   ceilometerNotificationImage:
                     type: string
                   ceilometerProxyImage:
--- a/apis/bases/core.openstack.org_openstackversions.yaml
+++ b/apis/bases/core.openstack.org_openstackversions.yaml
@@ -64,8 +64,6 @@ spec:
                     type: string
                   ceilometerIpmiImage:
                     type: string
-                  ceilometerMysqldExporterImage:
-                    type: string
                   ceilometerNotificationImage:
                     type: string
                   ceilometerSgcoreImage:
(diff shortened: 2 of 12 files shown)


bindata/operator/operator.yaml
```diff
@@ -76,10 +76,7 @@ spec:
         env:
         - name: OPENSTACK_RELEASE_VERSION
           value: '{{ .OpenstackReleaseVersion }}'
-{{ range $envName, $envValue := .OpenStackServiceRelatedImages }}
-        - name: {{ $envName }}
-          value: {{ $envValue }}
-{{ end }}
+        envCustomImage: replace_me
```
stuggi, line 79: is this change ok?

jlarriba: I dont know, if I dont run "pre-commit run -a", the pre-commit CI fails. If I run it, it adds these changes to the code.

jlarriba: It still fails the pre-commit check, so removing this change

stuggi:
> PreCommit does not allow us to remove the image field:
> 
> ```
> ERROR: "NoFieldRemoval": crd/openstackcontrolplanes.core.openstack.org version/v1beta1 field/^.status.containerImages.ceilometerMysqldExporterImage may not be removed
> ```
> 
> Is there a way to overlook this without having to specifically modify the crd-schema-checker.sh script?

we can override the precommit-check `/override ci/prow/precommit-check`

jlarriba:
> > PreCommit does not allow us to remove the image field:
> > ```
> > ERROR: "NoFieldRemoval": crd/openstackcontrolplanes.core.openstack.org version/v1beta1 field/^.status.containerImages.ceilometerMysqldExporterImage may not be removed
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is there a way to overlook this without having to specifically modify the crd-schema-checker.sh script?
> 
> we can override the precommit-check `/override ci/prow/precommit-check`

Cool, only an approver can do that :)

abays:
recheck

stuggi:
recheck

abays:
The Zuul jobs have all passed in the `OR` of all attempts.  We could override at this point, I think.

jlarriba:
> The Zuul jobs have all passed in the `OR` of all attempts. We could override at this point, I think.

+1

PR #1333, opened 2025-03-03
drop use of RELATED_IMAGES some test-operator IMGs

dprince:
Drop use of the RELATED_IMAGES env naming convention for test-operator images that we do not release
downstream and instead just use normal Env variables. This should resolve bundle validation issues with
these related images not existing...

The normal ENV variables can be used upstream for
tests.

Jira: OSPRH-14450


stuggi:
recheck

PR #1336, opened 2025-03-04
Add more topologyRef envTests

fmount:
This patch adds some new `envTests` to verify that we properly manage the lifecycle of the `TopologyRef` parameter. It is possible to add, update and remove such parameter, and the `envTests` are supposed to cover the basic lifecycle.


tests/functional/ctlplane/base_test.go
```diff
@@ -632,3 +675,95 @@ func SimulateControlplaneReady() {
...
+			"name":      topology.Name,
+			"namespace": topology.Namespace,
+		},
+		"spec": spec,
+	}
+	return th.CreateUnstructured(raw)
+}
+
+// GetGlance
+func GetGlance(name types.NamespacedName) *glancev1.Glance {
```
fmount, line 709: In an ideal world these functions might be replaced (like we do for memcached) by the `api/tests` helpers provided by each service operator. This is something we need to reduce the amount of utility functions here.

fmount: https://issues.redhat.com/browse/OSPRH-14532 to track this work

PR #1337, opened 2025-03-05
Duplicate mount check

bshephar:
This duplicate check for inventory mounts is no longer necessary.

### DIFF
--- a/pkg/dataplane/util/ansibleee.go
+++ b/pkg/dataplane/util/ansibleee.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
-	"strings"
 
 	"github.com/openstack-k8s-operators/lib-common/modules/storage"
 	yaml "gopkg.in/yaml.v3"
@@ -37,9 +36,6 @@ type EEJob struct {
 	// ServiceAccountName allows to specify what ServiceAccountName do we want the ansible execution run with. Without specifying,
 	// it will run with default serviceaccount
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
-	// Inventory is the primary inventory that the ansible playbook will use to launch the job.
-	// Further inventories may be provided as ExtraMount in the `/runner/inventory/` path.
-	Inventory string `json:"inventory,omitempty"`
 	// Args are the command plus the playbook executed by the image. If args is passed, Playbook is ignored.
 	Args []string `json:"args,omitempty"`
 	// NetworkAttachments is a list of NetworkAttachment resource names to expose the services to the given network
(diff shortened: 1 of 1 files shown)


apis/dataplane/v1beta1/openstackdataplanenodeset_webhook.go
```diff
@@ -161,6 +164,25 @@ func (r *OpenStackDataPlaneNodeSet) validateNodes() (field.ErrorList, error) {
 
 }
 
+// duplicateMountCheck checks the NodeSet for duplicate mountPaths.
+func (r *OpenStackDataPlaneNodeSet) duplicateMountCheck(extraMount storage.VolMounts) field.ErrorList {
```
rabi, line 168: If you're passing `extraMount` as a parameter, you don't need the receiver, it can just be a function and not a method.

rabi (commented):
If we really want validation for extraMounts conflict, it should probably be in lib-common and implemented across all operators, as most of them use it an as interface to inject files like custom policy.

apis/dataplane/v1beta1/openstackdataplanenodeset_webhook.go
bshephar: True, I originally implemented it as a standalone check, but I realized that we were already iterating over extraMounts, so I could save us a loop by just inserting this check in there with that logic.

bshephar:
> If we really want validation for extraMounts conflict, it should probably be in lib-common and implemented across all operators, as most of them use it an as interface to inject files like custom policy.

I'm not really attached to it, how about I drop that commit from this PR, and we just cleanup the unnecessary inventory duplicate check. Then if we decide we want to implement the duplicate check, I'll look at adding it to lib-common.

bshephar:
recheck

pkg/dataplane/util/ansibleee.go
```diff
@@ -152,32 +151,6 @@ func (a *EEJob) JobForOpenStackAnsibleEE(h *helper.Helper) (*batchv1.Job, error)
 	if len(a.ServiceAccountName) > 0 {
 		job.Spec.Template.Spec.ServiceAccountName = a.ServiceAccountName
 	}
-	// Set primary inventory if specified as string
-	existingInventoryMounts := ""
-	if len(a.Inventory) > 0 {
```
slagle, line 157: nothing sets a.Inventory afaict, which is why we can drop this. It must be a leftover from when we migrated from openstack-ansibleee-operator. If that's correct, then lgtm.

bshephar: Yeah, I don't think there is anything setting it. I should probably drop the param from the struct as well. Updated the PR to remove it.

slagle (commented):
/retest


 stderr:
    fatal: unable to access 'https://github.com/openstack-k8s-operators/openstack-k8s-operators-ci/': Could not resolve host: github.com

slagle (commented):
recheck

```
  msg: 'Failed to download metadata for repo ''repo-setup-centos-appstream'': Yum repo downloading error: Downloading error(s): repodata/a613d657ef7ae97185066e1d88d9a7cfd608d45c714b7447445761854fffd5e5-primary.xml.gz - Cannot download, all mirrors were already tried without success; repodata/6dec6d4712493b6cde13dab4afed1b483482054bdb736add8f3a644ee649a7f6-filelists.xml.gz - Cannot download, all mirrors were already tried without success; repodata/af366105113cf771132e422051a31e5ff2dea19b913a0f937b363fc4b609a94a-modules.yaml.xz - Cannot download, all mirrors were already tried without success'
```

slagle (approved):
lgtm, once CI passes

bshephar:
recheck

bshephar:
recheck

bshephar:
recheck

bshephar:
Last failure was related to the Octavia bridge configuration. Unrelated to the changes proposed here.

PR #1358, opened 2025-03-13
Add edpmServiceType in all bundled dataplane service CRs

rabi:
edpmServiceType is already there for some services like nova, let's add it for others to avoid confusion. Also we've documented that it's required for all custom services.


bshephar (approved):

rabi:
` error: Process interrupted with signal interrupt, cancelling execution... `

bogdando:
makes sense, LGTM

PR #1361, opened 2025-03-14
Default Horizon secret to ControlPlane secret

bshephar:
If the user doesn't provide a secret for the Horizon template, we will default to the global secret for the OpenStackControlPlane.

### DIFF
--- a/apis/core/v1beta1/openstackcontrolplane_webhook.go
+++ b/apis/core/v1beta1/openstackcontrolplane_webhook.go
@@ -920,6 +920,10 @@ func (r *OpenStackControlPlane) DefaultServices() {
 			r.Spec.Horizon.Template = &horizonv1.HorizonSpecCore{}
 		}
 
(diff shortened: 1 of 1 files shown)


apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -920,6 +920,10 @@ func (r *OpenStackControlPlane) DefaultServices() {
 			r.Spec.Horizon.Template = &horizonv1.HorizonSpecCore{}
 		}
 
+		if r.Spec.Horizon.Template.Secret == "" {
```
fmount, line 923: isn't this [1] sufficient to inject this input?

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/openstack/horizon.go#L156

bshephar: Ah, yeah, it is indeed.

Ok, so we only need the webhook validation then.

PR #1351, opened 2025-03-10
Removing galera and rabbitmq from CR after their oscp removal

auniyal61:
Closes: #[OSPRH-8061](https://issues.redhat.com//browse/OSPRH-8061)
Closes: [OSPRH-8059](https://issues.redhat.com//browse/OSPRH-8059)

PR #1385, opened 2025-04-01
Vncproxy cert cleanup

auniyal61:
[OSPRH-10549](https://issues.redhat.com//browse/OSPRH-10549)


pkg/openstack/galera.go
```diff
@@ -28,14 +31,64 @@ const (
...
+				}
+
+				certName := fmt.Sprintf("galera-%s-svc", galeraObj.Name)
+				certSecretName := fmt.Sprintf("cert-galera-%s-svc", galeraObj.Name)
+
+				err = DeleteCertificate(ctx, helper, instance.Namespace, certName)
+				if err != nil {
+					delErrs = append(delErrs, fmt.Errorf("galera cert deletion for '%s' failed, because: %w", certName, err))
+				}
+				err = DeleteSecret(ctx, helper, instance.Namespace, certSecretName)
```
stuggi, line 68: same as https://github.com/openstack-k8s-operators/openstack-operator/pull/1351#discussion_r2022445828

pkg/openstack/nova.go
```diff
@@ -40,6 +44,54 @@ import (
...
+			if _, exists := (instance.Spec.Nova.Template.CellTemplates)[cell]; !exists {
+				if object.CheckOwnerRefExist(instance.GetUID(), cert.OwnerReferences) {
+					secretName := "cert-" + cert.Name
+
+					log.Info("Deleting novncproxy cert and secret for %s", "", cell)
+					err = DeleteCertificate(ctx, helper, instance.Namespace, cert.Name)
+					if err != nil {
+						delErrs = append(delErrs, fmt.Errorf("novncproxy cert deletion for '%s' failed, because: %w", cert.Name, err))
+					}
+					err = DeleteSecret(ctx, helper, instance.Namespace, secretName)
```
gibizer, line 77: ditto drop the Secret deletion and keep the Cert deletion as we discussed in the Galera / Rabbit PR

pkg/openstack/nova.go
```diff
@@ -40,6 +44,54 @@ import (
...
+	}
+	err := helper.GetClient().List(ctx, allCerts, listOpts...)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not get certs %w", err)
+	}
+
+	var delErrs []error
+	for _, cert := range allCerts.Items {
+		if strings.HasPrefix(cert.Name, "nova-novncproxy-cell") {
+			cell := strings.Split(cert.Name, "-")[2]
```
gibizer, line 67: this relies on a naming convention of the Cert. If the same naming convention is used at Cert creation the this will work. I suggest to look at the Cert creation logic to see if there is a way to pull out the naming convention into a single helper and user it in both places, so that if the naming changes then it changes for both place (Cert creation and here at Cert deletion)

stuggi: https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/openstack/nova.go#L290

pkg/openstack/common.go
```diff
@@ -795,3 +795,23 @@ func hasCertInOverrideSpec(overrideSpec route.OverrideSpec) bool {
 		overrideSpec.Spec.TLS.Certificate != "" &&
 		overrideSpec.Spec.TLS.Key != ""
 }
+
+func DeleteCertificate(
```
auniyal61, line 799: this will be removed once PR 1351  get merged

pkg/openstack/nova.go
auniyal61: done

pkg/openstack/galera.go
auniyal61: done

pkg/openstack/nova.go
auniyal61: `nova.Name + "-novncproxy-" + cellName + "-vencrypt",`

with novncproxy we have multiple pattern certs to delete 

```
cert-nova-novncproxy-cell1-public-route
cert-nova-novncproxy-cell1-public-svc   
cert-nova-novncproxy-cell1-vencrypt  
```
I don't think `nova.Name="nova"` should ever change to something else.

so we should be able to use "cert-"+"nova-novncproxy" + "cell"  as contant, 
and then concate str(cellNum), which is same as above

tests/functional/ctlplane/openstackoperator_controller_test.go
```diff
@@ -3102,6 +3102,32 @@ var _ = Describe("OpenStackOperator controller galera and rabbitmq", func() {
...
 
+		It("cell1 certs should be deleted from CR", func() {
+			// OSCtlplane := GetOpenStackControlPlane(names.OpenStackControlplaneName)
+			// Expect(OSCtlplane.Spec.Rabbitmq.Enabled).Should(BeTrue())
+			// th.Logger.Info("", ">>> 1 OSCP-", OSCtlplane)
+
+			// // enable Nova
+			// Eventually(func(g Gomega) {
+			// 	OSCtlplane := GetOpenStackControlPlane(names.OpenStackControlplaneName)
+			// 	OSCtlplane.Spec.Nova.Enabled = true
```
auniyal61, line 3113: To generate vncproxy certs we need to enable nova, which desabled by-default, 
while enabling it getting webhook validatation, eventually timeout error. https://paste.opendev.org/show/bScYDWf4XpgDGcLBZZHL/

is there any other way to create these certs in envtest.

stuggi: The webhook is enabled https://github.com/openstack-k8s-operators/openstack-operator/blob/main/tests/functional/ctlplane/suite_test.go#L329 . I fail to understand right now whats is different here compared to https://github.com/openstack-k8s-operators/openstack-operator/blob/main/tests/functional/ctlplane/openstackoperator_controller_test.go#L3070 .

just to note, in envtests the cert secrets won't get created since cert-manager is not running. if the cert secret is required, you'd have to create it like https://github.com/openstack-k8s-operators/openstack-operator/blob/main/tests/functional/ctlplane/openstackoperator_controller_test.go#L1270

stuggi: ok I get it now. The error message is not showing what is actual missing. you have to at least provide the empty nova.template to get it defaulted. 

I did a quick test using the bellow. you also have to make sure required cert secrets are there for other services which get reconciled before are there.
```
$ git diff
diff --git a/tests/functional/ctlplane/openstackoperator_controller_test.go b/tests/functional/ctlplane/openstackoperator_controller_test.go
index 845171a2..2651c799 100644
--- a/tests/functional/ctlplane/openstackoperator_controller_test.go
+++ b/tests/functional/ctlplane/openstackoperator_controller_test.go
@@ -45,6 +45,7 @@ import (
        "github.com/openstack-k8s-operators/lib-common/modules/common/service"
        "github.com/openstack-k8s-operators/lib-common/modules/common/tls"
        manilav1 "github.com/openstack-k8s-operators/manila-operator/api/v1beta1"
+       novav1 "github.com/openstack-k8s-operators/nova-operator/api/v1beta1"
        clientv1 "github.com/openstack-k8s-operators/openstack-operator/apis/client/v1beta1"
        corev1 "github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1"
        ovnv1 "github.com/openstack-k8s-operators/ovn-operator/api/v1beta1"
@@ -2931,6 +2932,14 @@ var _ = Describe("OpenStackOperator controller galera and rabbitmq", func() {
                        th.CreateCertSecret(names.RabbitMQCertName)
                        th.CreateCertSecret(names.RabbitMQCell1CertName)
 
+                       // create cert secrets for ovn instance
+                       DeferCleanup(k8sClient.Delete, ctx, th.CreateCertSecret(names.OVNNorthdCertName))
+                       DeferCleanup(k8sClient.Delete, ctx, th.CreateCertSecret(names.OVNControllerCertName))
+                       DeferCleanup(k8sClient.Delete, ctx, th.CreateCertSecret(names.NeutronOVNCertName))
+
+                       // create cert secrets for memcached instance
+                       DeferCleanup(k8sClient.Delete, ctx, th.CreateCertSecret(names.MemcachedCertName))
+
                        extGalera := CreateGaleraConfig(namespace, GetDefaultGaleraSpec())
                        extGaleraName.Name = extGalera.GetName()
                        extGaleraName.Namespace = extGalera.GetNamespace()
@@ -3103,18 +3112,20 @@ var _ = Describe("OpenStackOperator controller galera and rabbitmq", func() {
                })
 
                It("cell1 certs should be deleted from CR", func() {
-                       // OSCtlplane := GetOpenStackControlPlane(names.OpenStackControlplaneName)
-                       // Expect(OSCtlplane.Spec.Rabbitmq.Enabled).Should(BeTrue())
-                       // th.Logger.Info("", ">>> 1 OSCP-", OSCtlplane)
+                       OSCtlplane := GetOpenStackControlPlane(names.OpenStackControlplaneName)
+                       Expect(OSCtlplane.Spec.Rabbitmq.Enabled).Should(BeTrue())
+                       th.Logger.Info("", ">>> 1 OSCP-", OSCtlplane)
+
+                       // enable Nova
+                       Eventually(func(g Gomega) {
+                               OSCtlplane := GetOpenStackControlPlane(names.OpenStackControlplaneName)
+                               OSCtlplane.Spec.Nova.Enabled = true
+                               OSCtlplane.Spec.Nova.Template = &novav1.NovaSpecCore{}
 
-                       // // enable Nova
-                       // Eventually(func(g Gomega) {
-                       //      OSCtlplane := GetOpenStackControlPlane(names.OpenStackControlplaneName)
-                       //      OSCtlplane.Spec.Nova.Enabled = true
-                       //      g.Expect(k8sClient.Update(ctx, OSCtlplane)).Should(Succeed())
-                       // }, timeout, interval).Should(Succeed())
+                               g.Expect(k8sClient.Update(ctx, OSCtlplane)).Should(Succeed())
+                       }, timeout, interval).Should(Succeed())
 
-                       OSCtlplane := GetOpenStackControlPlane(names.OpenStackControlplaneName)
+                       OSCtlplane = GetOpenStackControlPlane(names.OpenStackControlplaneName)
                        th.Logger.Info("", ">>> 2 OSCP-", OSCtlplane)
 
                        Eventually(func(g Gomega) {
```

but creating nova still fails. since not all required services are enabled. which blocks nova to be enabled. It would be a requirement to also enable  Glance, Neutron, Placement as those are checked to be required for nova https://github.com/openstack-k8s-operators/openstack-operator/blob/main/apis/core/v1beta1/openstackcontrolplane_webhook.go#L202-L205

```
  [FAILED] Timed out after 10.000s.                                                                                                            
  The function passed to Eventually failed at /home/mschuppe/src/github.com/openstack-k8s-operators/openstack-operator/tests/functional/ctlplan
e/openstackoperator_controller_test.go:3125 with:                                                                                              
  Expected success, but got an error:                                                                                                          
      <*errors.StatusError | 0xc001b31400>:                                                                                                    
      admission webhook "vopenstackcontrolplane.kb.io" denied the request: OpenStackControlPlane.core.openstack.org "64a84e56-e819-4a74-baa9-9"
 is invalid: spec.nova.enabled: Invalid value: true: Nova requires these services to be enabled: Galera, Memcached, RabbitMQ, Keystone, Glance, Neutron, Placement. 
```

tests/functional/ctlplane/base_test.go
```diff
@@ -44,47 +44,50 @@ import (
...
+	DBCertName                         types.NamespacedName
+	DBCell1Name                        types.NamespacedName
+	DBCell1CertName                    types.NamespacedName
+	RabbitMQName                       types.NamespacedName
+	RabbitMQCertName                   types.NamespacedName
+	RabbitMQCell1Name                  types.NamespacedName
+	RabbitMQCell1CertName              types.NamespacedName
+	NoVNCProxyCell1CertPublicRouteName types.NamespacedName
+	NoVNCProxyCell1CertPublicSvcName   types.NamespacedName
+	NoVNCProxyCell1CertVencryptName    types.NamespacedName
```
auniyal61, line 70: these are the added ones

tests/functional/ctlplane/openstackoperator_controller_test.go
auniyal61: thakns @stuggi after enabling reset of the services, nova is enabled.

pkg/openstack/nova.go
```diff
@@ -394,6 +467,10 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
 			corev1beta1.OpenStackControlPlaneNovaReadyRunningMessage))
 	}
 
+	_, err = deleteUndefinedNoVNCProxyCertsByCellName(ctx, instance, helper)
+	if err != nil {
+		return ctrl.Result{}, err
```
gibizer, line 472: @auniyal61 please remind me how we did error handling in the Rabbit / Galera case? Did we set the some condition when the cleanup failed? If, so then I guess we want to follow that pattern here too.

pkg/openstack/nova.go
```diff
@@ -40,6 +45,74 @@ import (
...
+	allRoutes := &routev1.RouteList{}
+	listOpts = []client.ListOption{
+		client.InNamespace(instance.GetNamespace()),
+	}
+	err = helper.GetClient().List(ctx, allRoutes, listOpts...)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not get routes %w", err)
+	}
+
+	novncproxyPrefix := "nova-novncproxy-cell"
```
gibizer, line 74: This and the logic of pulling the cell name out of the Cert / Route name should be encapsulated made common with the similar logic that I assume exists for creating the Cert and Route with the matching name.

gibizer (commented):
It seems the envtest env needs work to be able to cover the change.

pkg/openstack/nova.go
auniyal61: there we were deleting certs first and on success only we were cleaning up real galera/rabbit
here were are only cleaning ip certs

gibizer: I went and looked. We set status conditions when the cleanup failed. So we should do so here too..


https://github.com/openstack-k8s-operators/openstack-operator/pull/1351/files#diff-3f5af30ce4253057b810133e0f7b0892abc85b9cb49951f5bb243738790b5dc8R194-R203

gibizer:
@auniyal61 
```diff
diff --git a/pkg/openstack/nova.go b/pkg/openstack/nova.go
index 780a4a6..fc4caef 100644
--- a/pkg/openstack/nova.go
+++ b/pkg/openstack/nova.go
@@ -282,6 +282,7 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
        }
 
        // cell Metadata and NoVNCProxy
+       Log.Info("XXX ReconcileNova iterating cells")
        for cellName, cellTemplate := range instance.Spec.Nova.Template.CellTemplates {
                // create certificate for Metadata agend if internal TLS and Metadata per cell is enabled
                if instance.Spec.TLS.PodLevel.Enabled &&
@@ -305,6 +306,8 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
 
                // NoVNCProxy check for/creating route if service is enabled
                if noVNCProxyEnabled(cellTemplate.NoVNCProxyServiceTemplate) {
+                       Log.Info("XXX ReconcileNova VNC enabled", "cellName", cellName)
+
                        if cellTemplate.NoVNCProxyServiceTemplate.Override.Service == nil {
                                cellTemplate.NoVNCProxyServiceTemplate.Override.Service = &service.RoutedOverrideSpec{}
                        }
@@ -320,6 +323,7 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
                        }
 
                        // make sure to get to EndpointConfig when all service got created
+                       Log.Info("XXX", "vnc services", svcs.Items)
                        if len(svcs.Items) == 1 {
                                endpointDetails, ctrlResult, err := EnsureEndpointConfig(
                                        ctx,
```

I added the above logs and run your test. This shows that the reconcile exists because openstack-operator waits for the NoVNCProxy k8s Service to be created (by nova-operator in a real env) before it generates the Certs. But in the env test there i no nova-operator so the k8s Service should be created manually from the test code. You can see an example how to create the k8s Service from the test in the test case "should have created keystone with route using custom cert secret". We need something similar than what is in the BeforeEach of that test case, calling `th.CreateService(` but not for the Keystone service but for the NoVncProxy service.

pkg/openstack/nova.go
auniyal61: ack done

pkg/openstack/nova.go
auniyal61: moved to common.go, but this is specific to when cell comes at index 2, which most of the cases, but should not be called for certs with string `rabbitmq-cell1-svc`
added note too for same, did not moved prefix string, as its specific to caller

PR #1397, opened 2025-04-06
Used named returns for checkDeployments func

bshephar:
The checkDeployments function uses typed return values and it can be mentally taxing to keep track of which values should be returned in which order. This change switches to named return values which makes it easier to keep track of which values are being returned.

Jira: https://issues.redhat.com/browse/OSPRH-15478


bshephar:
recheck

danpawlik:
Checking Quay RDO connection.

PR #1398, opened 2025-04-07
Temporarily skip ListMustHaveSSATags

fmount:
We recently introduced extraMounts in keystone, where we skip `ListMustHaveSSATags`. The reason behind that is related to the fact that `extraMounts`, as per the `crd-schema-checker`, require this tag, which is not defined in the most inner `VolumeSource` type that we import in `lib-common`. To avoid addition work and complication, and to not block any other patch in `opentstack- operator`, this patch skips the `ListMustHaveSSATags` validator the same way we do in `keystone-operator` [1].

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/main/hack/crd-schema-checker.sh#L5C35-L5C56

### DIFF
--- a/hack/crd-schema-checker.sh
+++ b/hack/crd-schema-checker.sh
@@ -3,7 +3,7 @@ set -euxo pipefail
 
 CHECKER=$INSTALL_DIR/crd-schema-checker
 
-DISABLED_VALIDATORS=NoMaps,NoBools # TODO: https://issues.redhat.com/browse/OSPRH-12254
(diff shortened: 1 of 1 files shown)


fmount:
> wondering if we should keep it and instead override it when needed as it is only checked when we add new CRD changes. otherwise we'd not add annotations where it is possible.

I see, you mean that basically once we merge the new field that subsequent patches can land because it not part of the diff anymore. If I'm reading it correctly, I agree to keep the check and override only what we need.
@stuggi can I then close this PR?

fmount:
ok as per the previous comment I'm closing this PR.

PR #1403, opened 2025-04-15
Add warning about NodeSet updates with OSDP failures

bshephar:
When a NodeSet is updated while it is associated with a failing OSDPD, we will update the NodeSet but NOT update the Ansible inventory. This change adds a warning for users so that they are aware of the fact that the OSDPD needs to be deleted before the Ansible inventory will be updated.

Jira: OSPRH-15668


bshephar:
recheck

PR #1413, opened 2025-04-24
Convert floats to ints after unmarshalling

bshephar:
When unmarshalling values from json input, all numbers are set to floats. This has unexpected side effects when passing them to Ansible where Ansible is expecting integers. This change adds a helper function to convert all float values to integers.

Jira: https://issues.redhat.com/browse/OSPRH-15097

### DIFF
--- a/pkg/dataplane/inventory.go
+++ b/pkg/dataplane/inventory.go
@@ -39,7 +39,6 @@ import (
 // getAnsibleVarsFrom gets ansible vars from ConfigMap/Secret
 func getAnsibleVarsFrom(ctx context.Context, helper *helper.Helper, namespace string, ansible *dataplanev1.AnsibleOpts) (map[string]interface{}, error) {
 	result := make(map[string]interface{})
-	var value interface{}
 
 	for _, dataSource := range ansible.AnsibleVarsFrom {
 		configMap, secret, err := util.GetDataSourceCmSecret(ctx, helper, namespace, dataSource)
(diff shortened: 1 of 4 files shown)


bshephar:
Result:
```
[m3@osp-df-3 openstack-operator]$ oc get secret dataplanenodeset-openstack-edpm-bootc -o jsonpath='{.data.inventory}' | base64 -d | yq '."openstack-edpm-bootc".vars.rhc_organization'
12345678
```

pkg/dataplane/inventory.go
```diff
@@ -39,51 +39,87 @@ import (
...
+	for k, v := range data {
+		if len(prefix) > 0 {
+			k = prefix + k
+		}
+
+		err := json.Unmarshal([]byte(v), &value)
+		if err != nil {
+			value = v
+		}
+
```
slagle, line 78: here you could just call convertToInt on value, and then see my comment in that func

pkg/dataplane/inventory.go
```diff
@@ -39,51 +39,87 @@ import (
...
+				result[key] = convertToInt(val)
+			}
+		default:
+			result[k] = convertToInt(v)
+		}
+	}
+	return nil
+}
+
+func convertToInt(v any) any {
```
slagle, line 116: then here you would range over v and check it's type. If it's a map, recursively call convertToInt again on the map's values.

pkg/dataplane/inventory.go
```diff
@@ -39,51 +39,87 @@ import (
...
 
+// processConfigMapData processes the key-value pairs from ConfigMap and adds them to the result map.
+func processConfigMapData(data map[string]string, prefix string, result map[string]any) error {
+	var value any
+	for k, v := range data {
+		if len(prefix) > 0 {
+			k = prefix + k
+		}
+
+		err := json.Unmarshal([]byte(v), &value)
```
slagle, line 74: could use  https://pkg.go.dev/encoding/json#NewDecoder and then https://pkg.go.dev/encoding/json#Decoder.UseNumber to keep all numbers. and then could be smart about which ones to convert to int or float.

bshephar: Ok, I used the `decoder.UseNumber()`. This seems to work now:
```
[m3@osp-df-3 openstack-operator]$ oc get secret dataplanenodeset-openstack-edpm-bootc -o jsonpath='{.data.inventory}' | base64 -d | grep rhc_org
        rhc_organization: "1234567890"
```

bshephar: Prior to the change, it was in the inventory as:
```
        rhc_organization: 1.23456789e+09
```

pkg/dataplane/inventory.go
bshephar: I don't think we need the function anymore, after implementing the `UseNumber()` method, it's a bit cleaner without the need for the helper function.

slagle (approved):

PR #1424, opened 2025-04-30
Implementation of common notifications bus interface

bogdando:
Let other services to consume the common notifications bus config
from the top scope spec.

Make it working for Nova yet, with more services to come by.

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/1402

Jira: [OSPRH-15390](https://issues.redhat.com//browse/OSPRH-15390)


go.mod
```diff
@@ -127,3 +127,5 @@ replace github.com/openshift/api => github.com/openshift/api v0.0.0-202408300231
 
 // custom RabbitmqClusterSpecCore for OpenStackControlplane (v2.9.0_patches_tag)
 replace github.com/rabbitmq/cluster-operator/v2 => github.com/openstack-k8s-operators/rabbitmq-cluster-operator/v2 v2.6.1-0.20241017142550-a3524acedd49 //allow-merging
+
+replace github.com/openstack-k8s-operators/nova-operator/api => ../nova-operator/api
```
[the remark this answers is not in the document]
gibizer: (if you point this override to the work holding the commit you need here then CI will be able to pick it up too and test the commits together.

pkg/openstack/nova.go
```diff
@@ -80,6 +80,25 @@ func ReconcileNova(ctx context.Context, instance *corev1beta1.OpenStackControlPl
 		instance.Spec.Nova.Template.TopologyRef = instance.Spec.TopologyRef
 	}
 
+	// When there's no NotificationsBusInstance referenced in the Service Template,
+	// inject the top-level one. Allow overriding notifications bus instance locally
+	// in the services templates, unless it is "" top scope, which always wins.
```
gibizer, line 85: This feels unnecessarily complicated. I suggest to do it simply:
* if the value in the template is not `nil` the use that value (either `""` or `"<name>"`)
* if the value in the template is `nil` then copy the value from the top level field to the template field (again that can be either `""` or `"<name>"`.

bogdando: this implements the designed CRD, let's move discussions there

go.mod
```diff
@@ -127,3 +127,5 @@ replace github.com/openshift/api => github.com/openshift/api v0.0.0-202408300231
 
 // custom RabbitmqClusterSpecCore for OpenStackControlplane (v2.9.0_patches_tag)
 replace github.com/rabbitmq/cluster-operator/v2 => github.com/openstack-k8s-operators/rabbitmq-cluster-operator/v2 v2.6.1-0.20241017142550-a3524acedd49 //allow-merging
+
+replace github.com/openstack-k8s-operators/nova-operator/api => github.com/bogdando/nova-operator/api v0.0.0-20250512104915-dcbaf22bcfbe
```
bogdando, line 131: TODO: remove it after the dependency Pr merged

bogdando:
I see fr3 branch was created recently, so unblocking

fmount:
@bogdando #1591 should cover this patch as well at this point.

fmount:
@bogdando this patch can be closed as it was implemented by #1591.

PR #1426, opened 2025-05-02
Add CRD sync check

dprince:
This github action runs every 2 hours to check if CRDs are in sync by running 'force-bump, then bindata' and finally
checking if there are CRD changes in the bindata directory

Also adds a badge to display the workflow status on the README

### DIFF
--- a/.github/workflows/crd-sync-check.yaml
+++ b/.github/workflows/crd-sync-check.yaml
@@ -1,0 +1,34 @@
+# a simple boolean to check of a CRD sync check is needed
+name: CRD sync check
+
+on:
+  schedule:
+    - cron: '0 */2 * * *' # every 2 hours
+  workflow_dispatch:
+
+jobs:
+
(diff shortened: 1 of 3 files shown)


.github/workflows/crd-sync-check.yaml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  schedule:
+    - cron: '0 */2 * * *' # every 2 hours
+  workflow_dispatch:
+
+jobs:
+
+  crd-sync-check:
+    if: github.repository_owner == 'openstack-k8s-operators'
+    name: Check if
```
bshephar, line 13: This is where the trailing whitespace checker is failing. Just need to remove the space after the if

dprince: should be fixed now. thanks

stuggi (approved):
/lgtm

PR #1428, opened 2025-05-05
Leader election lease tunables

hjensas:
Set leader election LeaseDuration, RenewDealine and RetryPeriod via env variables.

The values used are the ones mentioned as recommended in:
  https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#high-availability

Jira: [OSPRH-16335](https://issues.redhat.com//browse/OSPRH-16335)

### DIFF
--- a/apis/go.mod
+++ b/apis/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/openstack-k8s-operators/infra-operator/apis v0.6.1-0.20250424140239-2d89c1d9f3ec
 	github.com/openstack-k8s-operators/ironic-operator/api v0.6.1-0.20250423054051-4b1b486312a5
 	github.com/openstack-k8s-operators/keystone-operator/api v0.6.1-0.20250420193521-f2b86a82d7e8
-	github.com/openstack-k8s-operators/lib-common/modules/common v0.6.1-0.20250423055245-3cb2ae8df6f0
+	github.com/openstack-k8s-operators/lib-common/modules/common v0.6.1-0.20250505061650-7cb2f323fb35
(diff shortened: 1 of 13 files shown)


hjensas:
> The way you have it now everything is hard coded. I think the way it should work would be that the values are picked up from the CSV and we propagate them accordingly into how the controller-managers get launched via the OpenStack initialization resource.
> 
> So it would be CSV -> controller-operator --> all the controller-managers (including openstack-operator-controller-manager). Sound reasonable?

mm, reasonable yes - but I am lacking the full understanding of the CSV -> controller-operator work. Sorry.
Are there docs, or can you point me at an example how a similar thing is already done?

 I did something, but I feel it is not quite what you are suggesting still ... if it feels, easier to just take over than explaining ...

dprince:
> mm, reasonable yes - but I am lacking the full understanding of the CSV -> controller-operator work. Sorry. Are there docs, or can you point me at an example how a similar thing is already done?

It would work somewhat similarly to how the RELATED_IMAGES work.

> 
> I did something, but I feel it is not quite what you are suggesting still ... if it feels easier to just take over than explaining feel free to just do it. :)

Sure. I can help implement this part.

hjensas:
I don't gr0k how to do this, will let @dprince do this part.

dprince:
@dprince what you did in the second push looks pretty close to what we need i think. Trying it out today

stuggi:
> @dprince what you did in the second push looks pretty close to what we need i think. Trying it out today

@dprince I have a PR prepared for this. I am just waiting on the current quay.io issue to be resolved to properly test it

PR #1430, opened 2025-05-07
AppCred support

Deydra71:
[OSPRH-14738](https://issues.redhat.com/browse/OSPRH-14738)

This PR add ApplicationCredential support enabling both global defaults and service-specific overrides in OpenStackControlPlane.

CRD updates:
* New top-level `spec.applicationCredential ` section with `enabled`, `expirationDays` and `gracePeriodDays`
* Added applicationCredential field to each service section that has a corresponding Keystone user
* Defaults guarantee visible `enabled:false` in every supported service, while `expirationDays` and `gracePeriodDays` are hidden unless specified directly (in that case global values are used).

Controller logic:
* To enable AppCred both global and service specific must be `enable: true` 
* For each enabled service, controller creates AC CR
* Deletes service AC CRs when either global or service `enabled` is turned off

 Example:
```
 spec:
  applicationCredential:
    enabled: true
    expirationDays: 14
    gracePeriodDays: 7
  barbican:
    applicationCredential:
      enabled: true
      expirationDays: 2
      gracePeriodDays: 1
  cinder:
    applicationCredential:
      enabled: true
```

In the example barbican is using days overrides, while cinder is using default values.

Depends-On: https://github.com/openstack-k8s-operators/keystone-operator/pull/567


danpawlik:
recheck

internal/openstack/applicationcredential.go
```diff
@@ -0,0 +1,430 @@
...
+		Enabled   bool
+		ACSection *corev1beta1.ServiceAppCredSection
+	}
+
+	// Collect each service's enabled flag and AC section
+	svcs := []svcAC{
+		{"glance", instance.Spec.Glance.Enabled, instance.Spec.Glance.ApplicationCredential},
+		{"nova", instance.Spec.Nova.Enabled, instance.Spec.Nova.ApplicationCredential},
+		{"swift", instance.Spec.Swift.Enabled, instance.Spec.Swift.ApplicationCredential},
+		{"ceilometer", instance.Spec.Telemetry.Enabled, instance.Spec.Telemetry.ApplicationCredentialCeilometer},
```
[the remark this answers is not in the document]
Deydra71: Yes, good catch. We don't want to rely on solely "disabled implies empty template fields"

api/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -833,6 +946,97 @@ type WatcherSection struct {
...
+	// +kubebuilder:validation:Optional
+	// +listType=atomic
+	AccessRules []ACRule `json:"accessRules,omitempty"`
+}
+
+// +kubebuilder:validation:XValidation:rule="!(has(self.expirationDays) && has(self.gracePeriodDays)) || self.gracePeriodDays < self.expirationDays",message="gracePeriodDays must be smaller than expirationDays"
+// ServiceAppCredSection allows service-specific overrides of the global AC configuration
+type ServiceAppCredSection struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
```
stuggi: see my comment on the global parameter, should this one default to true?

api/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -833,6 +946,97 @@ type WatcherSection struct {
...
+	// AccessRules lets the service override either the global rules
+	// +kubebuilder:validation:Optional
+	// +listType=atomic
+	AccessRules []ACRule `json:"accessRules,omitempty"`
+}
+
+// ACRule describes a single access rule for an ApplicationCredential
+// +k8s:openapi-gen=true
+type ACRule struct {
+	// Service is the name of the service to target (e.g. "identity").
```
stuggi: from where to get the service name? can we link a reference or will it be in product doc or how to use ac rules?

api/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -833,6 +946,97 @@ type WatcherSection struct {
...
+
+	// +kubebuilder:validation:Optional
+	// Roles to assign to the ApplicationCredential
+	Roles []string `json:"roles,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// Whether the AC should be unrestricted
+	Unrestricted *bool `json:"unrestricted,omitempty"`
+
+	// AccessRules lets the service override either the global rules
```
stuggi: is there an `or` missing?

api/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -810,6 +810,10 @@ func (r *OpenStackControlPlane) DefaultServices() {
 			r.Spec.Cinder.Template = &cinderv1.CinderSpecCore{}
 		}
 		r.Spec.Cinder.Template.Default()
+		// Default Auth fields for Application Credentials
+		if r.Spec.Cinder.Template.CinderAPI.Auth.ApplicationCredentialSecret == "" {
```
stuggi: we are setting it to a string of the secret name in the webhook, even the secret might not be there, not sure if I understand the intend of this. shouldn't we only set the ApplicationCredentialSecret in the controller if AC is enabled and the secret got created?

stuggi: I think we should remove any defaulting via webhooks. if e.g. for glance AC gets enabled in its ReconcileGlance https://github.com/openstack-k8s-operators/openstack-operator/blob/main/internal/openstack/glance.go#L68 the operator should request the AC and check/wait until the AC is ready. if ready, it sets the parameter in the template. if it gets disabled, it removes the parameter in the template, or nils it. not sure what it is.

api/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -833,6 +946,97 @@ type WatcherSection struct {
...
+	ApplicationCredential *ServiceAppCredSection `json:"applicationCredential"`
+}
+
+// +kubebuilder:validation:XValidation:rule="self.gracePeriodDays < self.expirationDays",message="gracePeriodDays must be smaller than expirationDays"
+// ApplicationCredentialSection defines the desired configuration for ApplicationCredentials
+type ApplicationCredentialSection struct {
+	// Enabled indicates whether an ApplicationCredential should be created
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	Enabled bool `json:"enabled"`
```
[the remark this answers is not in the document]
auniyal61: I am inclined towards having a single switch, but double knob makes sense too.

do you think we can enable a logs for same, so suppose operator only enabled in respective service but they wont see AC's until the global is enabled too.
to debug they will start with operator logs of respective service - will it make sense to add an warning or error log saying global one is not enabled.

config/operator/deployment/kustomization.yaml
```diff
@@ -16,14 +16,20 @@ images:
...
   target:
     kind: Deployment
-    name: openstack-operator-controller-init
+    name: openstack-operator-controller-operator
+    namespace: system
+- patch: '[{"op": "replace", "path": "/spec/template/spec/containers/0/env/0", "value":
+    {"name": "OPENSTACK_RELEASE_VERSION", "value": "0.5.0-1767944881"}}]'
+  target:
+    kind: Deployment
+    name: openstack-operator-controller-operator
```
stuggi: I think we don't want to include the file config/operator/deployment/kustomization.yaml in the PR.

internal/openstack/barbican.go
```diff
@@ -61,6 +62,40 @@ func ReconcileBarbican(ctx context.Context, instance *corev1beta1.OpenStackContr
...
+	barbicanSecret := instance.Spec.Barbican.Template.Secret
+	if barbicanSecret == "" {
+		barbicanSecret = instance.Spec.Secret
+	}
+
+	acReady, acResult, err := EnsureApplicationCredentialForService(
+		ctx,
+		helper,
+		instance,
+		"barbican",
```
stuggi: should or could this be `barbican.Name`, could remove the need for another static string.

config/operator/deployment/kustomization.yaml
Deydra71: Yes, this is a generated remnant when I built the image, I will remove the changes here.

Deydra71:
Note: we need to update cinder, manila and heat after they are bumped, because we changed the placement of `Auth` spec in these service operators

PR #1442, opened 2025-05-14
Leader election lease tunables

dprince:
Set leader election LeaseDuration, RenewDealine and RetryPeriod via env variables.

The values used are the ones mentioned as recommended in:
  https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#high-availability

Co-authored-by: Dan Prince <dprince@redhat.com>

Jira: [OSPRH-16335](https://issues.redhat.com//browse/OSPRH-16335)

### DIFF
--- a/config/operator/deployment/deployment.yaml
+++ b/config/operator/deployment/deployment.yaml
@@ -69,6 +69,12 @@ spec:
           value: quay.io/openstack-k8s-operators/openstack-operator:latest
         - name: ENABLE_WEBHOOKS
           value: false
+        - name: LEASE_DURATION
+          value: 30
+        - name: RENEW_DEADLINE
+          value: 20
+        - name: RETRY_PERIOD
+          value: 5
         image: controller:latest
         name: operator
         securityContext:
--- a/bindata/operator/managers.yaml
+++ b/bindata/operator/managers.yaml
@@ -1,5 +1,8 @@
 {{ $namespace := .OperatorNamespace }}
 {{ $kubeRbacProxyImage := .KubeRbacProxyImage }}
+{{ $leaseDuration := .LeaseDuration }}
+{{ $renewDeadline := .RenewDeadline }}
+{{ $retryPeriod := .RetryPeriod }}
 {{ range $operatorName, $operatorImage := .OperatorImages }}
 apiVersion: apps/v1
 kind: Deployment
@@ -36,6 +39,12 @@ spec:
 {{ else }}
           value: 'false'
 {{ end }}
+        - name: LEASE_DURATION
+          value: '{{ $leaseDuration }}'
+        - name: RENEW_DEADLINE
+          value: '{{ $renewDeadline }}'
+        - name: RETRY_PERIOD
+          value: '{{ $retryPeriod }}'
         image: {{ $operatorImage }}
         livenessProbe:
           httpGet:
--- a/bindata/operator/operator.yaml
+++ b/bindata/operator/operator.yaml
@@ -76,6 +76,12 @@ spec:
         env:
         - name: OPENSTACK_RELEASE_VERSION
           value: '{{ .OpenstackReleaseVersion }}'
+        - name: LEASE_DURATION
+          value: '{{ .LeaseDuration }}'
+        - name: RENEW_DEADLINE
+          value: '{{ .RenewDeadline }}'
+        - name: RETRY_PERIOD
+          value: '{{ .RetryPeriod }}'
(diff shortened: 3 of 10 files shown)


dprince:
@hjensas Resurected and this PR works for me. I made a few adjustments. k8s env's like the values to be strings. I also split the envs into individual variables for the templates as this just seemed cleaner. Also the managers.yaml template is a bit special since it was nested loops.

I also adjusted our defaults to be a bit lower than you had them. I'm using 30/20/5 for the LEASE_DURATION/RENEW_DEADLINE/RETRY_PERIOD. This should be twice the previous defaults and lower the load a bit but also provide a reasonable restart/failover period. See: https://pkg.go.dev/k8s.io/client-go/tools/leaderelection If you want to make a change with this PR you just edit the CSV for openstack-operator and it applies them to all our operators accordingly. Increasing the settings would have the impact of making things get applied slower I think.

config/operator/deployment/deployment.yaml
```diff
@@ -69,6 +69,12 @@ spec:
           value: quay.io/openstack-k8s-operators/openstack-operator:latest
         - name: ENABLE_WEBHOOKS
           value: false
+        - name: LEASE_DURATION
```
dprince, line 72: @hjensas I'm inclined to decrease these back towards the defaults. Perhaps 30/20/5 would be a reasonable default for us. And then folks who want to try out OpenShift cluster high availability mode have that option which we could document. The defaults seem to be https://pkg.go.dev/k8s.io/client-go/tools/leaderelection 15/10/2 I think.

hjensas: ok, yeah my eyes when *blink blink* a couple of times when I saw how much difference there was the [openshift/library-go](https://github.com/openshift/library-go/blob/4b9033d00d37b88393f837a88ff541a56fd13621/pkg/config/leaderelection/leaderelection.go#L96-L109) compared to https://pkg.go.dev/k8s.io/client-go/tools/leaderelection. 

I also noticed in my testing that if the operators where already running - the "old" operator would terminate and then the new one would wait for over a minute to get the lock. It made me think, it would be good if controlled termination of a leader released/deleted the lease ... 

Let's try it a bit lower, 30/20/5 sounds good. I can test that a bit, and if I need to patch the csv in labs to bump it a bit more that is ok.

hjensas (approved):
In this job I see: https://softwarefactory-project.io/zuul/t/rdoproject.org/build/2f399e8aa8da40cfa514ec52a9dda98a

[openstack-operator-controller-operator](https://logserver.rdoproject.org/2f3/rdoproject.org/2f399e8aa8da40cfa514ec52a9dda98a/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack-operators/pods/openstack-operator-controller-operator-55cd9dfc9d-tgtv6/logs/operator.log)
```
2025-05-14T20:50:08.193Z	INFO	setup	manager configured with lease duration	{"seconds": 137}
2025-05-14T20:50:08.193Z	INFO	setup	manager configured with renew deadline	{"seconds": 107}
2025-05-14T20:50:08.193Z	INFO	setup	manager configured with retry period	{"seconds": 26}
```

And it propagates to the other operators.
[ironic-operator-controller-manager](https://logserver.rdoproject.org/2f3/rdoproject.org/2f399e8aa8da40cfa514ec52a9dda98a/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack-operators/pods/ironic-operator-controller-manager-69b97dbcf8-4k7vb/logs/manager.log)
```
2025-05-14T20:50:54Z	INFO	setup	manager configured with lease duration	{"seconds": 137}
2025-05-14T20:50:54Z	INFO	setup	manager configured with renew deadline	{"seconds": 107}
2025-05-14T20:50:54Z	INFO	setup	manager configured with retry period	{"seconds": 26}
```

config/operator/deployment/deployment.yaml
stuggi: > ok, yeah my eyes when _blink blink_ a couple of times when I saw how much difference there was the [openshift/library-go](https://github.com/openshift/library-go/blob/4b9033d00d37b88393f837a88ff541a56fd13621/pkg/config/leaderelection/leaderelection.go#L96-L109) compared to https://pkg.go.dev/k8s.io/client-go/tools/leaderelection.

thank you for that link. I was wondering how lose values were chosen

PR #1447, opened 2025-05-21
Bump default memory limit for controller manager to 512Mi

stuggi:
In tests it was seen that the openstack-operator controller manager reaches just bellow the current set limit of 256Mi. This change bumps the default to 512Mi for all controller managers, except the openstack-operator-controller-operator which could be customized using the CSV. It also introduces a MANAGER_MEMORY_LIMIT env var in the CSV which can be used to change the memory limit of all the other controller managers.

Jira: [OSPRH-16802](https://issues.redhat.com//browse/OSPRH-16802)

### DIFF
--- a/bindata/operator/managers.yaml
+++ b/bindata/operator/managers.yaml
@@ -3,6 +3,7 @@
 {{ $leaseDuration := .LeaseDuration }}
 {{ $renewDeadline := .RenewDeadline }}
 {{ $retryPeriod := .RetryPeriod }}
+{{ $managerMemoryLimit := .ManagerMemoryLimit }}
 {{ range $operatorName, $operatorImage := .OperatorImages }}
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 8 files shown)


dprince:
@stuggi should we only bump the controller-manager for openstack-operator-controller-manager which manages the largest amount or resources where we actually hit the limit. It seems like over reservation of memory for all controller managers could lead to less utilization in compact OCP clusters where we have namespace isolation.

bindata/operator/managers.yaml
```diff
@@ -62,7 +63,7 @@ spec:
         resources:
           limits:
             cpu: 500m
-            memory: 256Mi
+            memory: '{{ $managerMemoryLimit }}'
           requests:
             cpu: 10m
             memory: 128Mi
```
dprince, line 69: @stuggi also, I think you would want to bump 'requests' too right?

stuggi:
going with https://github.com/openstack-k8s-operators/openstack-operator/pull/1449 for now

PR #1455, opened 2025-05-26
Add delay on checking BAD_OR_MISSING_NODE_SELECTOR_COUNT or service certs

danpawlik:
The script might make some troubles for OpenShift API to reply
when there is no delay on repeating command.
Also add few seconds delay on checking service certs to avoid errors on
some infrastructures like:

        case.go:398: command "echo \"Checking rotation of non API service certificates...\"
        NAMES..." exceeded 3 sec timeout, context deadline exceeded

### DIFF
--- a/tests/kuttl/common/osp_check_noapi_service_certs.sh
+++ b/tests/kuttl/common/osp_check_noapi_service_certs.sh
@@ -73,6 +73,7 @@ for database in "${!database_secrets[@]}"; do
 
     for port in $ports; do
         echo "Connecting to $database on port $port..."
+        sleep 5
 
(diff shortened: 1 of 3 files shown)


danpawlik:
recheck

danpawlik:
recheck

tests/kuttl/common/osp_check_noapi_service_certs.sh
```diff
@@ -56,6 +56,7 @@ done
 
 # Gather the ClusterIP and ports for databases
 for database in "${!database_secrets[@]}"; do
+    sleep 5
```
stuggi, line 59: just wondering whats different to the db checks then to the above services. I mean should we have those sleeps also in the above?

danpawlik: right, maybe here it is not needed - it should handle.
Next steps - iterating on ports is most problematic.

evallesp (approved):

PR #1456, opened 2025-05-26
Change parent for openstack-operator-kuttl job

danpawlik:
The parent should be different to make the tests running.


abays (approved):
/lgtm

PR #1462, opened 2025-05-29
Add retry on getting ISSUER name

danpawlik:
It happens on some infras that the issuer is empty. Probably it does not have enough time to apply changes.
This commit adds retry for getting the ISSUER name.


stuggi (approved):
/lgtm

PR #1483, opened 2025-06-06
Add nproc equivalent for macOS

ciecierski:
Add hw.logicalcpu for NUM_PROCS functional parrarel run to Makefile. hw.logicalcpu is equivalent of nproc on macOS

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,7 +188,13 @@ golangci-lint:
 	$(LOCALBIN)/golangci-lint run --fix
 
 MAX_PROCS := 5
+UNAME_S := $(shell uname -s)
+ifeq ($(UNAME_S),Linux)
 NUM_PROCS := $(shell expr $(shell nproc --ignore 2) / 2)
+endif
+ifeq ($(UNAME_S),Darwin)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -188,7 +188,13 @@ golangci-lint:
 	$(LOCALBIN)/golangci-lint run --fix
 
 MAX_PROCS := 5
+UNAME_S := $(shell uname -s)
+ifeq ($(UNAME_S),Linux)
 NUM_PROCS := $(shell expr $(shell nproc --ignore 2) / 2)
+endif
+ifeq ($(UNAME_S),Darwin)
+NUM_PROCS := $(shell expr $(shell sysctl -n hw.logicalcpu) / 2)
```
bshephar, line 196: I was personally just installing `coreutils` from homebrew, which contains `nproc`:
```
❯ nproc --ignore 2
8
❯ uname -s
Darwin

```

Which comes from:
https://formulae.brew.sh/formula/coreutils
https://github.com/coreutils/coreutils/blob/037b93106155c5e11681837305b541808fbdb9b5/src/nproc.c

PR #1485, opened 2025-06-11
Do not update status when reconciler exits due to panic

ratailor:
Currently we use deferred call to always update the status field when Reconcile call exits, which seems correct for any other error except when Reconciler exits due to panic.

This change adds a call to recover function and try to handle panic and log error.

Closes: OSPRH-17071


ratailor:
recheck

PR #1493, opened 2025-06-18
ORC operator integration

dprince:
Install the ORC operator as part of our initialization resource


dprince:
This uses the preliminary ORC bundle from https://github.com/k-orc/openstack-resource-controller/pull/447 to demonstrate installing ORC as part of the openstack-operators initialization resource

dprince:
> Could potentially mark this as DNM for the current release window

yes, I did not mark as draft because I wanted CI to run :)

PR #1495, opened 2025-06-23
[OSPRH-16305] Introduce OpenStackLightspeed

lpiwowar:
Introduce the OpenStackLightspeed custom resource that manages the
deployment of Lightspeed-related functionalities. As of now, we expect
the OLS operator to be present in the cluster. Otherwise, enabling of
OpenStackLightspeed will fail. The Lightspeed functionalities can be
enabled through the OpenStackLightspeed as follows:

```
spec:
  llmEndpoint: <URL of the LLM>
  llmEndpointType: <endpoint type>
  llmCredentials: <secret containing token>
  modelName: <model name>
  tlsCACertBundle: <config map containg cert>
```

Only one OpenStackLightspeed CR can configure OpenStack Lightspeed at a
single time. If the user tries to manage the Lightspeed functionalities
through two or more OpenStackLightspeed instances, the user will be
warned, and the changes from the newest OpenStackLightspeed instance
will be ignored.


ratailor (commented):
Please fix one minor nit in comment and other for adding the pattern, we replicated in all operators, when you respin the PR.

controllers/lightspeed/openstacklightspeed_controller.go
```diff
@@ -0,0 +1,256 @@
...
+	OLSOperatorInstalled, err := lightspeed.IsOLSOperatorInstalled(ctx, helper)
+	if !OLSOperatorInstalled || err != nil {
+		return ctrl.Result{}, fmt.Errorf("installation of OpenShift LightSpeed not detected: %w", err)
+	}
+
+	// NOTE: We cannot consume the OLSConfig definition directly from the OLS operator's code due to
+	// a conflict in Go versions. When this comment was written, the min. required Go version for
+	// openstack-operator was 1.21 whereas OLS operator required at least Go version 1.23. Once the
+	// Go versions catch up with each other we should consider consuming OLSConfig directly from OLS
+	// operator and updating this code and any subsequent code that consumes this structure.
```
dprince, line 149: using the k8s unstructured client to avoid dependencies is a valid thing to do. Even if we temporarily get our golang version aligned with lightspeed it is worth considering whether directly importing the project for just this struct is worth it.

controllers/core/openstackcontrolplane_controller.go
```diff
@@ -484,6 +486,13 @@ func (r *OpenStackControlPlaneReconciler) reconcileNormal(ctx context.Context, i
 		return ctrlResult, nil
 	}
 
+	ctrlResult, err = openstack.ReconcileLightSpeed(ctx, instance, version, helper)
```
dprince, line 489: It isn't clear to me that the LightSpeed configuration is specifically related to our ControlPlane CRs. It seems you would have 1 instance of Lightspeed per cluster... not per namespace right? So I wonder if instead of having your integration point here it might make sense to do this in the OpenStack initialization resource which is 1 per cluster. The point of the existing initialization resource is to install operators but it could also create inject the openstack lightspeed configuration as well. It could be enabled/disabled via a parameter in the spec of that resource

dprince (changes requested):

controllers/core/openstackcontrolplane_controller.go
lpiwowar: > It seems you would have 1 instance of Lightspeed per cluster... not per namespace right?

Yes, but it will not be the case in the future. 

We are currently limited by OpenShift Lightspeed, which is missing some functionality that does not allow us to do per-namespace configuration. However, the long-term goal is to allow different configurations of OpenStack Lightspeed for different OpenStack deployments on top of the same OpenShift cluster. 

Ultimately, OpenStack Lightspeed falls into a similar bucket as `Horizon` or `OpenStackClient`. Just a different interface to interact with OpenStack.

controllers/lightspeed/openstacklightspeed_controller.go
lpiwowar: Ack, it makes sense, thanks! Should I get rid of the comment, or is it OK if we keep it here?

apis/lightspeed/v1beta1/openstacklightspeed_types.go
```diff
@@ -0,0 +1,115 @@
...
+	// in the secret that holds the token should be "apitoken".
+	LLMCredentials *string `json:"llmCredentials"`
+
+	// +kubebuilder:validation:Required
+	// Name of the model to use at the API endpoint provided in LLMEndpoint
+	ModelName *string `json:"modelName"`
+
+	// +kubebuilder:validation:Required
+	// Configmap name containing a CA Certificates bundle
+	TLSCACertBundle *string `json:"TLSCACertBundle"`
```
stuggi, line 64: json should be `tlsCACertBundle`

stuggi: * which CA bundle is this? 
* the one from the ctlplane?
* I guess it would required the CA for the ctlplane, but also for the llm endpoint?
* should it lookup the CA bundle from the ctlplane?
* should the user add the llm endpoint CA to the ctlplane and then the full bundle managed by the ctlplane gets added to the container here?

apis/lightspeed/v1beta1/webhook_suite_test.go
```diff
@@ -0,0 +1,143 @@
+/*
+Copyright 2025.
```
stuggi, line 2: we do all envtest via `test/functional` so could either drop this file or shorten it as https://github.com/openstack-k8s-operators/openstack-operator/blob/main/apis/core/v1beta1/webhook_suite_test.go ?

pkg/lightspeed/funcs.go
```diff
@@ -0,0 +1,237 @@
...
+			"type": StringPtrValue(instance.Spec.LLMEndpointType),
+			"url":  StringPtrValue(instance.Spec.LLMEndpoint),
+		},
+	}
+
+	if err := uns.SetNestedSlice(olsConfig.Object, providersPatch, "spec", "llm", "providers"); err != nil {
+		return err
+	}
+
+	indexID, err := getIndexID(instance.Spec.RAGImage)
```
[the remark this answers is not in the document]
stuggi: we need to use SHA in the RELATED images to be able to support disconnected environments where images can be automatic mirrored using the SHA. not sure if that is conflicting with this?

pkg/lightspeed/funcs.go
```diff
@@ -0,0 +1,237 @@
...
+			"indexID":   indexID,
+			"indexPath": OpenStackLightspeedVectorDBPath,
+		},
+	}
+
+	if err := uns.SetNestedSlice(olsConfig.Object, openstackRAG, "spec", "ols", "rag"); err != nil {
+		return err
+	}
+
+	tlsCaCertBundle := StringPtrValue(instance.Spec.TLSCACertBundle)
```
stuggi, line 145: same questions as in the api file, if this should be the CA bundle from the ctlplane instead.

pkg/lightspeed/funcs.go
```diff
@@ -0,0 +1,237 @@
...
+func PatchOLSConfig(
+	olsConfig *uns.Unstructured,
+	instance *lightspeedv1beta1.OpenStackLightspeed,
+	helper *common_helper.Helper,
+) error {
+	// 1. Patch the Providers section
+	providersPatch := []interface{}{
+		map[string]interface{}{
+			"credentialsSecretRef": map[string]interface{}{
+				"name": StringPtrValue(instance.Spec.LLMCredentials),
```
stuggi, line 109: is it good to use StringPtrValue and possibly pass an empty string as the name? wouldn't it be better to check all the required files for being nil and return an explicite error? or is the intend to get the error from the api server when try to apply the obj with the missing required params?

stuggi (commented):
maybe a general question for my understanding of OpenStackLightspeed. is it intended that it talks to the openstack api of the ctlplane it gets installed with?

pkg/lightspeed/funcs.go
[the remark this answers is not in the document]
lpiwowar: Ok, that is the response I was afraid of. Yes, it is in conflict. But I understand, what you are saying makes sense.

This is the image that we want to use `quay.io/openstack-lightspeed/rag-content:os-docs-2024.2`. The tag contains information about what `indexID` value to use when consuming this image. Ultimately, this is what gets passed to the `OLSConfig`:

```
- name: quay.io/openstack-lightspeed/rag-content:os-docs-2024.2
  indexID: os-docs-2024.2
```

The value of `indexID` might be different for different images. The value is also present in ENV var inside the image [1]. I'll try to find the best way how to access the ENV var or an alternative to this. 

[1] https://github.com/openstack-lightspeed/rag-content/blob/5e8339870c3600120bf6080e68a6ffb14a115d61/Containerfile#L53

apis/lightspeed/v1beta1/webhook_suite_test.go
lpiwowar: Ok, thanks, I'll update the file!

lpiwowar:
>  Is it intended that it talks to the openstack api of the ctlplane it gets installed with?

@stuggi: Not right now, but this might come in the future.

pkg/lightspeed/funcs.go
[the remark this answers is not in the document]
lpiwowar: Ok, so for now, I've introduced a temporary solution that discovers the `indexID` value by spawning a pod with the RAG image. I do not like it, but it is a temporary thing. I've talked with the OpenShift Lightspeed team and the `IndexID` parameter will most likely disappear from `OLSConfig` in the future or will be marked as optional. Then we can remove the autodiscovery of the `indexID` value here in openstack-oeprator.

pkg/lightspeed/funcs.go
lpiwowar: I've added the checks for nil for the Optional parameters to ensure that we do not pass `""` string to those if they are not set. For the other parameters, I think we should be good because they are marked as required and if they are not set they would be caught during the validation of the incoming `OpenStackControlPlane`. I'm marking as resolved. Is this resolution ok?

lpiwowar:
> Just out of curiosity, why do we want integrate it with the openstack-operator if there is no interaction with the environments it deploys. What is the benefit from it?

I understand, I think I would be asking the same. We know that now it won't interact with the OpenStack cluster, but we also know that it will interact with the OpenStack env in the future. There are also additional stuff that we might want to deploy using openstack-operator when enabling the OpenStack Lightspeed like MCP servers (but this will come in the future).

We need to lay some foundational code that we can extend as we move forward with OpenStack LightSpeed. This is the whole purpose of this PR. Lay something we can build on top of and improve.

What do you think about the code @stuggi? Is there some things you want me to change?

controllers/lightspeed/openstacklightspeed_controller.go
```diff
@@ -0,0 +1,272 @@
...
+	return ctrl.Result{}, nil
+}
+
+// reconcileDelete reconciles the deletion of OpenStackLightspeed instance
+func (r *OpenStackLightspeedReconciler) reconcileDelete(
+	ctx context.Context,
+	helper *common_helper.Helper,
+	instance *lightspeedv1beta1.OpenStackLightspeed,
+) (ctrl.Result, error) {
+	if ok := controllerutil.RemoveFinalizer(instance, helper.GetFinalizer()); !ok {
```
stuggi, line 230: removing the finalizer on the instance should be the last thing in the reconcileDelete, after all the cleanup steps are complete.
Also the return value indicates of whether it updated the object's list of finalizers. I don't think return an error in case of `!ok` is right. It would mean that there was no finalizer and its an error. I think you could just call it. I general we'd have to call patch or update to write it, but since it gets deleted, it should be ok to not update itself.

controllers/lightspeed/openstacklightspeed_controller.go
```diff
@@ -0,0 +1,272 @@
...
+		return ctrl.Result{}, fmt.Errorf("remove finalizer failed")
+	}
+
+	olsConfig, err := lightspeed.GetOLSConfig(ctx, helper)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return ctrl.Result{}, err
+	} else if err != nil {
+		return ctrl.Result{}, nil
+	}
```
stuggi, line 239: not sure I understand this return. the above returns if there is an error and it is not `IsNotFound` . Do we have to return on `IsNotFound`?

I think what you want to do is to only run the bellow if no `IsNotFound` error happened. Otherwise you won't have an instance of `olsConfig`.

controllers/lightspeed/openstacklightspeed_controller.go
```diff
@@ -0,0 +1,272 @@
...
+
+	ownerLabel := olsConfig.GetLabels()[lightspeed.OpenStackLightspeedOwnerIDLabel]
+	if ownerLabel == "" {
+		return ctrl.Result{}, fmt.Errorf("OLSConfig is not managed by OpenStackLightspeed instance")
+	} else if ownerLabel != string(instance.GetObjectMeta().GetUID()) {
+		return ctrl.Result{}, fmt.Errorf("OLSConfig is managed by different OpenStackLightspeed instance")
+	}
```
stuggi, line 246: wouldn't it be enough to check the Owner UUID? could do a check similar to https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/openstack/common.go#L451-L458

stuggi: do we set the owner somewhere? could not find it, but maybe I have missed it

controllers/lightspeed/openstacklightspeed_controller.go
```diff
@@ -0,0 +1,272 @@
...
+func (r *OpenStackLightspeedReconciler) GetLogger(ctx context.Context) logr.Logger {
+	return log.FromContext(ctx).WithName("Controllers").WithName("OpenStackLightspeed")
+}
+
+// +kubebuilder:rbac:groups=lightspeed.openstack.org,resources=openstacklightspeeds,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=lightspeed.openstack.org,resources=openstacklightspeeds/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=lightspeed.openstack.org,resources=openstacklightspeeds/finalizers,verbs=update
+// +kubebuilder:rbac:groups=ols.openshift.io,resources=olsconfigs,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=ols.openshift.io,resources=olsconfigs/status,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=ols.openshift.io,resources=olsconfigs/finalizers,verbs=get;list;watch;create;update;patch;delete
```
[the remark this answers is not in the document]
lpiwowar: True, thank you! Done:)

controllers/lightspeed/openstacklightspeed_controller.go
lpiwowar: Honestly, I do not understand it either in the hindsight. I updated the logic. Thank you!

controllers/core/openstackcontrolplane_controller.go
dprince: I think it would be okay to have Lightspeed configured "alongside" the Controlplane. But I do not think it is a core service and thus should not be part of the Controlplane. I'm -2 on implementing the feature this way at this time.

lpiwowar: > But I do not think it is a core service and thus should not be part of the Controlplane.

Thanks for the response. In that case, I have some follow questions, please:

- What level of integration would be acceptable for you at this moment? Are we talking just the OpenStackLightspeed CR without integration in the OpenStackControlPlane CR?

- How does the OpenStack Lightpseed differ for you from Horizon or OpenStackClient? In the end, it is planned to be another interface for the deployed Openstack. I'm just trying to understand the reasoning.

The first one is probably the most important one because our team needs to find a way how to do this initial integration somehow.

apis/core/v1beta1/conditions.go
```diff
@@ -325,6 +328,15 @@ const (
 	// OpenStackControlPlaneClientReadyErrorMessage
 	OpenStackControlPlaneClientReadyErrorMessage = "OpenStackControlPlane Client error occured %s"
 
+	// OpenStackControlPlaneLightSpeedReadyErrorMessage
+	OpenStackControlPlaneLightSpeedReadyErrorMessage = "OpenStackControlPlane LightSpeed error occured %s"
```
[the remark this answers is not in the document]
lpiwowar: done

apis/lightspeed/v1beta1/openstacklightspeed_types.go
```diff
@@ -0,0 +1,116 @@
...
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Provider Type"
+	// Type of the provider serving the LLM
+	LLMEndpointType *string `json:"llmEndpointType"`
+
+	// +kubebuilder:validation:Required
+	// Name of the model to use at the API endpoint provided in LLMEndpoint
+	ModelName *string `json:"modelName"`
+
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Optional
```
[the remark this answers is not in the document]
lpiwowar: Thanks!:) Left-over comment the CRD was generated properly. But surely only one comment needs to stay here.

controllers/core/openstackcontrolplane_controller.go
dprince: not because it is an AI component as much as it seems like it is a *new* component which could change greatly in a short period of time. We want the OpenStackControlplane to be as stable as we can make it.

If you need the configuration to be namespaced I think that is easily solved by simply having a CR in each namespace it is required. It doesn't mean we have to keep lightspeed in the same CR as the OpenStackControlplane right?

I think the suggestion here is somewhat minor. Keep the lightspeed functionality in the openstack-operator as a separate set API/controllers

lpiwowar:
recheck

```
  - message: 'error using catalogsource openstack-operators/openstack-operator-index:
      failed to list bundles: rpc error: code = Unavailable desc = connection error:
      desc = "transport: Error while dialing: dial tcp 10.217.5.115:50051: connect:
      connection refused"'
```

https://logserver.rdoproject.org/eda/rdoproject.org/eda2155606d240c8837a6381f718bdeb/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack-operators/subscriptions/openstack-operator

lpiwowar:
recheck

lpiwowar:
@dprince or @stuggi can I please ask for a review when you have time? The PR has been updated to drop the OpenStackControlPlane changes.

lpiwowar (commented):
Thanks @stuggi, for taking a look. I addressed your comments.

controllers/lightspeed/openstacklightspeed_controller.go
```diff
@@ -0,0 +1,332 @@
...
+
+	olsConfig.SetGroupVersionKind(olsConfigGVK)
+	olsConfig.SetName(lightspeed.OLSConfigName)
+
+	_, err = controllerutil.CreateOrPatch(ctx, r.Client, &olsConfig, func() error {
+		// Check if the OpenStackLightspeed instance that is being processed owns the OLSConfig. If
+		// it is owned by other OpenStackLightspeed instance stop the reconciliation.
+		olsConfigLabels := olsConfig.GetLabels()
+		ownerLabel := ""
+		if val, ok := olsConfigLabels[lightspeed.OpenStackLightspeedOwnerIDLabel]; ok {
```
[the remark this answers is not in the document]
stuggi: you are right, forgot about that it returns the zero value for the map's value type.

lpiwowar:
recheck

/retest

lpiwowar:
@dprince if you could please take a look once you have time, all the @stuggi's comments should be addressed :pray:

apis/core/v1beta1/openstackversion_types.go
```diff
@@ -166,6 +166,7 @@ type ContainerTemplate struct {
 	TestTobikoImage                   *string `json:"testTobikoImage,omitempty"`
 	TestHorizontestImage              *string `json:"testHorizontestImage,omitempty"`
 	TestAnsibletestImage              *string `json:"testAnsibletestImage,omitempty"`
+	OpenstackLightspeedImage          *string `json:"openstackLightspeedImage,omitempty"`
```
dprince, line 169: I think it would be good to drop the parameter from the OpenStackVersion and Controlplane resources and just have Lightspeed obtain its Image directly from the ENV variable. This would also work in the case where Lightspeed gets installed ahead of the Controlplane if that use case materializes.

dprince (changes requested):

dprince (approved):

PR #1503, opened 2025-06-25
Check if OctaviaRsyslogImage is not nil

stuggi:
OctaviaRsyslog was introduced in FR2 time frame. If there is an update with enabled octavia from before, the current targetVerion images won't have an OctaviaRsyslogImage which results in a nil ptr dereference error and the controller manage panics.

This adds a check, like done in other places if the images is non nil and otherwise falls back to an empty string.

Jira: [OSPRH-17647](https://issues.redhat.com//browse/OSPRH-17647)


abays (approved):
/lgtm

PR #1507, opened 2025-06-25
upgrade catalog: use digest

dprince:
This PR does 2 things:

-converts image URLS into digest versions. This should pin the
 references to when they got built

-adds a sha/digest tag to the catalog index. This should fix
 persistence issues and make it possible to use a catalog
 from a specific sha/digest without it getting deleted on quay.io


stuggi (approved):
/lgtm

PR #1518, opened 2025-06-30
Fix rabbitmq api version

olliewalsh:
Should be v2.9.0 as we use a fork pinned to this version

### DIFF
--- a/apis/go.mod
+++ b/apis/go.mod
@@ -83,7 +83,7 @@ require (
 	github.com/prometheus/client_model v0.6.0 // indirect
 	github.com/prometheus/common v0.53.0 // indirect
 	github.com/prometheus/procfs v0.13.0 // indirect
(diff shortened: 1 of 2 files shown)


apis/go.mod
```diff
@@ -83,7 +83,7 @@ require (
 	github.com/prometheus/client_model v0.6.0 // indirect
 	github.com/prometheus/common v0.53.0 // indirect
 	github.com/prometheus/procfs v0.13.0 // indirect
-	github.com/rabbitmq/cluster-operator/v2 v2.11.0 // indirect
+	github.com/rabbitmq/cluster-operator/v2 v2.9.0 // indirect
```
stuggi, line 86: we have https://github.com/openstack-k8s-operators/renovate-config/blob/main/default.json5#L155 to keep it bellow 2.10. but the replace bellow will override it anyways

PR #1519, opened 2025-07-03
Add interface to allow customize of operator deployments

stuggi:
Adds an interface which allows to customize the following parameters of a controller-manager operator deployment
- change the replica (0/1)
- change the resources (requests/limits) of the manager container

It keeps the current default resource requests as is. For easier gotemplate it this adds an internal custom struct type, which allows to just loop over a list of operator.Operator for service operators and a dedicated instance for the openstack-operator-controller-manager and the rabbitmq-cluster-operator-manager.

E.g. changing the default cpu limits for the infra operator and disabling the openstack-baremetal operator would be: 
```
...
spec:
  operatorOverrides:
  - name: infra
    controllerManager:
      resources:
        limits:
          cpu: 600m
          memory: 2Gi
    replicas: 1
  - name: openstack-baremetal
    replicas: 0
```

To add a new service operator has the following high level steps:
- in the `openstack_types.go` api
  - add a new const with the name
  - add an entry for the defaults in the OperatorList var
  - add the name also to the Enum of the OperatorSpec.Name
- (optional) if any specific env var need be passed to the operator, e.g. related images list, or enable webhooks, check the applyOperator() in the `openstack_controller.go` .

Jira: OSPRH-17785


stuggi:
recheck

PR #1536, opened 2025-07-17
Use structured logging

ratailor:
This change uses GetLogger instead of direct use of log.FromContext to support structured logging.


ratailor:
recheck

PR #1539, opened 2025-07-21
Make script input vars mandatory

stuggi:
Make MAIN_VERSION, FEATURE_RELEASE_VERSION, and FEATURE_RELEASE_BRANCH mandatory.

When run via the job, those vars get set in the workflow [1]. Lets make them mandatory, instead of bumping them in the workflow and the script on a new feature branch. If run manually they should be set.

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/5f8fbc7cd3b79a363d9fc2016e6e3c5b239b4d5a/.github/workflows/catalog-openstack-operator-upgrades.yaml#L68

Jira: [OSPRH-18397](https://issues.redhat.com//browse/OSPRH-18397)

### DIFF
--- a/hack/catalog-build-olm-upgrade.sh
+++ b/hack/catalog-build-olm-upgrade.sh
@@ -10,9 +10,11 @@ echo $URL | sed -e "s|:.*|@$DIGEST|"
 
 }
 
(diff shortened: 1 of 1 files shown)


abays (approved):
/lgtm

PR #1543, opened 2025-07-24
Fix some minor issues in ipam.go

rabi:
- Fix flawed conditional in createOrPatchDNSData()
- Rmove nested for loop when deleting ipsets
- break when ctlplane network is found in reserveIPs()
- Fix SA1006 and other staticcheck linter issues


stuggi (approved):
/lgtm

rabi:
phew `No clusters in pool are ready to be claimed` .. We have too many prow jobs now and have limited infra.. Can we reduce them please unless we plan to add to the infra ?

PR #1548, opened 2025-07-29
Switch Network Exporter image to be generic

karelyatin:
Network Exporter being extended to be used on Control plane pods, so make it generic.

Resolves: #[OSPRH-18529](https://issues.redhat.com//browse/OSPRH-18529)


pkg/dataplane/inventory.go
```diff
@@ -349,7 +349,7 @@ func resolveGroupAnsibleVars(template *dataplanev1.NodeTemplate, group *ansible.
 		group.Vars["edpm_telemetry_podman_exporter_image"] = containerImages.EdpmPodmanExporterImage
 	}
 	if template.Ansible.AnsibleVars["edpm_telemetry_openstack_network_exporter_image"] == nil {
-		group.Vars["edpm_telemetry_openstack_network_exporter_image"] = containerImages.EdpmOpenstackNetworkExporterImage
+		group.Vars["edpm_telemetry_openstack_network_exporter_image"] = containerImages.OpenstackNetworkExporterImage
```
stuggi, line 352: depending on the order of reconciliation of the CRs when the new operator version starts, like in getOpenstackNetworkExporterImg, we should probably check if `containerImages.OpenstackNetworkExporterImage` not being nil. In the case where the osversion `status.containerImages` was not yet updated to have the OpenstackNetworkExporterImage it might be.

PR #1552, opened 2025-08-06
Fix tolerations API placement and implementation

stuggi:
Move tolerations from ContainerSpec to OperatorSpec level since tolerations are deployment/pod-level configuration, not container-level.

Changes:
- Move Tolerations field from ContainerSpec to OperatorSpec in API types
- Update override logic to use correct field path (opOvr.Tolerations)
- Fix test cases to use OperatorSpec.Tolerations instead of ControllerManager.Tolerations
- Fix YAML indentation in sample configuration

This corrects the API design to follow Kubernetes conventions where tolerations are specified at the pod template level, not per container.


abays (approved):
/lgtm

PR #1554, opened 2025-08-06
[OSPRH-18744] Handle variant apiVersions for refData cleanup

abays:
Some `refData` in `Operator` resources can lack a group in its `apiVersion`.  We need to adjust our `Operator` clean-up logic to account for this.

Jira: https://issues.redhat.com/browse/OSPRH-18744


stuggi (approved):
/lgtm

PR #1571, opened 2025-08-20
Create TLS certs for OVN exporter services

karelyatin:
As part of Metrics TLS support in the Related-Issue, exporter will
be enabled per pod, so created a dedicated cert for this.

Related-Issue: [OSPRH-12568](https://issues.redhat.com//browse/OSPRH-12568)

### DIFF
--- a/pkg/openstack/ovn.go
+++ b/pkg/openstack/ovn.go
@@ -36,6 +36,15 @@ func ReconcileOVN(ctx context.Context, instance *corev1beta1.OpenStackControlPla
 		instance.Spec.Ovn.Template = &corev1beta1.OvnResources{}
 	}
 
+	// Create TLS certificate for OVN metrics services when TLS is enabled
+	if instance.Spec.Ovn.Enabled && instance.Spec.TLS.PodLevel.Enabled {
+		if err := EnsureOVNMetricsCert(ctx, instance, helper); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/openstack/ovn.go
```diff
@@ -365,9 +366,10 @@ func ReconcileOVNController(ctx context.Context, instance *corev1beta1.OpenStack
 		certRequest := certmanager.CertificateRequest{
 			IssuerName: instance.GetOvnIssuer(),
 			CertName:   fmt.Sprintf("%s-ovndbs", "ovncontroller"),
+			// Cert needs to be valid for the individual pod services in the daemonset so make this a wildcard cert
 			Hostnames: []string{
-				fmt.Sprintf("%s.%s.svc", serviceName, instance.Namespace),
-				fmt.Sprintf("%s.%s.svc.%s", serviceName, instance.Namespace, dnsSuffix),
+				fmt.Sprintf("%s*.%s.svc", serviceName, instance.Namespace),
+				fmt.Sprintf("%s*.%s.svc.%s", serviceName, instance.Namespace, dnsSuffix),
```
dprince, line 372: will this cause an immediate restartof the OVN Controller when openstack-operator is updated? If so should we add code such that it only occurs during the minor update window?

dprince (changes requested):

PR #1591, opened 2025-09-05
Add NotificationsBusInstance API field

fmount:
This patch introduces the `notificationsBusInstance` (optional) parameter at `API` level.
When an override is not defined, it is propagated to the underlying storage components where this field is implemented.

### DIFF
--- a/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
+++ b/apis/bases/core.openstack.org_openstackcontrolplanes.yaml
@@ -9582,6 +9582,8 @@ spec:
                 additionalProperties:
                   type: string
                 type: object
+              notificationsBusInstance:
+                type: string
               nova:
                 properties:
                   apiOverride:
--- a/apis/core/v1beta1/openstackcontrolplane_types.go
+++ b/apis/core/v1beta1/openstackcontrolplane_types.go
@@ -128,6 +128,13 @@ type OpenStackControlPlaneSpec struct {
 	Rabbitmq RabbitmqSection `json:"rabbitmq,omitempty"`
 
 	// +kubebuilder:validation:Optional
+	// NotificationsBusInstance - the name of RabbitMQ Cluster CR to select a Messaging
+	// Bus Service instance used by all services that produce or consume notifications.
+	// Avoid colocating it with RabbitMQ services used for PRC.
+	// That instance will be pushed down for services, unless overriden in templates.
+	NotificationsBusInstance *string `json:"notificationsBusInstance,omitempty"`
+
+	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// Memcached - Parameters related to the Memcached service
 	Memcached MemcachedSection `json:"memcached,omitempty"`
--- a/pkg/openstack/cinder.go
+++ b/pkg/openstack/cinder.go
@@ -130,6 +130,12 @@ func ReconcileCinder(ctx context.Context, instance *corev1beta1.OpenStackControl
 		instance.Spec.Cinder.Template.TopologyRef = instance.Spec.TopologyRef
 	}
 
+	// When no NotificationsBusInstance is referenced in the subCR (override)
+	// try to inject the top-level one if defined
+	if instance.Spec.Cinder.Template.NotificationsBusInstance == nil {
+		instance.Spec.Cinder.Template.NotificationsBusInstance = instance.Spec.NotificationsBusInstance
+	}
+
 	Log.Info("Reconciling Cinder", "Cinder.Namespace", instance.Namespace, "Cinder.Name", cinderName)
 	op, err := controllerutil.CreateOrPatch(ctx, helper.GetClient(), cinder, func() error {
 		instance.Spec.Cinder.Template.CinderSpecBase.DeepCopyInto(&cinder.Spec.CinderSpecBase)
--- a/pkg/openstack/glance.go
+++ b/pkg/openstack/glance.go
@@ -74,6 +74,12 @@ func ReconcileGlance(ctx context.Context, instance *corev1beta1.OpenStackControl
 		instance.Spec.Glance.Template.TopologyRef = instance.Spec.TopologyRef
 	}
 
+	// When no NotificationsBusInstance is referenced in the subCR (override)
+	// try to inject the top-level one if defined
+	if instance.Spec.Glance.Template.NotificationBusInstance == nil {
+		instance.Spec.Glance.Template.NotificationBusInstance = instance.Spec.NotificationsBusInstance
+	}
+
 	// When component services got created check if there is the need to create a route
 	if err := helper.GetClient().Get(ctx, types.NamespacedName{Name: glanceName, Namespace: instance.Namespace}, glance); err != nil {
 		if !k8s_errors.IsNotFound(err) {
--- a/apis/core/v1beta1/zz_generated.deepcopy.go
+++ b/apis/core/v1beta1/zz_generated.deepcopy.go
@@ -1178,6 +1178,11 @@ func (in *OpenStackControlPlaneSpec) DeepCopyInto(out *OpenStackControlPlaneSpec
 	in.Cinder.DeepCopyInto(&out.Cinder)
 	in.Galera.DeepCopyInto(&out.Galera)
 	in.Rabbitmq.DeepCopyInto(&out.Rabbitmq)
+	if in.NotificationsBusInstance != nil {
+		in, out := &in.NotificationsBusInstance, &out.NotificationsBusInstance
+		*out = new(string)
+		**out = **in
+	}
 	in.Memcached.DeepCopyInto(&out.Memcached)
 	in.Ovn.DeepCopyInto(&out.Ovn)
 	in.Neutron.DeepCopyInto(&out.Neutron)
--- a/bindata/crds/crds.yaml
+++ b/bindata/crds/crds.yaml
@@ -9746,6 +9746,8 @@ spec:
                 additionalProperties:
                   type: string
                 type: object
+              notificationsBusInstance:
+                type: string
               nova:
                 properties:
                   apiOverride:
--- a/config/crd/bases/core.openstack.org_openstackcontrolplanes.yaml
+++ b/config/crd/bases/core.openstack.org_openstackcontrolplanes.yaml
@@ -9582,6 +9582,8 @@ spec:
                 additionalProperties:
                   type: string
                 type: object
+              notificationsBusInstance:
+                type: string
               nova:
                 properties:
                   apiOverride:
--- a/pkg/openstack/manila.go
+++ b/pkg/openstack/manila.go
@@ -118,6 +118,12 @@ func ReconcileManila(ctx context.Context, instance *corev1beta1.OpenStackControl
 		instance.Spec.Manila.Template.TopologyRef = instance.Spec.TopologyRef
 	}
 
+	// When no NotificationsBusInstance is referenced in the subCR (override)
+	// try to inject the top-level one if defined
+	if instance.Spec.Manila.Template.NotificationsBusInstance == nil {
+		instance.Spec.Manila.Template.NotificationsBusInstance = instance.Spec.NotificationsBusInstance
+	}
+
 	Log.Info("Reconciling Manila", "Manila.Namespace", instance.Namespace, "Manila.Name", "manila")
 	op, err := controllerutil.CreateOrPatch(ctx, helper.GetClient(), manila, func() error {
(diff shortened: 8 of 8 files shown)


apis/bases/core.openstack.org_openstackcontrolplanes.yaml
```diff
@@ -9582,6 +9582,8 @@ spec:
                 additionalProperties:
                   type: string
                 type: object
+              notificationsBusInstance:
```
fmount, line 9585: @gibizer @bogdando is there a reason behind the fact that the other patch calls this field `notificationsBus` instead of `notificationsBusInstance` like we did in the service operators?

fmount: ack, then let's keep it consistent with service operators and call it `notificationsBusInstance`.

apis/core/v1beta1/openstackcontrolplane_types.go
```diff
@@ -127,6 +127,13 @@ type OpenStackControlPlaneSpec struct {
 	// Rabbitmq - Parameters related to the Rabbitmq service
 	Rabbitmq RabbitmqSection `json:"rabbitmq,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// NotificationsBusInstance - the name of RabbitMQ Cluster CR to select a Messaging
+	// Bus Service instance used by all services that produce or consume notifications.
+	// Avoid colocating it with RabbitMQ services used for PRC.
+	// That instance will be pushed down for services, unless overriden in templates.
+	NotificationsBusInstance *string `json:"notificationsBusInstance,omitempty"`
```
fmount, line 135: Tested locally via ` OPENSTACK_IMG=quay.io/fpantano/openstack-operator-index:v0.3.3`.
If the field does not exist in the CR (is not specified), the CR is not changed, and this won't trigger any reconciliation loop in a minor update scenario. For this reason, having a `*string` ensures it is fully backward compatible with the previous versions.
Like we did for `Topologies`, if this field exists as an override, the top-level is ignored, otherwise it will be inherited by the top-level CR.

fmount:
Note that I found #1424 that can be considered a follow up of this patch, where a dedicated rabbitmq instance is introduced.

apis/core/v1beta1/openstackcontrolplane_types.go
gibizer, line 135: OK so the usual considerations:
* if this is nil then it means nothing to propagate to the service level values. If the service level value is also nil then it means notifications are disabled. And this is our default. 
* if this is set to the name of a rabbitmqcluster CR and the service level value is nil, then the rabbitmqcluster name is propagated to that service. This should be our normal way of enabling notififcations for ceilometer across the whole cluster.
* if this is set to the name of a rabbitmqcluster CR and the service level value is also set to a different rabbitmqcluster CR name then the sevice level value is kept for that service. This is the complicated case where ceilometer is not needed or only partially needed, and for some reason this specific service needs to use an independent rabbitmqcluster.

gibizer: We need to check if we need both "" and nil value here to mean different things. If not the we can either opt to use "" only, or define that "" and means the same for this field.

pkg/openstack/cinder.go
```diff
@@ -130,6 +130,12 @@ func ReconcileCinder(ctx context.Context, instance *corev1beta1.OpenStackControl
 		instance.Spec.Cinder.Template.TopologyRef = instance.Spec.TopologyRef
 	}
 
+	// When no NotificationsBusInstance is referenced in the subCR (override)
+	// try to inject the top-level one if defined
+	if instance.Spec.Cinder.Template.NotificationsBusInstance == nil {
+		instance.Spec.Cinder.Template.NotificationsBusInstance = instance.Spec.NotificationsBusInstance
```
gibizer, line 136: this inherits the "" value to the service level. We need to see if that is what we want

apis/bases/core.openstack.org_openstackcontrolplanes.yaml
gibizer: +1 for consistency

gibizer (commented):
Would it be possible to include the propagation to the nova level field too? I hope it is not too big of a complexity increase to take.

pkg/openstack/cinder.go
fmount: If you look at cinder/glance/manila implementation, for example [1], the parameter is the same, and at openstack-operator it realizes only a parameter inheritance, if required, so I don't see any problem w/ set/update/delete for storage CRs (basically it's the same thing we did for topologies).

[1] https://github.com/openstack-k8s-operators/manila-operator/blob/main/api/v1beta1/manila_types.go#L135

apis/core/v1beta1/openstackcontrolplane_types.go
fmount: > We need to check if we need both "" and nil value here to mean different things. If not the we can either opt to use "" only, or define that "" and means the same for this field.

I think checking if that what the user sets is not an empty string is something we can catch from webhooks, so we can cover updates of the same field (while kubebuilder would help us only at CR creation time) and we can decide to default it to nil. However, I think that raising an error (usual webhook flow) would be better to have a human operator fix the input or remove the parameter entirely.

apis/core/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -1122,3 +1130,24 @@ func (r *OpenStackControlPlane) ValidateTopology(basePath *field.Path) *field.Er
 	}
 	return nil
 }
+
+func (r *OpenStackControlPlane) ValidateNotificationsBusInstance(basePath *field.Path) *field.Error {
```
fmount, line 1134: @gibizer this should help to clarify the question about empty string or malformed input mentioned in another comment.
I can add envTests to validate this function (I'll do in my next iteration)

fmount:
> Would it be possible to include the propagation to the nova level field too? I hope it is not too big of a complexity increase to take.

Of course, done. Added nova propagation as well.

pkg/openstack/glance.go
```diff
@@ -74,6 +74,12 @@ func ReconcileGlance(ctx context.Context, instance *corev1beta1.OpenStackControl
 		instance.Spec.Glance.Template.TopologyRef = instance.Spec.TopologyRef
 	}
 
+	// When no NotificationsBusInstance is referenced in the subCR (override)
+	// try to inject the top-level one if defined
+	if instance.Spec.Glance.Template.NotificationBusInstance == nil {
```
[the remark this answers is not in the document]
fmount: Thank you, but let's hold that patch until we're ready to switch. As mentioned, this is out of scope for now because that small change does not justify getting us in the process of creating a conversion webhook. It will be part of a fix later in time.

fmount:
> Watcher section also supports NotificationsBusInstance. If merging this, It should be also covered for consistency.
> 
> Also, neutron supports it, although I'm not sure if there is any reason to skip it in this patch.

Interesting, thanks @amoralej, I wasn't aware about the current status. I started with storage but I was going to extend this patch as needed. Let me check the current status and add both neutron and watcher!

auniyal61:
how existing flow works for nova and neutron is
if we take example of nova - right now  user will create new rabbit, wire it with nova spec via `notificationsBusInstance`. doing so nova operator recognize new rabbit and creates a new transporturl and update confs so its aware of where to send notifications.

then consumer can listen/read on new rabbit.

---
can you please explain how this will be used from user perspective.

So I think this will works as single top-level switch to enable/disable external notifications from respected  service operator.
user will create `new-rabbit` wire it to top-level `notificationsBusInstance` and for all supported services/operators, individual transporturl will be created and they will start publishing external notifications to single dedicated new-rabbit  

is above correct ?

fmount:
> how existing flow works for nova and neutron is if we take example of nova - right now user will create new rabbit, wire it with nova spec via `notificationsBusInstance`. doing so nova operator recognize new rabbit and creates a new transporturl and update confs so its aware of where to send notifications.
> 
> then consumer can listen/read on new rabbit.
> 
> can you please explain how this will be used from user perspective.
> 
> So I think this will works as single top-level switch to enable/disable external notifications from respected service operator. user will create `new-rabbit` wire it to top-level `notificationsBusInstance` and for all supported services/operators, individual transporturl will be created and they will start publishing external notifications to single dedicated new-rabbit
> 
> is above correct ?

Correct. It perfectly fine to create a dedicated rabbitmq instance and make this parameter point to the new rabbitmq instance name. By doing this all the underlying services will point to it and emit notifications. From a service perspective this mechanism is not different from the regular rabbimq/transportURL request that is already implemented as a pattern across the board. This part is clearly handled in each service (with a dedicated `NotificationBusInstanceReady` condition [1] that is evaluated by each service).
What this patch does is:
- prevent adding a rabbitmq name that refers to an instance that doesn't exist
- propagate the top-level field to the underlying components if it is not present as an ovveride (which is yet another pattern we use to follow for most of the service field)

Any processing logic already exists in the service operators to properly reconcile and configure the services.

[1] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/condition/conditions.go#L72

amoralej:
I've tested locally this and works as intended. webhook validates properly, valid values are propagated correctly and service level parameters overrides it when set.

My only suggestion would be getting it covered by a kuttl test.

You plan to also add some coverage in deployment edpm tests?

bogdando:
lgtm, but mind adding unit tests from the earlier linked implementation patch?

amoralej:
> > I've tested locally this and works as intended. webhook validates properly, valid values are propagated correctly and service level parameters overrides it when set.
> > My only suggestion would be getting it covered by a kuttl test.
> > You plan to also add some coverage in deployment edpm tests?
> 
> Hi @amoralej @bogdando , I assume that each operator might have envTest coverage from the point where the parameter is propagated to the service (see [1] for example, but it's similar for all storage operators) - so I think we might do not need dedicated `kuttl` tests. However I agree that dedicated `envTests` are useful, and I'm going to add them. Because I see a lot of refactoring that I need to do to depuplicate some code, I prefer to create a follow up patch that is focused on enhancing envTests. How do you see it? I can have an additional commit on top of this one, but it might result harder to review, while we could land this patch and I can create a dedicated PR for envTests. I'm saying this because envTests need to:
> 
>     1. dedup the `Entry` map for each service to not be Topology specific, but to cover this use case
> 
>     2. enable nova, which is not enabled in Topology tests
> 
>     3. eventually create a ctlplane with an additional rabbitmq to test overrides
> 
> 
> All the above might deserve a dedicated patch.
> 

I'd be inclined to do a new commit in the same PR. Testing the new feature with envtest before merging it wins to simpler review in this case for me.

BTW, we have deployment tests with notification enabled in watcher-operator CI, we will move to using top-level notificationsBusInstance when this is merged.

> [1] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/test/functional/cinder_controller_test.go#L1243-L1338 [2] #1424

fmount:
> > > > I've tested locally this and works as intended. webhook validates properly, valid values are propagated correctly and service level parameters overrides it when set.
> > > > My only suggestion would be getting it covered by a kuttl test.
> > > > You plan to also add some coverage in deployment edpm tests?
> > > 
> > > 
> > > Hi @amoralej @bogdando , I assume that each operator might have envTest coverage from the point where the parameter is propagated to the service (see [1] for example, but it's similar for all storage operators) - so I think we might do not need dedicated `kuttl` tests. However I agree that dedicated `envTests` are useful, and I'm going to add them. Because I see a lot of refactoring that I need to do to depuplicate some code, I prefer to create a follow up patch that is focused on enhancing envTests. How do you see it? I can have an additional commit on top of this one, but it might result harder to review, while we could land this patch and I can create a dedicated PR for envTests. I'm saying this because envTests need to:
> > > ```
> > > 1. dedup the `Entry` map for each service to not be Topology specific, but to cover this use case
> > > 
> > > 2. enable nova, which is not enabled in Topology tests
> > > 
> > > 3. eventually create a ctlplane with an additional rabbitmq to test overrides
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > All the above might deserve a dedicated patch.
> > 
> > 
> > I'd be inclined to do a new commit in the same PR. Testing the new feature with envtest before merging it wins to simpler review in this case for me.
> > BTW, we have deployment tests with notification enabled in watcher-operator CI, we will move to using top-level notificationsBusInstance when this is merged.
> > > [1] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/test/functional/cinder_controller_test.go#L1243-L1338 [2] #1424
> 
> Agreed, I will add that in my next iteration! Thanks @amoralej

Done

stuggi:
FYI, there is an issue with the adoption test, which make it to fail.

PR #1616, opened 2025-09-25
Remove omitempty for baremetalSetTemplate

rabi:
This seems to break after we removed omitempty where there are defaults in https://github.com/openstack-k8s-operators/openstack-baremetal-operator/pull/324.


rabi:
recheck RETRY_LIMIT

PR #1621, opened 2025-09-25
Update catalog fixes

dprince:
OLM update catalog automation fixes
    
    This workflow isn't always triggered so allow workflow_dispatch
    to call manually
    
    Also updated to correctly detect and set branch name


abays (approved):
/lgtm

PR #1622, opened 2025-09-26
Convert baremetalSetTemplate field to pointer

rabi:
As this field is optional we should make it a pointer which ensures proper optional validation i.e  struct is optional, but if it is provided, some of its internal fields can be required.

Depends-On: https://github.com/openstack-k8s-operators/openstack-baremetal-operator/pull/328


abays (approved):
/lgtm

PR #1633, opened 2025-10-02
WIP Add CSV entries for new graphical console images

steveb:


steveb:
The last pushed change was a rebase

PR #1584, opened 2025-08-29
Add pullspecs replace script for CSV image updates

dprince:
Introduces automation script to update RELATED_IMAGE_ environment variables in ClusterServiceVersion files with current environment values, enabling dynamic container image reference management.

To use:

 make bundle
 bash hack/pullspecs_replaceatron.sh

NOTE: for test purposes you can run hack/export_related_images.sh to set most related images except for OPERATOR images which will be needed to set manually.

PR #1680, opened 2025-11-03
extend pullspecs_replaceatron logic for downstream

yazug:
Expand REGEX for which images to pin/replace out to include KUBE_RBAC_PROXY
also use logic so that if a VAR is empty but defined it replaces the pin with empty instead of complaining

Followup to https://github.com/openstack-k8s-operators/openstack-operator/pull/1584

### DIFF
--- a/hack/pullspecs_replaceatron.sh
+++ b/hack/pullspecs_replaceatron.sh
@@ -19,7 +19,7 @@ cp "$CSV_FILE" "${CSV_FILE}.backup"
 echo "Updating RELATED_IMAGE_ environment variables in $CSV_FILE..."
 
 # Extract all RELATED_IMAGE_ env var names from the CSV file
-RELATED_IMAGE_VARS=$(grep -o 'RELATED_IMAGE_[A-Z_]*' "$CSV_FILE" | sort -u)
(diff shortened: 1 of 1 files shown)


hack/pullspecs_replaceatron.sh
```diff
@@ -19,15 +19,15 @@ cp "$CSV_FILE" "${CSV_FILE}.backup"
 echo "Updating RELATED_IMAGE_ environment variables in $CSV_FILE..."
 
 # Extract all RELATED_IMAGE_ env var names from the CSV file
-RELATED_IMAGE_VARS=$(grep -o 'RELATED_IMAGE_[A-Z_]*' "$CSV_FILE" | sort -u)
+RELATED_IMAGE_VARS=$(grep -o -e 'RELATED_IMAGE_[A-Z0-9_]*' -e '[A-Z0-9_]*_IMAGE_URL_DEFAULT' -e 'KUBE_RBAC_PROXY' "$CSV_FILE" | sort -u)
```
stuggi, line 22: nit: we no longer depend on kube rbac proxy, but it does not hurt to have it

stuggi (approved):
/lgtm

PR #1718, opened 2025-12-03
Add support for configuring environment variables for operators

lmiccini:
This change allows users to specify custom environment variables for operator containers through the operatorOverrides field in the OpenStack CR.

Changes:
- Add Env field to ContainerSpec in the API types
- Implement mergeEnvVars() to merge custom environment variables with defaults
- Update SetOverrides() to apply environment variable overrides
- Enhance operator deployment templates to support both Value and ValueFrom for environment variables (supporting secrets, configmaps, and field refs)

Usage example:
```
    apiVersion: operator.openstack.org/v1beta1
    kind: OpenStack
    metadata:
      name: openstack
    spec:
      operatorOverrides:
      - name: infra
        controllerManager:
          env:
          - name: SOME_ENV_VAR
            value: "some-value"
```

Co-Authored-By: Claude <noreply@anthropic.com>


config/operator/deployment/kustomization.yaml
```diff
@@ -14,3 +14,28 @@ images:
...
+  target:
+    kind: Deployment
+    name: openstack-operator-controller-operator
+    namespace: system
+- patch: '[{"op": "replace", "path": "/spec/template/spec/containers/0/env/1", "value":
+    {"name": "OPERATOR_IMAGE_URL", "value": "quay.io/openstack-k8s-operators/openstack-operator:latest"}}]'
+  target:
+    kind: Deployment
+    name: openstack-operator-controller-operator
+    namespace: system
```
stuggi: should probably not be part of the PR?

lmiccini:
> @lmiccini as a generate feature what you have here is good. and we could add it.
> 
> I'm not sure about the use case mentioned in the PR for rabbitmq/cluster-operator. If the intent is to allow multiple rabbit operators within a cluster that would be dangerous due to there being only 1 CRD per OCP cluster (the CRDs could be different version, etc).

indeed, removed references to the rabbitmq cluster operator and made this generic.

PR #1743, opened 2025-12-16
Prune certain services from three KUTTL tests

abays:
First pass at trying to prune some potentially unneeded services from the `OpenStackControlPlane` during KUTTL tests.  The approach taken here showed a 10-15% performance boost.


abays:
This did not make a significant difference, so closing.

PR #1864, opened 2026-03-26
Fix MinorUpdateOVNDataplane race when OVN image unchanged between versions

stuggi:
Commit https://github.com/openstack-k8s-operators/openstack-operator/commit/63fd70544f0407d2565e503fbb6b49b2b8223f46 removed the nodeset.IsReady() check from
DataplaneNodesetsOVNControllerImagesMatch to fix the minor update
workflow getting stuck when unrelated deployments were running. That
check was too strict — it blocked when any deployment was in progress,
even if the OVN update had already completed.

However, the remaining pure image comparison is too loose. When the
OVN controller image does not change between two OpenStack versions,
the nodeset already has the matching image from the previous update.
The OpenStackVersion controller then sets MinorUpdateOVNDataplane=True
immediately, before the edpm-ovn-update deployment finishes. This
causes the subsequent minor update steps (controlplane update, edpm
services update) to proceed while the OVN dataplane deployment is
still running — resulting in both dataplane deployments running
concurrently.

Fix this with two mechanisms:

1. IsDataplaneDeploymentRunningForServiceType: a new generic function
   that lists all in-progress OpenStackDataPlaneDeployment resources,
   resolves which services each deploys (from ServicesOverride or the
   nodeset's service list), and checks the service's EDPMServiceType
   to identify deployments of a given type (e.g. "ovn").
   EDPMServiceType is a fixed identifier on the service spec,
   independent of the service or deployment resource name.

2. Saved condition state tracking: when the OVN image is unchanged
   between the deployed and target versions, image comparison alone
   cannot distinguish "already updated" from "not yet updated." To
   handle this, we use the saved condition state (captured before
   Init resets conditions each reconciliation) to track whether a
   running OVN deployment has been observed during this update cycle:
   - Running OVN deployment seen → set condition False(RequestedReason)
   - No running deployment + previous condition was RequestedReason →
     deployment completed, proceed
   - No running deployment + previous condition was NOT
     RequestedReason → deployment not created yet, keep waiting

   The OpenStackVersion controller watches nodesets, so when a
   deployment starts (nodeset becomes not-Ready) and completes
   (nodeset becomes Ready), reconciliation is triggered, ensuring
   we observe the running deployment at least once.

When the OVN image differs between versions, the existing image match
check is sufficient — the nodeset's ContainerImages are only updated
on successful deployment completion, so a match proves the deployment
ran.

### DIFF
--- a/internal/controller/core/openstackversion_controller.go
+++ b/internal/controller/core/openstackversion_controller.go
@@ -291,6 +291,75 @@ func (r *OpenStackVersionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 				Log.Info("Waiting on OVN Dataplane updates to complete")
 				return ctrl.Result{}, nil
 			}
+
+			// When the OVN controller image is the same between the deployed
+			// version and the target version, the image comparison above always
+			// passes because the nodeset already has the matching image from
+			// the previous update. In this case we need additional checks to
+			// confirm the OVN dataplane deployment for this update cycle has
+			// actually completed.
+			//
+			// We use the saved condition state (from before Init reset) to
+			// track whether we have observed a running OVN deployment during
+			// this update cycle:
+			// - If we see a running OVN deployment now: set condition False
+			//   (RequestedReason) to record that we observed one
+			// - If no running OVN deployment AND the previous condition was
+			//   False/RequestedReason: the deployment we saw previously has
+			//   completed → proceed (fall through to set True)
+			// - If no running OVN deployment AND the previous condition was
+			//   NOT False/RequestedReason (e.g. still Unknown from Init):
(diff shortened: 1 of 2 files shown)


abays (approved):
/lgtm

rabi (commented):
This is getting really messy :smile:  We do create nodset/deployment to validate the CRs[1] after update. Probably we need to move the edpm_deploy_cleanup before we check for openstackctlplane ready in L195.

https://github.com/openshift/release/blob/main/ci-operator/step-registry/openstack-k8s-operators/kuttl/openstack-k8s-operators-kuttl-commands.sh#L173

bshephar (commented):
Drive by comment

rabi (approved):

PR #1868, opened 2026-03-31
[b/r] Add OpenStackBackupConfig controller and backup/restore labeling

stuggi:
- Add OpenStackBackupConfig CRD and controller that watches CRD instances across operators and labels namespace resources (Secrets, ConfigMaps, NADs) with backup.openstack.org labels for backup/restore integration
- Wire backup/restore labeling into the ControlPlane controller: CA cert secrets get backup labels via SecretTemplate, internal service cert requests get restore=false (via lib-common default), and ReconcileBackupConfig creates/updates the BackupConfig CR with spec defaults

Commit 1: [b/r] Add OpenStackBackupConfig controller                                                                                                                                       

Introduces the backup.openstack.org/v1beta1 API group with the OpenStackBackupConfig CRD. The controller:                                                                                  
- Discovers CRD instances by reading backup.openstack.org/restore and backup.openstack.org/restore-order labels from CRD schemas  (only on start, not on each reconcile and creates a cache) and applies them to all instances. This allows to have a dynamic approach where new CRDs just require the labels and the controller the rbac perms.
- Labels Secrets, ConfigMaps, and NADs in the namespace with configurable restore ordering
- Supports per-resource annotation overrides (backup.openstack.org/restore, backup.openstack.org/restore-order) to customize or exclude individual resources
- Includes envtest coverage

Commit 2: [b/r] Add backup/restore labels to ControlPlane controller                                                                                                                       

Integrates backup/restore into the existing ControlPlane reconciliation:                                                                                                                   
- ReconcileBackupConfig in internal/openstack/backup.go creates the OpenStackBackupConfig CR with spec defaults via CreateOrPatch
- CA cert secrets labeled at creation time in ca.go via SecretTemplate
- Labels custom cert-manager Issuers
- Internal service cert requests labeled with restore=false via lib-common (regenerated by cert-manager on restore)
- CRD label additions for backup.openstack.org/restore and backup.openstack.org/restore-order on ControlPlane, Version, DataPlaneNodeSet, and DataPlaneService types

Jira: [OSPRH-22912](https://redhat.atlassian.net/browse/OSPRH-22912)
Jira: [OSPRH-22913](https://redhat.atlassian.net/browse/OSPRH-22913)
Jira: [OSPRH-26645](https://redhat.atlassian.net/browse/OSPRH-26645)

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/680
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/684
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/685


api/backup/v1beta1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
+/*
+Copyright 2022.
```
abays: One immediate nit that could probably be fixed across a few files here: just make sure copyright year is 2026

internal/openstack/backup.go
```diff
@@ -0,0 +1,94 @@
...
+	op, err := controllerutil.CreateOrPatch(ctx, helper.GetClient(), backupConfig, func() error {
+		// Note: We do NOT set ownerReference here. OpenStackBackupConfig is a configuration
+		// resource that users may customize. It should persist even if the ControlPlane is
+		// deleted, and should be backed up/restored with user customizations intact.
+
+		// Set spec defaults. CRD schema defaults only apply when fields are
+		// absent from the request, but Go serializes zero-value structs as
+		// empty objects which bypasses CRD defaulting.
+		if backupConfig.Spec.DefaultRestoreOrder == "" {
+			backupConfig.Spec.DefaultRestoreOrder = "10"
```
abays: Maybe declare a `const` for this `10` somewhere and use that here?

stuggi: good catch, I already have those const in lib-common. missed to update, thanks

internal/controller/backup/openstackbackupconfig_controller.go
```diff
@@ -0,0 +1,632 @@
...
+	return r.labelResourceItems(ctx, log, items, instance.Spec.NetworkAttachmentDefinitions, defaultLabels)
+}
+
+// labelCRInstances labels CR instances based on CRD backup-restore labels
+// This labels CRs like OpenStackControlPlane, OpenStackVersion, NetConfig, etc.
+// based on their CRD's backup/restore configuration.
+func (r *OpenStackBackupConfigReconciler) labelCRInstances(ctx context.Context, log logr.Logger, instance *backupv1beta1.OpenStackBackupConfig) (int, error) {
+	// Fallback: build cache if SetupWithManager failed to populate it.
+	// Note: watches for CRD instance types are only registered at setup time,
+	// so CR instance changes won't trigger reconciliation in this case.
```
abays, line 240: And we don't need to watch for CRD changes during runtime, because new CRDs would only be added during an update, in which case a new OpenStack operator is created (which re-inits the cache).  Is that true?

stuggi: yes correct. with an update we get a new openstack-op image which has the new CRDs and the operator/controllers get restarted with the new version

config/samples/backup_v1beta1_openstackbackupconfig.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: backup.openstack.org/v1beta1
+kind: OpenStackBackupConfig
+metadata:
+  labels:
+    app.kubernetes.io/name: openstack-operator
+    app.kubernetes.io/managed-by: kustomize
+  name: openstackbackupconfig-sample
+spec:
+  # Target namespace to watch for resources
+  targetNamespace: openstack
```
abays:
```suggestion
```

abays: It seems this field doesn't exist.  I'm guessing you just use the `metadata.Namespace` instead.

api/backup/v1beta1/openstackbackupconfig_types.go
abays, line 1: Is there only supposed to be one `OpenStackBackupConfig` per namespace?  If so, should we add a validation webhook to block creating one if one already exists?  Also if so, maybe in the case that the user creates the `OpenStackBackupConfig` first, the `OpenStackControlPlane` webhook should block creating an `OpenStackControlPlane` that doesn't have the same name as `OpenStackBackupConfig` (and I guess we would need to take `OpenStackVersion` into consideration too, since its name also has to be the same)?

config/samples/backup_v1beta1_openstackbackupconfig.yaml
stuggi: good catch, I initially added the targetNamespace, but then switched to metadata.Namespace instead. seems I missed to remove that.

api/backup/v1beta1/openstackbackupconfig_types.go
stuggi: yes, it only makes sense to have one OpenStackBackupConfig per namespace, similar to the OpenStackVersion. I am not sure if we have to restrict it to be the same name, like we do it with the openstackversion. We could probably just check if there is already OpenStackBackupConfig and if so, we do not create the default one? We'd not delete them as a user might have done any customization, espacially if it was the auto generated. I think we can add that webhook/check in a follow up?

abays: Sure, I think addressing this in a follow-up is fine.

stuggi: I have created https://redhat.atlassian.net/browse/OSPRH-29193 for this

abays (approved):
/lgtm

PR #1873, opened 2026-04-07
Add OLM upgrade support to Makefile

stuggi:
Add REPLACES variable to inject `replaces` field into the CSV for OLM upgrade paths. Add PREV_BUNDLE_IMG and CATALOG_BUNDLE_IMGS variables to support incremental and from-scratch catalog builds with full upgrade chains. Fix kustomization.yaml patch accumulation by clearing old patches before regenerating.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,10 @@ ifeq ($(USE_IMAGE_DIGESTS), true)
 	BUNDLE_GEN_FLAGS += --use-image-digests
 endif
 
+# REPLACES is the previous version that this version replaces (for OLM upgrades)
+# Example: make bundle REPLACES=openstack-operator.v0.6.0 VERSION=0.6.1
+REPLACES ?=
+
 # Set the Operator SDK version to use. By default, what is installed on the system is used.
 # This is useful for CI or a project to utilize a specific version of the operator-sdk toolkit.
 OPERATOR_SDK_VERSION ?= v1.41.1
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -447,12 +457,30 @@ SHELL_EXPORT = $(foreach v,$(MAKE_ENV),$(v)='$($(v))')
...
+
 # The image tag given to the resulting catalog image (e.g. make catalog-build CATALOG_IMG=example.com/operator-catalog:v0.2.0).
 CATALOG_IMG ?= $(IMAGE_TAG_BASE)-index:v$(VERSION)
 
 # Set CATALOG_BASE_IMG to an existing catalog image tag to add $BUNDLE_IMGS to that image.
+# When using a base catalog, only add the NEW bundle (base already has everything else)
+# Alternatively, set CATALOG_BUNDLE_IMGS directly to specify all bundles (for building from scratch without a base)
 ifneq ($(origin CATALOG_BASE_IMG), undefined)
 FROM_INDEX_OPT := --from-index $(CATALOG_BASE_IMG)
+CATALOG_BUNDLE_IMGS ?= $(BUNDLE_IMG)
```
abays, line 477: Does it matter here that this will exclude images from `pin-bundle-images.sh`?

stuggi: I am not sure if I fully got your question. If we have CATALOG_BASE_IMG, the base catalog already contains all the dependency bundles, I think we only need to add the new one `BUNDLE_IMG` (singular).
in L482, when we have no base, we use `BUNDLE_IMGS` (plural) which includes all dependencies.

But maybe I missed a situation?

stuggi: as discussed on slack, you catched a valid point, thanks!! changed and rebased

abays (approved):
/lgtm

ciecierski:
@sathlan Just fyi that there is being added mechanism for OLM update testing

PR #1881, opened 2026-04-12
Remove go.work before container build

stuggi:
The go.work file is generated by the CI environment which may have a higher Go version than the build image (go-toolset). This causes build failures when go.work references a Go version that the build image does not provide. Since the Dockerfile builds a single module, go.work is not needed and can be safely removed before building.


stuggi:
recheck

fmount (approved):
/lgtm

PR #1883, opened 2026-04-13
[18.0-fr5] Remove go.work before container build

stuggi:
The go.work file is generated by the CI environment which may have a higher Go version than the build image (go-toolset). This causes build failures when go.work references a Go version that the build image does not provide. Since the Dockerfile builds a single module, go.work is not needed and can be safely removed before building.

(cherry picked from commit 540876aa6b3635fd0719bb7b2b39540d752df8ac)


fmount (approved):
/lgtm

stuggi:
build went fine. I'll override failing jobs on later test errors as they are not related to this change in the Dockerfile.

PR #1885, opened 2026-04-13
Pin RabbitMQ image to 3.9 on 18.0-fr5

lmiccini:
The current-podified tag now resolves to RabbitMQ 4.2, which
requires the classic_mirrored_queue_version feature flag to be
enabled before upgrading. This breaks the minor update path from
fr2 to fr5 because the old mnesia data is incompatible with the
new RabbitMQ version.
    
Pin the image to quay.io/openstack-k8s-operators/rabbitmq:3.9
to keep the same RabbitMQ version across the update.


stuggi (approved):
/lgtm

lmiccini:
Recheck

lmiccini:
recheck

lmiccini:
recheck

lmiccini:
recheck

lmiccini:
recheck

PR #1888, opened 2026-04-14
Fix webhook config merge to match by name instead of index

lmiccini:
- `MergeWebhookConfigurationForUpdate` was copying the entire `clientConfig` from current to updated webhooks by **array index**. When the webhook arrays have different ordering (Kubernetes sorts alphabetically by name, but the template YAML has a different order), each webhook gets the `clientConfig` from the wrong entry, scrambling all the service paths.
- This caused admission requests to be routed to the wrong handler. For example, `mrabbitmq-v1beta1.kb.io` would get the path for `/mutate-network-openstack-org-v1beta1-reservation`, resulting in: `unable to decode rabbitmq.openstack.org/v1beta1, Kind=RabbitMq into *v1beta1.Reservation`
- Fix by matching webhooks by `name` field and only copying the `caBundle` (injected by cert-manager) rather than the entire `clientConfig`, preserving the correct service path from the updated template.

Jira: https://redhat.atlassian.net/browse/OSPRH-29026

### DIFF
--- a/internal/operator/bindata/merge.go
+++ b/internal/operator/bindata/merge.go
@@ -55,7 +55,10 @@ func MergeObjectForUpdate(current, updated *uns.Unstructured) error {
 	return nil
 }
 
-// MergeWebhookConfigurationForUpdate merges certs and clientConfig from the current object
+// MergeWebhookConfigurationForUpdate merges the caBundle from the current
+// webhook configuration into the updated one, matching webhooks by name.
+// This preserves the CA certificate injected by cert-manager while keeping
+// the service path and other clientConfig fields from the updated template.
 func MergeWebhookConfigurationForUpdate(current, updated *uns.Unstructured) error {
 	gvk := updated.GroupVersionKind()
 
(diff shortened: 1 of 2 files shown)


internal/operator/bindata/merge.go
```diff
@@ -65,17 +68,44 @@ func MergeWebhookConfigurationForUpdate(current, updated *uns.Unstructured) erro
...
+			cc, ok := cwMap["clientConfig"].(map[string]interface{})
+			if !ok {
+				continue
 			}
+			if caBundle, exists := cc["caBundle"]; exists {
+				caBundleByName[name] = caBundle
+			}
+		}
 
+		for _, webhook := range updated.Object["webhooks"].([]interface{}) {
```
abays: While it should never really happen, should we guard against a panic on the conversion here, like we do on lines 66-69?  I guess it would also be more consistent.

abays (approved):
/lgtm

fmount:
looks gook! thanks @lmiccini!!

lmiccini:
recheck

PR #1904, opened 2026-04-29
[b/r] Add singleton webhook for OpenStackBackupConfig

stuggi:
Add a validating webhook that enforces only one OpenStackBackupConfig instance per namespace. On create, the webhook lists existing instances and rejects the request if one already exists.

Update ReconcileBackupConfig to check for a pre-existing BackupConfig before creating the default. If a user has pre-created one, the controller uses it instead of creating a new one.

Jira: [OSPRH-22912](https://redhat.atlassian.net/browse/OSPRH-22912)
Jira: [OSPRH-22913](https://redhat.atlassian.net/browse/OSPRH-22913)
Jira: [OSPRH-29193](https://redhat.atlassian.net/browse/OSPRH-29193)

### DIFF
--- a/internal/webhook/backup/v1beta1/openstackbackupconfig_webhook.go
+++ b/internal/webhook/backup/v1beta1/openstackbackupconfig_webhook.go
@@ -1,0 +1,119 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package v1beta1 contains webhooks for backup API resources.
+package v1beta1
+
+import (
+	"context"
+	"fmt"
+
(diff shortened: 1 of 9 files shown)


internal/webhook/backup/v1beta1/openstackbackupconfig_webhook.go
```diff
@@ -0,0 +1,119 @@
...
+	configList, err := backupv1beta1.GetOpenStackBackupConfigs(ctx, backupConfig.Namespace, backupConfigWebhookClient)
+	if err != nil {
+		return nil, apierrors.NewForbidden(
+			schema.GroupResource{
+				Group:    backupv1beta1.GroupVersion.WithKind("OpenStackBackupConfig").Group,
+				Resource: backupv1beta1.GroupVersion.WithKind("OpenStackBackupConfig").Kind,
+			}, backupConfig.GetName(), &field.Error{
+				Type:   field.ErrorTypeForbidden,
+				Detail: err.Error(),
+			},
```
abays, line 80: Not really important, but is there some other type of general error that would make more sense here instead of `Forbidden`?  Or is this just the pattern we've used in webhook error handling across all operators?

stuggi: its the same as we also use in other places, like the osversion https://github.com/openstack-k8s-operators/openstack-operator/blob/main/api/core/v1beta1/openstackversion_webhook.go#L97 . maybe NewConflict is better. maybe we follow up to make them all the same.

abays (approved):
/lgtm

PR #1906, opened 2026-04-29
Generate OVN RBAC PKI

slawqo:
This patch adds generation of the rootca-ovn-rbac issuer which is used by the ovn-operator to sign per-node ovn-controller RBAC certificates. This CA is intentionally not added to the combined CA bundle as it is only used between the SB database and ovn-controller nodes.

When TLS is enabled, the reconciler passes the RBAC CA cert secret name to the SB DB cluster and the RBAC issuer name to OVNController so the ovn-operator can create cert-manager Certificate resources and verify client connections.

Related: #[OSPRH-1921](https://redhat.atlassian.net/browse/OSPRH-1921)

Related: #[OSPRH-1922](https://redhat.atlassian.net/browse/OSPRH-1922)

### DIFF
--- a/api/bases/dataplane.openstack.org_openstackdataplaneservices.yaml
+++ b/api/bases/dataplane.openstack.org_openstackdataplaneservices.yaml
@@ -151,13 +151,22 @@ spec:
                     OpenstackDataPlaneServiceCert defines the property of a TLS cert issued for
                     a dataplane service
                   properties:
+                    commonName:
+                      description: |-
+                        CommonName overrides how the certificate Common Name is derived.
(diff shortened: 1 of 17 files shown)


stuggi:
@slawqo as mentioned in the ovn pr. wondering if we need to use a new CA as we already have a dedicated CA for ovn. for testing, you should add a replace line in the go.mod files to point to your fork.

slawqo:
> @slawqo as mentioned in the ovn pr. wondering if we need to use a new CA as we already have a dedicated CA for ovn. for testing, you should add a replace line in the go.mod files to point to your fork.

yes, I made changes to do that

PR #1907, opened 2026-04-29
Fix NodeSelector nil-vs-pointer reconcile loop in all service reconcilers

lmiccini:
When the OSCP has no nodeSelector defined, instance.Spec.NodeSelector is a nil map. Taking its address (&instance.Spec.NodeSelector) creates a non-nil pointer to a nil map, which is structurally different from a nil pointer. controllerutil.CreateOrPatch detects this as a diff via reflect.DeepEqual and sends an update patch on every reconcile, bumping the sub-CR's Generation. The ObservedGeneration readiness check then reports the sub-CR as "in progress", triggering another OSCP reconcile and creating an infinite loop (~1 update/second).

Guard all 25 NodeSelector inheritance assignments with len(instance.Spec.NodeSelector) > 0 so the assignment is skipped when the OSCP nodeSelector is nil or empty, avoiding the spurious diff.


nhicher:
recheck

lmiccini:
recheck

PR #1916, opened 2026-05-06
Only init conditions for enabled services in InitConditions

stuggi:
Previously, InitConditions() added Unknown conditions for ALL services regardless of whether they were enabled. When the controller short- circuited before reaching a disabled service's reconciler (e.g., during minor update gating), the disabled service's Unknown condition persisted and was mirrored to the Ready condition, producing misleading status like "Octavia not started" even when Octavia was disabled.

Only create conditions for services that are actually enabled. Always-on conditions (CAs, Version, Infrastructure, BackupConfig, Client, Ready) remain unconditional. The service reconcilers still handle the enabled-to-disabled transition by removing their conditions.


abays (approved):
/lgtm

PR #1918, opened 2026-05-12
Add dataplane service dependency graph and fallback ordering

rabi:
- dependsOn on services; deployment plan is a topological sort into
  levels that run concurrently
- services without dependsOn depend on their list predecessor, so
  existing deployments keep serial order
- unresolvable dependsOn refs are skipped with a log message
- build per-service AnsibleEE spec only at job start; split Deploy
  into startLevel/isLevelReady
- useParallelExecution flag on deployment CR (default false); when
  not set, dep graph is not built, services run sequentially in list
  order and a warning is logged that dependsOn is ignored
- fail fast on missing service CRs: dedupe now returns
  MissingServiceError instead of logging and continuing, so a
  misconfigured service blocks the deployment early with per-service
  and nodeset conditions set, instead of failing at job start

jira: [OSPRH-29432](https://redhat.atlassian.net/browse/OSPRH-29432)


slagle (commented):
I didn't fully review the code, but I would good to move forward with the idea here if the FallbackToListOrder acts as a feature flag so to speak. Is that the case? I think some might just want to opt out completely from the depends on logic and use the old behavior.

rabi:
>  if the FallbackToListOrder acts as a feature flag so to speak. Is that the case?

Thanks for checking. No FallbackToListOrder isn't a feature flag, it was a migration aid, and I've since dropped it entirely in the new patch. No flag is needed because dependsOn is per-service opt-in, transparent from the nodeset/deployment spec pov.
                                                                                                                                                                                                  
     - services with dependsOn: topologically sorted into levels that run concurrently                                                                                                          
     - services without dependsOn: depend on their list predecessor, so existing deployments keep the old serial order                                                                                                
                                                                                                                                                                                                                 
On opting out completely, for custom services one simply don't set dependsOn to get the legacy list-order behavior. For the built-in services, dependsOn ships in the operator-managed service templates, so those follow dependency-based ordering atm and can't be individually overridden. 
 
If you think we still have to add an opt-in feature flag, I can add `useParallelExecution`  (default: false) in the deployment CR and one has to set it to true to leverage parallel execution. But deployment level flag (false) overriding service dependencies is kind of unintuitive, but we can probably give a log warning like "useParallelExecution is false, dependsOn ignored". Let me know.

internal/dataplane/depgraph.go
```diff
@@ -0,0 +1,187 @@
...
+			}
+		} else if i > 0 {
+			addDep(services[i-1])
+		}
+
+		deps[svc] = resolved
+	}
+
+	return deps, nil
+}
```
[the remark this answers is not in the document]
rabi: This is a design decision as  one should be able to run  a service  in `serviceOverride` without having their dependency in the list. We don't normally expect typos in the service definitions, if they are there it's ok to ignore and anyway deployment would fail.

rabi:
`case.go:380: Get "https://api.oko-16.oooci.ccitredhat.com:6443/apis/core.openstack.org/v1beta1/namespaces/openstack-kuttl-tests/openstackcontrolplanes/openstack-collapsed-cell": dial tcp 38.102.83.28:6443: i/o timeout`

slagle:
> If you think we still have to add an opt-in feature flag, I can add `useParallelExecution` (default: false) in the deployment CR and one has to set it to true to leverage parallel execution. But deployment level flag (false) overriding service dependencies is kind of unintuitive, but we can probably give a log warning like "useParallelExecution is false, dependsOn ignored". Let me know.

I do think we need a feature flag given this should also come with user facing docs, and I'm not sure we have that commitment for this release, so I hesitate to just go forward with a default behavior change.

Particularly, not all ansible can be run in parallel against the same node. Certain tasks would conflict (such as package modules or sshd restarts), while we have retries which would probably handle that sort of thing, I think these sorts of behaviors are what we would need to document.

rabi:
> Particularly, not all ansible can be run in parallel against the same node

I mean currently users add (custom) services in the order they want them to be executed in the service list and we honor them if there is no dependency in the service definition, so there won't be a scenario where we would be running some services in parallel when they're not expected to.

Having said that I'll add the flag if that makes us more comfortable.

internal/dataplane/depgraph.go
```diff
@@ -0,0 +1,197 @@
...
+
+	for _, svc := range services {
+		service, err := serviceCache.Get(ctx, helper, svc)
+		if err != nil {
+			return nil, err
+		}
+		serviceObjects[svc] = service
+		serviceType := service.Spec.EDPMServiceType
+		typeToName[serviceType] = svc
+		nameToType[svc] = serviceType
```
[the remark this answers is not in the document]
rabi: This is already defaulted in webhook.

internal/dataplane/depgraph.go
```diff
@@ -0,0 +1,197 @@
...
+				target, err := resolveDependency(dep, svc, nameToType[svc], typeToName, nameToType)
+				if err != nil {
+					return nil, err
+				}
+				if target == "" {
+					helper.GetLogger().Info("dependsOn entry not in nodeset service list, skipping",
+						"service", svc, "dependsOn", dep)
+					continue
+				}
+				addDep(target)
```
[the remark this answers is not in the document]
rabi: This is intended design as we expect a service with dependsOn can still run alone for a deployment without the services it dependsOn in the servicesOverride list of a deployment.

rabi:
recheck

internal/dataplane/depgraph.go
```diff
@@ -0,0 +1,197 @@
...
+			}
+		} else if i > 0 {
+			addDep(services[i-1])
```
slagle, line 106: One constraint of this design (AI assisted) is that if you have a custom services list where all the services could run in parallel, there doesn't seem to be a way to make that happen. The first service would always run in isolation alone first, then the rest would all run in parallel.

rabi: Yeah, it's a deliberate tradeoff I settled on that avoided adding more API surface (more service fields and complexity) for backward compatibility of honoring list order in case of missing dependencies.

slagle (approved):

rabi:
recheck POST_FAILURE

api/dataplane/v1beta1/openstackdataplaneservice_types.go
```diff
@@ -117,6 +117,15 @@ type OpenStackDataPlaneServiceSpec struct {
 	// to manage the service. If not set, will default to the
 	// OpenStackDataPlaneService name.
 	EDPMServiceType string `json:"edpmServiceType,omitempty" yaml:"edpmServiceType,omitempty"`
+
+	// DependsOn lists EDPMServiceType or CR names that must complete before
+	// this service runs; if empty, the list predecessor is used instead.
```
abays, line 122: If the `DependsOn` list empty, what is the "list predecessor" that is being referred to?  Is that list the `OpenStackDataPlaneNodeSet.spect.services` list?

rabi: Yes, if the DependsOn is not in a service spec, then their position in services list would be honored, so we don't break any existing jobs, deployments etc using custom services.

rabi: @abays Are you ok or still have concerns?

abays: I'm good with the mechanism, no worries.  I was just slightly unsure about the comment.  Thank you for clarifying!

abays:
@rabi, could this ordering issue happen with `useParallelExecution: true` and multiple NodeSets?

`DedupeServices` keeps a `DeployOnAllNodeSets` service only in the first NodeSet and removes it from later NodeSet plans. The new built-in configuration makes `ssh-known-hosts` such a global service, while `nova` declares `ssh-known-hosts` in `dependsOn`. For later NodeSets, `BuildServiceLevels` therefore cannot see that dependency and skips it, while the controller continues processing later NodeSets before the first NodeSet’s global Job has completed.

That appears to allow Nova on a later NodeSet to start while the global `ssh-known-hosts` Job is still running. Do you agree this can happen, or is there deployment-wide readiness handling I am missing?

rabi:
> For later NodeSets, `BuildServiceLevels` therefore cannot see that dependency and skips it, while the controller continues processing later NodeSets before the first NodeSet’s global Job has completed.

Very good question. I think it could happen before this change too (nova on different nodeset runs before shh_known_hosts finishes on the nodeset where it decided to run). Actually, nova service does not have any direct dependency on ssh_know_hosts during deployment, but is required post deployment for cold, live migration etc.

We can remove the dependency in service definition of nova if we want, but I decided to keep it for users to know that there is a dependency albeit post deployment.

rabi:
recheck

rabi:
> > For later NodeSets, `BuildServiceLevels` therefore cannot see that dependency and skips it, while the controller continues processing later NodeSets before the first NodeSet’s global Job has completed.
> 
> Very good question. I think it could happen before this change too (nova on different nodeset runs before shh_known_hosts finishes on the nodeset where it decided to run). Actually, nova service does not have any direct dependency on ssh_know_hosts during deployment, but is required post deployment for cold, live migration etc.
> 
> We can remove the dependency in service definition of nova if we want, but I decided to keep it for users to know that there is a dependency albeit post deployment.

Though it's ok for the example `ssh-known-hosts/nova` we discussed above, it's general design gap. I've created a jira https://redhat.atlassian.net/browse/OSPRH-37768 to track it.

oliashish:
LGTM

rabi:
recheck

PR #1922, opened 2026-05-18
Set EDPM service annotation on nova and ceilometer ACs

Deydra71:
Set the `keystone.openstack.org/edpm-service` annotation on nova and ceilometer ApplicationCredential CRs so the keystone-operator AC controller can gate secret rotation and deletion on EDPM NodeSet hash sync. Other services are not EDPM services and can proceed without the NodeSet check.

Depends-On: https://github.com/openstack-k8s-operators/keystone-operator/pull/685

Assisted-by: Claude Opus 4.6 noreply@anthropic.com


Deydra71:
recheck

Deydra71:
recheck

stuggi (approved):
/lgtm

PR #1925, opened 2026-05-20
Use prometheus-podman-exporter and mysqld_exporter images from oko quay

vyzigold:
Now that we have a workflow setup to automatically build prometheus-podman-exporter and mysqld_exporter latest images whenever there is new content in openstack-k8s-operators/prometheus-podman-exporter and openstack-k8s-operators/mysqld_exporter main branches. We can consume those images by default instead of relating on upstream images, which may have slightly different content.

Related PRs:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/1179
https://github.com/openstack-k8s-operators/telemetry-operator/pull/900

Closes: https://redhat.atlassian.net/browse/OSPRH-29306


vyzigold:
"CRD size report". First time seeing this. Interesting check, although I'm not sure it reports correct numbers. I'm pretty sure the default images aren't stored in the oscp CRD, so there shouldn't be a change to the size. And even if they were, it should grow by a few bytes, not shrink by 11kB.

vyzigold:
recheck

jlarriba (approved):

abays (approved):
/lgtm

PR #1927, opened 2026-05-20
Add OpenStackAssistant CRD with MCP server support

dprince:
Introduces a new OpenStackAssistant custom resource
    (assistant.openstack.org/v1beta1) that deploys a managed Goose AI agent
    pod for cluster diagnostics via Lightspeed Stack.

    OpenStackAssistant CRD and controller:
    - New CRD with spec fields for provider type, container image,
      Lightspeed Stack backend configuration, node selectors, and
      additional environment variables
    - GooseConfig supports model selection, recipe ConfigMaps (registered
      as Goose slash commands), hints ConfigMaps (written to .goosehints),
      and MCP server references
    - Controller creates a ServiceAccount, ClusterRole with read-only RBAC
      for cluster diagnostics, ClusterRoleBinding, ConfigMap with Goose
      configuration and entrypoint script, and the assistant Pod
    - Watches referenced Secrets and ConfigMaps; reconciles on changes and
      tracks input hashes to detect drift
    - Defaulting webhook sets the container image from an environment
      variable fallback
    - Condition-based status reporting (ServiceAccount, RBAC, ConfigMap,
      Pod readiness)

    MCP server sidecar support for OpenStackClient:
    - New MCPConfig struct (enabled flag, containerImage) on the
      OpenStackClient CR spec
    - When enabled, the OpenStackClient controller adds a rhos-mcps MCP
      server sidecar container sharing the same clouds.yaml/secure.yaml
      credential mounts
    - Controller creates a ConfigMap with rhos-mcps config (openstack
      enabled, openshift disabled, allow_write: false) and a Service on
      port 8080 for the MCP endpoint
    - OpenStackAssistant can reference an OpenStackClient CR by name via
      the openstackClientRef field; the controller auto-computes the
      service URL and TLS CA configuration

    Tests:
    - Unit tests for the OpenStackAssistant controller covering
      reconciliation, pod creation, config generation, and status
      conditions
    - Unit tests for helper functions (entrypoint script generation,
      config building, hash computation)


lpiwowar (dismissed):
I do not mean it in a bad way and I understand this can be overwritten:), but I believe the introduction of the OpenStackAssistant feature as it is can be detrimental to the feedback gathering mechanism currently in place for OpenStack Lightspeed.

As of now, there is no mechanism in place that can differentiate transcripts gathered from the goose part and the OpenStackLightspeed part. If both the goose part and the OpenStackLightspeed part are deployed in an environment, then both transcripts get mixed up, and at the end of the transcript gathering pipeline, it won't be possible for us to distinguish between them. It will be hard for us to find lessons to be learned for both OpenStackLightspeed and the OpenStackAssistant.

I believe we should find a way how to resolve this nicely.

dprince:
> I do not mean it in a bad way and I understand this can be overwritten:), but I believe the introduction of the OpenStackAssistant feature as it is can be detrimental to the feedback gathering mechanism currently in place for OpenStack Lightspeed.
> 
> As of now, there is no mechanism in place that can differentiate transcripts gathered from the goose part and the OpenStackLightspeed part. If both the goose part and the OpenStackLightspeed part are deployed in an environment, then both transcripts get mixed up, and at the end of the transcript gathering pipeline, it won't be possible for us to distinguish between them. It will be hard for us to find lessons to be learned for both OpenStackLightspeed and the OpenStackAssistant.
> 
> I believe we should find a way how to resolve this nicely.

Would it be reasonable to file your suggestion as a new feature and work on it separately? This PR is already quite large and implements the basis of a CLI. In the meantime if a user really wanted isolated transcripts they could deploy 2 separate OpenStackLightspeed stacks and approach it that way as well.

lpiwowar:
@dprince I did a little investigation today and I have to admit I need to tone down my concerns a bit. Sorry about that. I still think the transcript/feedback gathering story is something we should look at, but it is less urgent (blocking) than I initially thought.

1.  The `v1/responses` endpoint in lightspeed-stack does not gather transcripts when the client sends `store: false` (the server side default is `store: true`) [1][2].

2. If I'm reading it correctly Goose always explicitly sets `store: false` in its `/v1/responses` requests [3]. 

This resolves my main concern -> that transcripts from OpenStack Lightspeed and Goose would be mixed together. Since Goose skips transcripts gathering entirely (`store: false`), there is no way for the transcripts and feedback to be mixed up -> transcript and feedback collection can be treated as a standalone feature to address later.

[1] https://github.com/lightspeed-core/lightspeed-stack/blob/main/src/app/endpoints/responses.py#L302
[2] https://developers.openai.com/api/reference/python/resources/responses/methods/create
[3] https://github.com/aaif-goose/goose/blob/86846ec325c74940a2948c375fc3b49b92e51918/crates/goose/src/providers/formats/openai_responses.rs#L578

---
The transcript and feedback gathering feature would probably require (as already discussed + additional new things based on the investigation)

1.  lightspeed-stack needs to distinguish between requests coming from Goose and from the console UI, and propagate that information into the gathered transcripts and feedback (e.g., via a custom HTTP header).

2. Goose needs to be configurable to send `store: true` when the user has transcript and feedback gathering enabled.

No. 1. and no 2. gets as to the point where we are collecting transcripts and feedback but then we face this problem:

3.  If I am not mistaken then Goose does not send `previous_response_id` in its `/v1/responses` requests, so lightspeed-stack treats each request as a new conversation. This means that a 5 message conversation in Goose would appear as 5 separate conversations in the gathered data.

lpiwowar:
I'm removing "requesting changes" since the "issues" mentioned in my last comment can be resolve separately.

dprince:
recheck

dprince:
@coderabbitai full review

dprince:
@coderabbitai resume

api/assistant/v1beta1/openstackassistant_types.go
```diff
@@ -0,0 +1,238 @@
...
+
+	// MCPServers lists MCP server endpoints to configure as Goose extensions.
+	// +kubebuilder:validation:Optional
+	MCPServers []MCPServerRef `json:"mcpServers,omitempty"`
+}
+
+// OpenStackAssistantSpec defines the desired state of OpenStackAssistant
+type OpenStackAssistantSpec struct {
+	// ContainerImage for the assistant container (will be set to environmental default if empty).
+	// +kubebuilder:validation:Optional
```
abays, line 120: I think we usually make these required, since the defaulting will take care of it if it's not provided.

internal/controller/client/openstackclient_controller.go
```diff
@@ -301,19 +330,226 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		if err != nil {
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				clientv1.OpenStackClientReadyCondition,
+				condition.RequestedReason,
+				condition.SeverityInfo,
+				"waiting for internal Keystone endpoint"))
+			return ctrl.Result{RequeueAfter: time.Duration(5) * time.Second}, nil
+		}
+
+		// Create TLS certificate for MCP server when TLS is enabled
```
stuggi, line 350: today all certs get created in the ctlplane controller. since the openstackclient also gets created in the ctlplane, would it architectural better align if we also create it in the ctlplane controller and pass the cert secret to the openstackclient?

internal/controller/client/openstackclient_controller.go
```diff
@@ -301,19 +330,226 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+			"configVarsHash":    configVarsHash,
+		})
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("error calculating MCP Service hash: %w", err)
+		}
+		_, err = controllerutil.CreateOrPatch(ctx, r.Client, mcpService, func() error {
+			if mcpService.Annotations == nil {
+				mcpService.Annotations = map[string]string{}
+			}
+			mcpService.Annotations["client.openstack.org/config-hash"] = mcpServiceHash
```
stuggi, line 465: What is this hash annotation used for?

internal/controller/assistant/suite_test.go
```diff
@@ -0,0 +1,120 @@
+/*
+Copyright 2022.
```
abays, line 2:
```suggestion
Copyright 2026.
```

internal/controller/assistant/openstackassistant_controller.go
```diff
@@ -0,0 +1,978 @@
...
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			assistantv1.OpenStackAssistantReadyErrorMessage,
+			err.Error()))
+		return ctrl.Result{}, err
+	}
+	configVars[instance.Spec.LightspeedStack.ProviderSecret] = env.SetValue(providerSecretHash)
+
+	// Validate optional CaBundleSecret
+	if instance.Spec.LightspeedStack.CaBundleSecretName != "" {
```
stuggi: would it make sense that any optional CA should be added to the ctlplane (via the current documented procedure) and the assistant just consumes the full bundle from the namespace of the ctlplane? or do we expect that an assistant gets created outside a ctlplane namespace?

api/assistant/v1beta1/openstackassistant_types.go
```diff
@@ -0,0 +1,238 @@
...
+	// Must contain key "lightspeed.json".
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	ProviderSecret string `json:"providerSecret"`
+
+	// CaBundleSecretName is the name of a Secret containing CA certs
+	// to trust for TLS connections to the lightspeed-stack endpoint.
+	// The Secret must contain a key "ca-bundle.crt" with PEM-encoded certs.
+	// +kubebuilder:validation:Optional
+	CaBundleSecretName string `json:"caBundleSecretName,omitempty"`
```
stuggi, line 53: should probably embed tls.Ca inline for consistency to align with how we do it in other CRDs, like the openstackclient:

```
	// +operator-sdk:csv:customresourcedefinitions:type=spec
	// Secret containing any CA certificates which should be added to deployment pods
	tls.Ca `json:",inline"`
```

Also for a user workflow we could default it the tls.CABundleSecret and a user could add any CA for the assistant to the namespace wide bundle via the ctlplane?

internal/controller/assistant/openstackassistant_controller.go
```diff
@@ -0,0 +1,978 @@
...
+
+				mcpSvcName := mcp.OpenStackClientRef + "-mcp"
+				if osclient.Spec.CaBundleSecretName != "" {
+					mcpCaBundleSecretName = osclient.Spec.CaBundleSecretName
+				}
+				scheme := "http"
+				if osclient.Spec.CaBundleSecretName != "" {
+					scheme = "https"
+				}
+				mcpURL := fmt.Sprintf("%s://%s.%s.svc:8080/openstack/", scheme, mcpSvcName, instance.Namespace)
```
stuggi, line 350: instead of constructing the mcp url, should the openstackclient store the url in its status and like we have for memcache or galera to have a `GetMcpEndpoint` or similar to query it? would also cover/make obsolete the scheme condition check above.

internal/controller/assistant/openstackassistant_controller.go
```diff
@@ -0,0 +1,978 @@
...
+						condition.ErrorReason,
+						condition.SeverityWarning,
+						assistantv1.OpenStackAssistantReadyErrorMessage,
+						"OpenStackClient "+mcp.OpenStackClientRef+" does not have MCP enabled"))
+					return ctrl.Result{}, nil
+				}
+
+				mcpSvcName := mcp.OpenStackClientRef + "-mcp"
+				if osclient.Spec.CaBundleSecretName != "" {
+					mcpCaBundleSecretName = osclient.Spec.CaBundleSecretName
```
stuggi, line 344: related to my comment on the spec re ca

internal/controller/assistant/openstackassistant_controller.go
```diff
@@ -0,0 +1,978 @@
...
+				return ctrl.Result{}, nil
+			}
+			return ctrl.Result{}, err
+		}
+		configVars["mcp-ca-bundle"] = env.SetValue(mcpCaHash)
+	}
+
+	// Build combined CA bundle when MCP TLS is in use, merging the internal CA
+	// (tls-ca-bundle.pem) with the lightspeed CA (ca-bundle.crt) if present.
+	// This handles same-secret, different-secret, and MCP-only cases.
```
stuggi, line 379: I think this would be not needed, if we add any additional CA to the ctlplane bundle and consume the env bundle.

internal/controller/assistant/openstackassistant_controller.go
```diff
@@ -0,0 +1,978 @@
...
+			Resources: []string{"network-attachment-definitions"},
+			Verbs:     []string{"get", "list", "watch"},
+		},
+		{
+			APIGroups: []string{"cert-manager.io"},
+			Resources: []string{"certificates", "issuers"},
+			Verbs:     []string{"get", "list", "watch"},
+		},
+		{
+			APIGroups: []string{
```
stuggi, line 749: looks like this list is not complete? like watcher and cyborg, but also others not in here, or could we just use `*.openstack.org` ?

stuggi (commented):
two general question are:
* to use the ctlplane CA bundle and use the documented procedure on how to add additional CAs there, or have the controller to create the bundle as it currently uis
* also if we create the mcp cert in the openstackclient controller or if we create it in the ctlplane controller like we do with other certs. But depends on if we expect that additional openstackclient instances get created which also need an mcp. then you'd either have to manually issue the cert or we keep it as is.

Also I think we could reuse more lib-common functions and how we implemented things in other controllers. like the 3 example commits added in 
https://github.com/dprince/openstack-operator/compare/assistant...stuggi:openstack-operator:mcp-tls-lib-common?expand=1

internal/controller/assistant/openstackassistant_controller.go
```diff
@@ -0,0 +1,978 @@
...
+				}
+
+				if osclient.Spec.MCP == nil || !osclient.Spec.MCP.Enabled {
+					instance.Status.Conditions.Set(condition.FalseCondition(
+						assistantv1.OpenStackAssistantReadyCondition,
+						condition.ErrorReason,
+						condition.SeverityWarning,
+						assistantv1.OpenStackAssistantReadyErrorMessage,
+						"OpenStackClient "+mcp.OpenStackClientRef+" does not have MCP enabled"))
+					return ctrl.Result{}, nil
```
abays, line 339: This won't requeue unless a watch wakes up the reconciler.  Perhaps we should do:

```suggestion
					return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
```

dprince: should be fixed in the latest PR

internal/controller/assistant/openstackassistant_controller.go
dprince: fixed in the latest push

internal/controller/assistant/openstackassistant_controller.go
dprince: see latest push

internal/controller/assistant/openstackassistant_controller.go
dprince: moved this to openstackclient controller

internal/controller/client/openstackclient_controller.go
dprince: see latest push

api/client/v1beta1/openstackclient_types.go
```diff
@@ -35,6 +37,18 @@ type OpenStackClientSpec struct {
...
+
+	// +kubebuilder:validation:Optional
+	// MCPContainerImage for the MCP server sidecar container (set via OpenStackVersion)
+	MCPContainerImage string `json:"mcpContainerImage,omitempty"`
+}
+
+// MCPConfig defines optional MCP server sidecar configuration
+type MCPConfig struct {
+	// Enabled controls whether the MCP server sidecar is added to the pod.
+	// +kubebuilder:validation:Optional
```
stuggi, line 49: the optional could probably be removed as we do not use a ptr here, its either enabled or disabled. 

```
type MCPConfig struct {
	// Enabled controls whether the MCP server sidecar is added to the pod.
	// +kubebuilder:default=false
	Enabled bool `json:"enabled,omitempty"`
}
```

internal/controller/assistant/openstackassistant_controller.go
```diff
@@ -0,0 +1,959 @@
...
+// +kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=clusterroles,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=clusterrolebindings,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=client.openstack.org,resources=openstackclients,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=endpoints;events;persistentvolumeclaims;pods/log,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=nodes;persistentvolumes,verbs=get;list;watch
+// +kubebuilder:rbac:groups=apps,resources=daemonsets;replicasets;statefulsets,verbs=get;list;watch
+// +kubebuilder:rbac:groups=assistant.openstack.org,resources=*,verbs=get;list;watch
+// +kubebuilder:rbac:groups=backup.openstack.org,resources=*,verbs=get;list;watch
+// +kubebuilder:rbac:groups=batch,resources=cronjobs,verbs=get;list;watch
+// +kubebuilder:rbac:groups=cyborg.openstack.org,resources=*,verbs=get;list;watch
```
stuggi: does it need access to cyborg ?

internal/controller/assistant/openstackassistant_controller.go
```diff
@@ -0,0 +1,959 @@
...
+
+	cl := condition.CreateList(
+		condition.UnknownCondition(assistantv1.OpenStackAssistantReadyCondition, condition.InitReason, assistantv1.OpenStackAssistantReadyInitMessage),
+		condition.UnknownCondition(condition.ServiceAccountReadyCondition, condition.InitReason, condition.ServiceAccountReadyInitMessage),
+		condition.UnknownCondition(condition.RoleReadyCondition, condition.InitReason, condition.RoleReadyInitMessage),
+		condition.UnknownCondition(condition.RoleBindingReadyCondition, condition.InitReason, condition.RoleBindingReadyInitMessage),
+	)
+	instance.Status.Conditions.Init(&cl)
+	instance.Status.ObservedGeneration = instance.Generation
+
```
stuggi, line 200: Having this below finalizer early return will result in Ready=True on first reconcile. Move above deletion-timestamp check like we do e.g. in https://github.com/openstack-k8s-operators/openstack-operator/blob/main/internal/controller/core/openstackcontrolplane_controller.go#L227

internal/controller/assistant/suite_test.go
```diff
@@ -0,0 +1,120 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package assistant
+
+import (
```
stuggi, line 19: I think we remove this file here and tests to test/functional/... ? do to you plan to have envtests for the assistant?

internal/controller/client/openstackclient_controller.go
```diff
@@ -301,19 +331,257 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				clientv1.OpenStackClientReadyCondition,
+				condition.RequestedReason,
+				condition.SeverityInfo,
+				"waiting for internal Keystone endpoint"))
+			return ctrl.Result{RequeueAfter: time.Duration(5) * time.Second}, nil
+		}
+
+		// MCP TLS defaults to enabled independently of the downstream CA bundle.
+		mcpTLSEnabled := true
```
stuggi, line 355: is it expected that this is hard coded true? I mean the else is never reached with this.

api/assistant/v1beta1/openstackassistant_types.go
```diff
@@ -0,0 +1,245 @@
...
+	SchemeBuilder.Register(&OpenStackAssistant{}, &OpenStackAssistantList{})
+}
+
+// IsReady - returns true if OpenStackAssistant is reconciled successfully
+func (instance OpenStackAssistant) IsReady() bool {
+	return instance.Status.Conditions.IsTrue(OpenStackAssistantReadyCondition)
+}
+
+// RbacConditionsSet - set the conditions for the rbac object
+func (instance OpenStackAssistant) RbacConditionsSet(c *condition.Condition) {
```
stuggi, line 205: RbacConditionsSet uses a value receiver on both OpenStackClient (api/client/v1beta1/openstackclient_types.go:122) and OpenStackAssistant, so any call to Set that has to append to the conditions slice silently drops the update — the mutation is lost on the copy.

Fix is one character each:
```
// before
func (instance OpenStackAssistant) RbacConditionsSet(c *condition.Condition) {
// after
func (instance *OpenStackAssistant) RbacConditionsSet(c *condition.Condition) {
```
I think we have this issue across controllers/operators

internal/controller/assistant/openstackassistant_controller.go
```diff
@@ -0,0 +1,959 @@
...
+		}
+	}
+
+	return requests
+}
+
+func (r *OpenStackAssistantReconciler) findObjectsForSrc(ctx context.Context, src client.Object) []reconcile.Request {
+	requests := []reconcile.Request{}
+
+	Log := r.GetLogger(context.Background())
```
stuggi, line 930: context.Background() -> ctx ?

internal/controller/assistant/openstackassistant_controller.go
```diff
@@ -0,0 +1,959 @@
...
+	assistantPod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      instance.Name,
+			Namespace: instance.Namespace,
+		},
+	}
+	if err := r.Get(ctx, client.ObjectKeyFromObject(assistantPod), assistantPod); err != nil {
+		if !k8s_errors.IsNotFound(err) {
+			return ctrl.Result{}, fmt.Errorf("failed to get OpenStackAssistant pod %s: %w", assistantPod.Name, err)
+		}
```
abays, line 441: Maybe we should check the pod ownership here.  For instance, if someone created a `OpenStackAssistant` named `openstackclient`, it would delete the `openstackclient` pod belonging to the control plane, right?

internal/controller/client/openstackclient_controller.go
```diff
@@ -301,19 +331,257 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+
+		mcpConfigCM := &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      instance.Name + "-mcp-config",
+				Namespace: instance.Namespace,
+			},
+		}
+		_, err = controllerutil.CreateOrPatch(ctx, r.Client, mcpConfigCM, func() error {
+			mcpConfigCM.Data = map[string]string{
+				"config.yaml": openstackclient.MCPConfigYAML(instance.Name, instance.Namespace, instance.Spec.CaBundleSecretName, mcpTLSEnabled),
```
stuggi, line 434: MCPConfigYAML is called twice here and bellow. could call it once and store it in a var

internal/webhook/assistant/v1beta1/openstackassistant_webhook.go
```diff
@@ -0,0 +1,102 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	assistantv1beta1 "github.com/openstack-k8s-operators/openstack-operator/api/assistant/v1beta1"
+)
+
+// nolint:unused
```
stuggi, line 33: its actually used, do we heed the `nolint:unused`?

api/assistant/v1beta1/conditions.go
```diff
@@ -0,0 +1,52 @@
...
+	OpenStackAssistantReadyInitMessage = "OpenStack Assistant not started"
+
+	// OpenStackAssistantReadyRunningMessage
+	OpenStackAssistantReadyRunningMessage = "OpenStack Assistant in progress"
+
+	// OpenStackAssistantReadyMessage
+	OpenStackAssistantReadyMessage = "OpenStack Assistant created"
+
+	// OpenStackAssistantReadyErrorMessage
+	OpenStackAssistantReadyErrorMessage = "OpenStack Assistant error occured %s"
```
stuggi, line 39: "occured" → "occurred"

api/assistant/v1beta1/openstackassistant_types.go
```diff
@@ -0,0 +1,245 @@
...
+
+	// Goose contains Goose-specific provider configuration.
+	// Only applicable when provider is "goose".
+	// +kubebuilder:validation:Optional
+	Goose *GooseConfig `json:"goose,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// Secret containing any CA certificates which should be added to the assistant pod.
+	tls.Ca `json:",inline"`
```
stuggi, line 142: should the ca secret be defaulted to `combined-ca-bundle` https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/tls/tls.go#L44 via the webhook and only have to be set by the user if really needed?

internal/controller/client/openstackclient_controller.go
```diff
@@ -301,19 +331,257 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+			return ctrl.Result{RequeueAfter: time.Duration(5) * time.Second}, nil
+		}
+
+		// MCP TLS defaults to enabled independently of the downstream CA bundle.
+		mcpTLSEnabled := true
+		if mcpTLSEnabled {
+			issuerName := tls.DefaultCAPrefix + string(service.EndpointInternal)
+			mcpSvcName := instance.Name + "-mcp"
+			certRequest := certmanager.CertificateRequest{
+				IssuerName: issuerName,
```
abays, line 360: What if `spec.tls.podLevel.internal.ca.customIssuer` is used instead?  Looks like we're hard-coding `rootca-internal` here.

internal/controller/client/openstackclient_controller.go
```diff
@@ -301,19 +331,257 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+			if err != nil {
+				instance.Status.Conditions.Set(condition.FalseCondition(
+					clientv1.OpenStackClientReadyCondition,
+					condition.ErrorReason,
+					condition.SeverityWarning,
+					clientv1.OpenStackClientReadyErrorMessage,
+					err.Error()))
+				return ctrl.Result{}, err
+			}
+			configVars["mcp-tls"] = env.SetValue(certHash)
```
abays, line 405: From Codex Astra:

The certificate contributes to the pod hash, but the Secret watch only indexes the CA bundle and existing OpenStack configuration inputs. Changes to `cert-<name>-mcp-tls` therefore don’t trigger reconciliation. Because the certificate is mounted using `subPath`, the sidecar retains the old certificate until another event or periodic resync recreates the pod. Add a watch mapping for this `Secret`.

internal/controller/client/openstackclient_controller.go
stuggi: Sorry, I think my comment was misunderstood. I didn't mean to move cert creation into the openstackclient controller itself — I meant that the cert should be created in internal/openstack/openstackclient.go on the ctlplane side (where we already orchestrate all the other ctlplane-managed certs), and the resulting secret name passed down into the OpenStackClient CR as a parameter. That way the ctlplane owns all cert lifecycle and the openstackclient controller just consumes what it's given, rather than reaching up to manage its own cert. would that make sense?

config/operator/default_images.yaml
```diff
@@ -153,8 +153,12 @@ spec:
           value: quay.io/podified-antelope-centos9/openstack-octavia-worker:current-podified
         - name: RELATED_IMAGE_OCTAVIA_RSYSLOG_IMAGE_URL_DEFAULT
           value: quay.io/podified-antelope-centos9/openstack-rsyslog:current-podified
+        - name: RELATED_IMAGE_OPENSTACK_ASSISTANT_IMAGE_URL_DEFAULT
+          value: quay.io/openstack-s2i-containers/openstack-goose:master-latest
         - name: RELATED_IMAGE_OPENSTACK_CLIENT_IMAGE_URL_DEFAULT
           value: quay.io/podified-antelope-centos9/openstack-openstackclient:current-podified
+        - name: RELATED_IMAGE_OPENSTACK_MCP_IMAGE_URL_DEFAULT
```
rabi, line 160: Is this not going to be managed with openstackversion when one is not using standalone openstackclient?  I don't even see it in [SetupDefaults()](https://github.com/openstack-k8s-operators/openstack-operator/blob/main/api/client/v1beta1/openstackclient_types.go#L137)

api/assistant/v1beta1/openstackassistant_types.go
```diff
@@ -0,0 +1,245 @@
...
+	// namespace that has MCP enabled. The controller reads the service URL
+	// published in the OpenStackClient status.
+	// Mutually exclusive with URL.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MinLength=1
+	OpenStackClientRef string `json:"openstackClientRef,omitempty"`
+}
+
+// GooseConfig defines Goose-specific provider configuration
+type GooseConfig struct {
```
rabi: Do we need to make it tightly coupled with goose? I feel it would be better to keep it agent agnostic and config generated by a entrypoint script that generates the config based on env vars. 

Also, recipes are very goose specific.

api/assistant/v1beta1/openstackassistant_types.go
```diff
@@ -0,0 +1,245 @@
...
+)
+
+const (
+	// OpenStackAssistantContainerImage is the fall-back container image for OpenStackAssistant
+	OpenStackAssistantContainerImage = "quay.io/openstack-s2i-containers/openstack-goose:master-latest"
+)
+
+// ProviderType defines the AI agent provider
+// +kubebuilder:validation:Enum=goose
+type ProviderType string
```
rabi: It's confusing to use provider for both the agent/harness and provider (LLM or Lightspeed).

api/assistant/v1beta1/openstackassistant_types.go
```diff
@@ -0,0 +1,245 @@
...
+)
+
+// LightspeedStackSpec defines connectivity to the Lightspeed Stack AI backend
+type LightspeedStackSpec struct {
+	// ProviderSecret is the name of a Secret containing the lightspeed
+	// provider config JSON (custom_providers/lightspeed.json content).
+	// Must contain key "lightspeed.json".
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	ProviderSecret string `json:"providerSecret"`
```
rabi: Is it necessary to expect users to provide lightspeed provider config? It would only require base_url and model, we get the api key from service account token anyway right?

internal/openstackclient/funcs.go
```diff
@@ -113,13 +107,140 @@ func ClientPodSpec(
...
+			{
+				Name:      "mcp-config",
+				MountPath: "/home/cloud-admin/.config/openstack/clouds.yaml",
+				SubPath:   "clouds.yaml",
+				ReadOnly:  true,
+			},
+			{
+				Name:      "openstack-config-secret",
+				MountPath: "/home/cloud-admin/.config/openstack/secure.yaml",
+				SubPath:   "secure.yaml",
```
rabi, line 121: Should not this mounted as readonly too?

rabi (commented):
I've a few inline comments. 

Also, keeping in mind how difficult it is to make CRD changes later, I think it would probably be better to keep the option of using different harnesses open rather than making it goose specific. One can have the flexibility to change the Assistant image with entrypoint scripts doing the required configuation for the agent.

dprince:
@rabi agree with the direction of you comments (making the implemention generic). I had tried to do this with "provider" before but agree that didn't go far enough. I've moved the entrypoint.sh into the openstack-goose containerfile. One reason I exposed the entire provider config before was to allow it to container multiple models. To continue supporting this pattern I added additionalModels.

@stuggi @abays your comments should also be addressed.

dprince:
@rabi entrypoint PR is here: https://github.com/openstack-k8s-operators/s2i-openstack-containers/pull/216

rabi:
> @rabi entrypoint PR is here: [openstack-k8s-operators/s2i-openstack-containers#216](https://github.com/openstack-k8s-operators/s2i-openstack-containers/pull/216)

Thanks! I've a few comments there. I guess this has to wait for  s2i PR to merge first.

dprince:
recheck

dprince:
recheck

rabi (commented):
LGTM

stuggi (approved):
/lgtm

PR #1932, opened 2026-05-28
Add TLS certificate for InstanceHA metrics endpoint

lmiccini:
Create a cert-manager Certificate for the InstanceHA metrics service when pod-level TLS is enabled. The certificate uses the internal issuer with wildcard hostnames for the namespace, following the same pattern as EnsureOVNMetricsCert. The resulting secret (cert-instanceha-metrics) is consumed by the infra-operator InstanceHA controller.

### DIFF
--- a/internal/openstack/instanceha.go
+++ b/internal/openstack/instanceha.go
@@ -2,7 +2,11 @@ package openstack
 
 import (
 	"context"
+	"fmt"
 
+	certmgrv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
+	"github.com/openstack-k8s-operators/lib-common/modules/certmanager"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/clusterdns"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/configmap"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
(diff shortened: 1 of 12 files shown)


internal/openstack/instanceha.go
```diff
@@ -21,6 +25,22 @@ const (
...
 // ReconcileInstanceHa reconciles the instance HA configuration for the OpenStack control plane
 func ReconcileInstanceHa(ctx context.Context, instance *corev1beta1.OpenStackControlPlane, version *corev1beta1.OpenStackVersion, helper *helper.Helper) (ctrl.Result, error) {
+	Log := GetLogger(ctx)
+
+	if instance.Spec.TLS.PodLevel.Enabled {
+		_, err := EnsureInstanceHAMetricsCert(ctx, instance, helper)
+		if err != nil {
+			Log.Error(err, "Failed to ensure InstanceHA metrics certificate")
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				corev1beta1.OpenStackControlPlaneInstanceHaCMReadyCondition,
```
abays: Is there a possible semantic mismatch here?  I just mean that this error is related to the TLS cert, not the CM, right?  I guess it would be extra work to add a new condition though, and perhaps this error would be rare, so this is probably fine.

abays (approved):
/lgtm

internal/openstack/instanceha.go
lmiccini: I didn't want to add a new condition just for the certificate and I kinda "cheated" putting all the eggs in one basket, especially since the instanceha is somehow a second class citizen of the controlplane (we just create the prerequisites cm but the rest is done by infra).

internal/openstack/instanceha.go
```diff
@@ -21,6 +25,26 @@ const (
...
+			Log.Error(err, "Failed to ensure InstanceHA metrics certificate")
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				corev1beta1.OpenStackControlPlaneInstanceHaTLSReadyCondition,
+				condition.ErrorReason,
+				condition.SeverityWarning,
+				corev1beta1.OpenStackControlPlaneInstanceHaTLSReadyErrorMessage,
+				err.Error()))
+			return ctrl.Result{}, err
+		}
+		instance.Status.Conditions.Set(condition.TrueCondition(
```
fmount, line 42: @lmiccini are these conditions initialized somewhere to `Unknown`? Maybe it doesn't matter as they're not part of the main condition pool.

lmiccini: no they are not, they're just cosmetic and set during reconcile to signal stuff is ready or not

dciabrin (approved):

PR #1938, opened 2026-06-12
Allow ACME-issued certificates for public endpoint routes

marbindrakon:
When a user supplies a custom TLS secret for a public endpoint route
(via `apiOverride`), the operator required the secret to contain
`tls.crt`, `tls.key` **and** `ca.crt`. Certificates issued by an ACME
issuer (e.g. Let's Encrypt) do not populate `ca.crt`: the issuing chain
is delivered inline in `tls.crt`, and the signing root is publicly
trusted, so a separate `ca.crt` entry is neither produced nor needed.
This made it impossible to use ACME-issued certificates for public
endpoints without synthesising a `ca.crt` key.

This makes `ca.crt` optional on that path:

- The secret key check is extracted into `validateRouteCertSecret`,
  which now requires only `tls.crt` and `tls.key`.
- The route's `CACertificate` is set only when the secret actually
  provides `ca.crt`, instead of writing an empty value.

Certificate chains continue to be served correctly because cert-manager
packs the full issuing chain into `tls.crt`; `CACertificate` is only
used to advertise an additional custom CA, which remains supported when
present. Re-encrypt (TLSE) routes are unaffected — their
`DestinationCACertificate` is still sourced from the internal CA bundle.

This covers the "bring your own secret" flow. Having the operator
manage an ACME `Certificate` directly (which additionally requires
guarding the custom-issuer CA-extraction path and splitting public vs
internal SAN profiles) is intentionally out of scope here.

Jira: [RHOSRFE-109](https://redhat.atlassian.net/browse/RHOSRFE-109)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/openstack/common.go
+++ b/internal/openstack/common.go
@@ -616,13 +616,9 @@ func (ed *EndpointDetail) CreateRoute(
 				return ctrl.Result{}, err
 			}
 
-			// check if secret has the expected entries tls.crt, tls.key and ca.crt
-			if certSecret != nil {
-				for _, key := range []string{"tls.crt", "tls.key", "ca.crt"} {
-					if _, exist := certSecret.Data[key]; !exist {
-						return ctrl.Result{}, fmt.Errorf("certificate secret %s does not provide %s", *ed.Route.TLS.SecretName, key)
-					}
-				}
+			// check the secret has the required tls.crt and tls.key entries
+			if err := validateRouteCertSecret(certSecret, *ed.Route.TLS.SecretName); err != nil {
+				return ctrl.Result{}, err
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


marbindrakon:
As proposed, this patch will not stop a user from providing a privately-signed certificate secret. In that case, they would be responsible for adding the correct root to the CA bundle using the existing mechanism.

internal/openstack/common.go
stuggi: nit, here we probably want to remove cacert

stuggi:
thanks for submitting the PR! With this we could also optimize procedures in 
https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0/html/configuring_security_services/custom-tls-certificates_security-services#updating-the-control-plane-with-custom-certificates_tlse-rhoso
and
https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0/html/configuring_security_services/custom-tls-certificates_security-services#updating-the-control-plane-with-a-single-custom-certificate_tlse-rhoso
to not have to add the ca cert to the cert secret as we document to already add it to the bundle secret.

FYI @Deydra71 @vakwetu

internal/openstack/common.go
marbindrakon: Done

stuggi: thanks, if you could squash the commits, that would be good. Thanks!

stuggi (approved):
/lgtm

PR #1949, opened 2026-06-22
Set ImagePullPolicy to IfNotPresent on OpenStackClient pod

abays:
The OpenStackClient pod is created as a raw corev1.Pod directly, bypassing lib-common workload helpers where SetPullPolicyDefaults is applied. When ImagePullPolicy is not explicitly set and the image tag is "latest" or unset, Kubernetes defaults the policy to Always. Explicitly set PullIfNotPresent to avoid unnecessary image pulls.


stuggi (approved):
/lgtm

abays:
Fixed indentation, adding LGTM label back

PR #1961, opened 2026-06-26
Remove tj-actions/branch-names from upgrade catalog workflow

stuggi:
Replace tj-actions/branch-names with github.ref_name which provides the branch name natively without a third-party action. The tj-actions GitHub namespace was compromised in March 2025 (CVE-2025-30066) and using actions from that namespace is no longer recommended. Pass the value via env: to avoid shell interpolation of untrusted input in run: blocks.

Jira: [OSPRH-31981](https://redhat.atlassian.net/browse/OSPRH-31981)


abays (approved):
/lgtm

stuggi:
overriding kuttl test, the PR is not touching operator code

PR #1965, opened 2026-07-01
Move service operator RBAC from runtime bindata to OLM bundle

dprince:
Instead of applying service operator ClusterRoles, ClusterRoleBindings, Roles, RoleBindings, and ServiceAccounts at runtime via bindata, generate them during sync-bindata.sh and stage them into the OLM bundle at bundle build time. This removes the need for the openstack-operator to hold wildcard clusterrole/clusterrolebinding permissions and narrows the remaining role/rolebinding RBAC to specific verbs.


dprince:
recheck

stuggi:
looks good to me. @abays if you are also good with it, lets get it in

stuggi (approved):
/lgtm

PR #1968, opened 2026-07-03
Remove dependency on OVN for Octavia

gthiemonge:
Octavia can be used with a 3rd party SDN, don't require OVN to be enabled

JIRA: [OSPRH-32131](https://redhat.atlassian.net/browse/OSPRH-32131)


stuggi (approved):
/lgtm

PR #1973, opened 2026-07-07
Use ECDSA P-256 private keys for cert-manager certificates

fmount:
Switch all `cert-manager` `Certificate` resources from the default `RSA` algorithm to `ECDSA` `P-256`. This applies to `webhook` serving certs, `metrics` server certs, and all operator `bindata` `certificates`. The `sync-bindata.sh` script is updated accordingly (and it properly regenerates all the services files).

Jira: https://redhat.atlassian.net/browse/OSPRH-27314

### DIFF
--- a/bindata/operator/infra-operator-webhooks.yaml
+++ b/bindata/operator/infra-operator-webhooks.yaml
@@ -37,6 +37,9 @@ spec:
   issuerRef:
     kind: Issuer
     name: infra-operator-selfsigned-issuer
+  privateKey:
+    algorithm: ECDSA
+    size: 256
   secretName: infra-operator-webhook-server-cert
 ---
 apiVersion: admissionregistration.k8s.io/v1
--- a/bindata/operator/openstack-baremetal-operator-webhooks.yaml
+++ b/bindata/operator/openstack-baremetal-operator-webhooks.yaml
@@ -37,6 +37,9 @@ spec:
   issuerRef:
     kind: Issuer
     name: openstack-baremetal-operator-selfsigned-issuer
+  privateKey:
+    algorithm: ECDSA
+    size: 256
   secretName: openstack-baremetal-operator-webhook-server-cert
 ---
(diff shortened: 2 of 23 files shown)


fmount:
recheck

stuggi:
I guess its ok to get this in 18 as well? for the 18/19 branching, we plan to branch one more from main. we'll discuss in our tomorrows meeting

fmount:
> I guess its ok to get this in 18 as well? for the 18/19 branching, we plan to branch one more from main. we'll discuss in our tomorrows meeting

This is technically ok for 18 as well. The target is 19 of course, so we can also hold it until we branch a `stable-18`. We can discuss more sure.

fmount:
> > I guess its ok to get this in 18 as well? for the 18/19 branching, we plan to branch one more from main. we'll discuss in our tomorrows meeting
> 
> This is technically ok for 18 as well. The target is 19 of course, so we can also hold it until we branch a `stable-18`. We can discuss more sure.

@stuggi as this is not a goal for 18, my proposal is to land this after we branch `stable-18`.

bindata/operator/infra-operator-webhooks.yaml
```diff
@@ -37,6 +37,9 @@ spec:
   issuerRef:
     kind: Issuer
     name: infra-operator-selfsigned-issuer
+  privateKey:
+    algorithm: ECDSA
+    size: 256
```
bshephar, line 42: `256` is the default for `ECDSA` it seems:
https://github.com/cert-manager/cert-manager/blob/master/internal/apis/certmanager/types_certificate.go#L311-L319

Probably don't need to explicitly set the size. But, weird that they would default to `RSA` in 2026. I guess you could also set the size and not the algorithm too. But it's not a problem, just providing the feedback since I went to the trouble of looking up the defaults.

bshephar (approved):

stuggi (approved):
/lgtm

PR #1990, opened 2026-07-20
Detect TLS cert secret changes in NodeSet reconciler

oliashish:
Cert-manager renewals were invisible because cert secrets are owned by the Certificate CR, not the NodeSet. Add label-based detection in secretWatcherFn and hash comparison in checkDeployment to surface redeployment required signal after cert rotation.

Jira: [OSPRH-32671](https://redhat.atlassian.net/browse/OSPRH-32671)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/internal/controller/dataplane/openstackdataplanenodeset_controller.go
+++ b/internal/controller/dataplane/openstackdataplanenodeset_controller.go
@@ -596,6 +596,16 @@ func checkDeployment(ctx context.Context, helper *helper.Helper,
 				continue
 			}
 
+			hasCertSecretsChanged, err := checkCertSecretsChanged(ctx, helper, instance, deployment.Status.SecretHashes)
+
+			if err != nil {
+				return isNodeSetDeploymentReady, isNodeSetDeploymentRunning, isNodeSetDeploymentFailed, failedDeploymentName, err
+			}
+
+			if hasCertSecretsChanged {
+				continue
+			}
+
 			isNodeSetDeploymentReady = true
 			for k, v := range deployment.Status.ConfigMapHashes {
 				instance.Status.ConfigMapHashes[k] = v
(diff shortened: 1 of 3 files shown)


internal/controller/dataplane/openstackdataplanenodeset_controller.go
```diff
@@ -986,3 +1014,27 @@ func checkAnsibleVarsFromChanged(
...
+	instance *dataplanev1.OpenStackDataPlaneNodeSet,
+	deployedSecretHashes map[string]string,
+) (bool, error) {
+	currentCertHashes, err := deployment.GetCertSecretHashes(ctx, helper, instance.Namespace, instance.Name)
+	if err != nil {
+		return false, err
+	}
+
+	for name, currentHash := range currentCertHashes {
+		if deployedHash, exists := deployedSecretHashes[name]; exists && deployedHash != currentHash {
```
slagle, line 1039: a review with claude has pointed out:

>   The new checkCertSecretsChanged uses exists && deployedHash != currentHash — a new cert secret that wasn't present at deployment time is not treated as a change. This     
>   means if a new TLS service is added to the NodeSet after the last deployment, cert secrets for it won't trigger redeployment.                                              
> 
>   This may be intentional (new certs are picked up on the next deployment anyway), but the inconsistency is worth a comment or explicit design choice.

internal/controller/dataplane/openstackdataplanenodeset_controller.go
```diff
@@ -853,6 +863,24 @@ func (r *OpenStackDataPlaneNodeSetReconciler) secretWatcherFn(
 	ctx context.Context, obj client.Object,
 ) []reconcile.Request {
 	Log := r.GetLogger(ctx)
+
+	// Check if this is a cert secret (has both NodeSet and Service labels).
+	// Cert secrets created by EnsureTLSCerts carry these labels but are not
+	// listed in AnsibleVarsFrom, so the field-index lookup below would miss them.
+	labels := obj.GetLabels()
+	if nodeSetName, ok := labels[deployment.NodeSetLabel]; ok {
+		if _, hasSvcLabel := labels[deployment.ServiceLabel]; hasSvcLabel {
```
slagle: claude is recommending to also check the existence of the ServiceKeyLabel here in order to be more precise.

internal/controller/dataplane/openstackdataplanenodeset_controller.go
oliashish: Adding a new TLS service to the NodeSet means changing instance.Spec.Services. That changes instance.Spec, which changes ConfigHash which hashes the entire instance.Spec. Since ConfigHash != DeployedConfigHash, the deployment is already skipped before we reach here at
```
if deployment.Status.NodeSetHashes[instance.Name] != instance.Status.ConfigHash {
      continue
 }

```

So the existing spec-hash check already covers this case.

internal/controller/dataplane/openstackdataplanenodeset_controller.go
oliashish: Ack.
Added ServiceKeyLabel check for completeness. Technically redundant since EnsureTLSCerts always sets all three labels together, but makes the match more precise.

internal/controller/dataplane/openstackdataplanenodeset_controller.go
slagle: There is also the case of changing a service (that is already in the NodeSet spec) to add a TLS cert, but that seems pretty unlikely.

slagle:
recheck

slagle (approved):

internal/dataplane/hashes.go
```diff
@@ -139,3 +139,36 @@ func ProcessAnsibleVarsFrom(
...
+func GetCertSecretHashes(
+	ctx context.Context,
+	helper *helper.Helper,
+	namespace string,
+	nodeSetName string,
+) (map[string]string, error) {
+	labelSelector := map[string]string{
+		NodeSetLabel: nodeSetName,
+	}
+	secrets, err := secret.GetSecrets(ctx, helper, namespace, labelSelector)
```
rabi, line 155: Don't we have all secret names in the deployedSecretHashes? We can avoid this lus api call which is heavy. Also, we can reuses the label selector logic already in GetDeploymentHashesForService().

oliashish: deployedSecretHashes contains DataSource, AnsibleVarsFrom, and cert secrets in one map, there is no way
to distinguish without fetching each.
Reusing GetDeploymentHashesForService would need a get per service + a List per service/certKey, which is in nested loop.

rabi: We probably need to refactor the status struct as it's currently flat map and also nodeset now contains secret hashes for other nodesets via deployment. That would make all these much simpler.

PR #1992, opened 2026-07-20
Add per-branch CRD sync check workflows

stuggi:
Replace the generic crd-sync-check-fr.yaml (stale, pointed at fr5) with per-branch workflow files for 18.0-fr6 and 18-stable. This makes it easier to add/remove branches and avoids stale references.

Jira: [OSPRH-32639](https://redhat.atlassian.net/browse/OSPRH-32639)


abays (approved):
/lgtm

PR #1994, opened 2026-07-24
Add dedicated RELATED_IMAGE for instanceha instead of reusing openstackclient

4383:
Instanceha pods were using OpenstackClientImage, which prevented OLM from independently overriding the instanceha container image for RHOSO deployments. This adds a dedicated RELATED_IMAGE_INSTANCE_HA_IMAGE_URL_DEFAULT env var and InstanceHaImage field so instanceha can be configured separately.

### DIFF
--- a/api/bases/core.openstack.org_openstackcontrolplanes.yaml
+++ b/api/bases/core.openstack.org_openstackcontrolplanes.yaml
@@ -21099,6 +21099,8 @@ spec:
                     type: string
                   infraRedisImage:
                     type: string
+                  instanceHaImage:
+                    type: string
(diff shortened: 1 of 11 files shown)


lmiccini:
Thanks Herve, we had this before and we decided to avoid adding a new constant/field/etc since we are sharing the same image as the openstackclient. We can fix this in infra-operator by flipping the logic around, so spec -> configmap -> webhook defaulting instead of what we have now https://github.com/openstack-k8s-operators/infra-operator/pull/634

4383:
@lmiccini no problem, thank you, then I'm closing this PR

PR #2008, opened 2026-07-31
Set readOnlyRootFilesystem on service operator controller-manager containers

stuggi:
Harden the manager container securityContext by adding readOnlyRootFilesystem: true and dropping ALL capabilities. Mount an emptyDir at /tmp to support controller-runtime cert auto-generation and Go stdlib temporary file needs.

Jira: [OSPRH-34174](https://redhat.atlassian.net/browse/OSPRH-34174)


Deydra71:
VErified with barbican, keystone and glance operators that read only root FS was applied in controller managers without changing service operator code individually.

abays (approved):
/lgtm

PR #2013, opened 2026-08-04
Require confirm-delete annotation for OpenStackDataPlaneDeployment deletion

rabi:
Deleting a Deployment mid-execution may leave data plane nodes in an inconsistent or unrecoverable state. Require users to set the annotation dataplane.openstack.org/confirm-delete=true before deletion is allowed.

jira: [OSPRH-32794](https://redhat.atlassian.net/browse/OSPRH-32794)

### DIFF
--- a/api/dataplane/v1beta1/const.go
+++ b/api/dataplane/v1beta1/const.go
@@ -19,4 +19,8 @@ package v1beta1
 const (
 	// CtlPlaneNetwork - default ctlplane Network Name in NetConfig
 	CtlPlaneNetwork = "ctlplane"
+
+	// ConfirmDeleteAnnotation is the annotation key required to allow
+	// deletion of an OpenStackDataPlaneDeployment. The value must be "true".
+	ConfirmDeleteAnnotation = "dataplane.openstack.org/confirm-delete"
 )
--- a/api/dataplane/v1beta1/openstackdataplanedeployment_webhook.go
+++ b/api/dataplane/v1beta1/openstackdataplanedeployment_webhook.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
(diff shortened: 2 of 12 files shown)


slagle (approved):
LGTM, I think we should add this to the docs (upstream and downstream) somewhere. With an example one-liner that can be used to add the annotation:

`
oc annotate openstackdataplanedeployment <deployment-name> dataplane.openstack.org/confirm-delete=true
`

rabi:
> LGTM, I think we should add this to the docs (upstream and downstream) somewhere. With an example one-liner that can be used to add the annotation:
> 
> `oc annotate openstackdataplanedeployment <deployment-name> dataplane.openstack.org/confirm-delete=true`

The error clearly tells what the user has to do to .. Do you think we still need to document it (i.e how to add an annotation to q CR, which is basic ocp thing)? 

`Error from server (Forbidden): admission webhook "vopenstackdataplanedeployment-v1beta1.kb.io" denied the request: openstackdataplanedeployments.dataplane.openstack.org "<name>" is forbidden: deletion of a running deployment not allowed: set annotation dataplane.openstack.org/confirm-delete=true to confirm`

Maybe add in the releasenote that this is the new changed behavior.

slagle:
> > LGTM, I think we should add this to the docs (upstream and downstream) somewhere. With an example one-liner that can be used to add the annotation:
> > `oc annotate openstackdataplanedeployment <deployment-name> dataplane.openstack.org/confirm-delete=true`
> 
> The error clearly tells what the user has to do to .. Do you think we still need to document it (i.e how to add an annotation to q CR, which is basic ocp thing)?
> 
> `Error from server (Forbidden): admission webhook "vopenstackdataplanedeployment-v1beta1.kb.io" denied the request: openstackdataplanedeployments.dataplane.openstack.org "<name>" is forbidden: deletion of a running deployment not allowed: set annotation dataplane.openstack.org/confirm-delete=true to confirm`
> 
> Maybe add in the releasenote that this is the new changed behavior.

A releasenote would be good :+1:

PR #2038, opened 2026-08-18
Move watcher to s2i container images

amoralej:
This is just changing the variable used in defaulting logic to the new consolidated container image from s2i.

So far, this is just to test the approach. We may want to move all defaults to s2i in openstack-operator.


amoralej:
check-rdo

PR #2051, opened 2026-08-24
Add seccompProfile to sub-operator deployment template

abays:
The managers.yaml template generates Deployments for ~20 child operators. While the template already includes `readOnlyRootFilesystem` and capability drop (added in `010517ca`), it was missing `seccompProfile` at the pod security context level.

This adds `seccompProfile: type: RuntimeDefault` to match the umbrella operator's own deployment and satisfy the Restricted Pod Security Standard.


abays:
recheck

dprince (approved):

PR #2052, opened 2026-08-25
[DNM][kuttl] Drop null Job pod template creationTimestamp asserts

nemarjan:
cifmw-multinode-kuttl started failing this test after the GitHub check jobs moved from OCP 4.20 to 4.22. The dataplane deploy itself succeeds; KUTTL then compares live Jobs against a full YAML dump that still has Job.spec.template.metadata.creationTimestamp: null (leftover from oc get -o yaml).

On 4.20 the API still serialized that null key, so the assert matched. On 4.22 Kubernetes omits empty creationTimestamp (omitempty), so the key is missing from the live object. KUTTL waits the full timeout for a field that will never appear, then fails with:

  .spec.template.metadata.creationTimestamp: key is missing from map

Remove the dumped null fields from the Job asserts so the test does not depend on that serialization difference.

### DIFF
--- a/test/kuttl/tests/dataplane-deploy-global-service-test/01-assert.yaml
+++ b/test/kuttl/tests/dataplane-deploy-global-service-test/01-assert.yaml
@@ -205,7 +205,6 @@ spec:
   suspend: false
   template:
     metadata:
-      creationTimestamp: null
       labels:
         app: openstackansibleee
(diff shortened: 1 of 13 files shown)


abays:
recheck

abays (approved):
/lgtm

Valkyrie00 (approved):

abays:
recheck

PR #2057, opened 2026-08-31
Fix typo in heat Cfn as it is refered to as Cnf instead.

david-hill:
heat-cfn is the right service name instead of heat-cnf which confuses operators.


dprince:
@david-hill good catch but we can't just merge this as is. This will need to be coordinated with some migration path from old -> new value. Possibly a conversion webhook.

PR #2061, opened 2026-09-04
Improve logic to fix functional tests

rabi:
- latest-deployment sort returned 1 on equal/missing Ready transition times and same-second transitions would pick an arbitrary "latest". Tie-break on creationTimestamp, then name
- nodeset controller is purely event-driven; coalesced watch events left a stale "running" verdict unreconciled until timeout. Requeue after 15s while a deployment is running
- log the previously silent config-hash mismatch so a stuck nodeset is diagnosable from logs

### DIFF
--- a/api/dataplane/v1beta1/openstackdataplanedeployment_types.go
+++ b/api/dataplane/v1beta1/openstackdataplanedeployment_types.go
@@ -23,6 +23,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// DefaultDeploymentRequeueTime is the default requeue interval in seconds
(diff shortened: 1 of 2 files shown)


rabi:
recheck

rabi:
recheck

rabi:
recheck

rabi:
recheck

rabi:
`dial tcp: lookup api.oko-09.oooci.ccitredhat.com on 172.30.0.10:53: read udp 10.128.186.186:51803->172.30.0.10:53: i/o timeout`

slagle (approved):

rabi:
`No clusters in pool are ready to be claimed `

PR #2069, opened 2026-09-08
Propagate MariaDB TargetVersion to the Galera CR (OSPRH-32083)

zzzeek:
Adds ServiceDefaults.MariadbVersion to OpenStackVersion and propagates it to the Galera CR's spec.targetVersion, following the pattern already used by ServiceDefaults.RabbitmqVersion for the rabbitmq-server 3.9->4.2 upgrade.

Currently does not indicate a "default" version for mariadb-galera, as 10.5 is already deployed in existing deployments that dont have the "targetversion" field at all; setting it to 10.5 here would force a restart of those clusters.

Once this change is running on a rhel-10 environment with mariadb 10.11 available as the mariadb image,
the RELATED_IMAGE_MARIADB_IMAGE_URL_DEFAULT should be added as "10.11"; when deployed on an existing cluster that runs 10.5, this will apply the new version to the galera cluster which will then do a restart and run the mariadb upgrade scripts.  Doing so before a 10.11 image is present would put running clusters into a state that includes MariaDBServerUpgradeReady=False with UpgradeVersionMismatch (however those clusters would still remain in a ready state).

References: [OSPRH-32083](https://redhat.atlassian.net/browse/OSPRH-32083)

### DIFF
--- a/api/bases/core.openstack.org_openstackversions.yaml
+++ b/api/bases/core.openstack.org_openstackversions.yaml
@@ -258,6 +258,8 @@ spec:
                       type: string
                     manilaSharev1:
                       type: string
+                    mariadbVersion:
+                      type: string
                     rabbitmqVersion:
                       type: string
                   type: object
@@ -698,6 +700,8 @@ spec:
                   glanceWsgi:
                     type: string
                   manilaSharev1:
+                    type: string
+                  mariadbVersion:
                     type: string
                   rabbitmqVersion:
                     type: string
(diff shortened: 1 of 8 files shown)


lmiccini (approved):

dciabrin (approved):

stuggi (approved):
/lgtm

dciabrin:
/retest
unrelated CI failure
INFO[2026-09-17T16:46:28Z] trying next host                              error=failed to authorize: failed to fetch oauth token: unexpected status from GET request to https://image-registry.openshift-image-registry.svc:5000/openshift/token?scope=repository%3Aci-op-75xkw5zv%2Fpipeline%3Apull: 401 Unauthorized host=image-registry.openshift-image-registry.svc:5000

PR #2075, opened 2026-09-11
Fix crd-size workflow: set GH_TOKEN for gh api call

stuggi:
The "Compute CRD JSON size" step uses `gh api` to fetch the PR head's CRD file, but did not set GH_TOKEN. gh requires GH_TOKEN in Actions and exited with code 4 before producing output, failing the check.

Add GH_TOKEN: ${{ github.token }} to the step env. The existing
`contents: read` permission is sufficient for the contents API read.

### DIFF
--- a/.github/workflows/crd-size-badge.yaml
+++ b/.github/workflows/crd-size-badge.yaml
@@ -30,6 +30,7 @@ jobs:
       - name: Compute CRD JSON size
         id: size
         env:
+          GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


stuggi:
seen in https://github.com/openstack-k8s-operators/openstack-operator/actions/runs/34595540502/job/103271787488#logs

abays (approved):
/lgtm

abays:
/override "CRD Size Badge / crd-size (pull_request_target)"

PR #2078, opened 2026-09-15
[WIP] Remove IronicInspector references <JIRA:OSPRH-36973>

mumesan:
**Description:**
Inspector has been moved from a service to integrate with ironic as an agent, as a result all references to it need to be removed from ironic-operator. This PR removes it in openstack-operator so it compiles successfully when the references are removed from ironic-operator.


stuggi:
@mumesan does it mean that inspector is gone and no longer used and there is no migration needed for any config which a user set for it before?

mumesan:
> @mumesan does it mean that inspector is gone and no longer used and there is no migration needed for any config which a user set for it before?

Hi @stuggi , we've decided not to proceed with inspector removal from ironic-operator at this time and I'm going to close this PR.
For context on why the PR was created, in the latest version of ironic, it integrates inspector into it as an agent rather than as a standalone service. For now we're going to have both inspector and agent based introspection models as a transition period and remove the inspector code later.

PR #2082, opened 2026-09-16
[kuttl] Drop null Job pod template creationTimestamp asserts (18-stable)

nemarjan:
Cherry-pick of #2052 to 18-stable.

OCP 4.22 no longer serializes `creationTimestamp: null` in Job pod template
metadata (`omitempty`). The kuttl assertions still expect that field, causing
`dataplane-deploy-global-service-test` to fail with:

```
.spec.template.metadata.creationTimestamp: key is missing from map
```

Remove the dumped null fields from the Job asserts so the test does not
depend on that serialization difference.

Signed-off-by: nemarjan <nemarjan@redhat.com>

Made with [Cursor](https://cursor.com)


stuggi (approved):
/lgtm

PR #2087, opened 2026-09-18
[WIP] Propagate inspection interface annotation to Ironic CR using ServiceDefaults

mumesan:
#### Description:
The latest version of ironic removes inspector as the separate service to run introspection and uses IronicPythonAgent instead. This PR adds the IronicInspectInterface field to ServiceDefaults so that openstack-operator can tell ironic-operator which interface to use per deployment so as to maintain backwards compatibility with deployments that require inspector. 

Jira: [OSPRH-37063](https://redhat.atlassian.net/browse/OSPRH-37063)


internal/openstack/version.go
```diff
@@ -251,6 +264,9 @@ func InitializeOpenStackVersionServiceDefaults(ctx context.Context) *corev1beta1
 	// "10.11" in the same change that moves
 	// RELATED_IMAGE_MARIADB_IMAGE_URL_DEFAULT to a 10.11 image.
 
+	inspectInterface := getIronicInspectInterface(containerDefaults.IronicConductorImage)
```
steveb: I don't see any other service inferring a default value based on something like this. I think it would be best to just set this to "inspector" by default with a good comment on 18 vs 19 differences and learn how we're setting this to "agent" in the OpenStackVersion resources that represent 19 service images. Maybe @dprince can provide some context

mumesan: It is set to inspector as default, but you're right that inferring it from the image name is not ideal. I'll check if semver is an option and if so how that could be applied.

